PR #16, opened 2015-01-29
Added "callmonitor" for users of an AVM FRITZ!Box

paviro:
Probably only useful for German people, but maybe it can be modified for other phone systems as well.


lesander:
FRITZ!Box products are popular in the Netherlands too :)

paviro:
Nice to hear! :) You find a newer version here: https://github.com/paviro/MagicMirror-Extensions

MichMich:
I like the idea. But I don't want to include this in de main project. I think we should keep this as an extension.

PR #17, opened 2015-03-07
Update calendar.php

kaiser07:
Fixed some errors when getting data from icloud (SSL related errors)


MichMich:
Thanks for your afford!

Issue #23, opened 2015-06-26
Redistribution of Proprietary Font Files


woodrowbarlow:
Helvetica Neue is a [proprietary font](https://www.myfonts.com/fonts/linotype/neue-helvetica/licensing.html). I don't think that redistributing the font files in this way is considered fair use and it might result in legal action. I propose that Helvetica Neue remains the first option in the font-stack, but that the font files are removed and a secondary font (perhaps [Roboto](https://www.google.com/fonts#UsePlace:use/Collection:Roboto)?) is shipped with the software load. That way, if you have the Helvetica Neue font on your system (e.g., if you purchased it), that font will be used, and Roboto will be used otherwise.

PR #24, opened 2015-06-26
Replace Helvetica Neue with Open-Source Font

woodrowbarlow:
Helvetica Neue is a proprietary font. Redistributing it in this
way is not considered fair use. This commit removes the Helvetica
Neue fonts from the repository and instead includes Roboto. It is
written in such a way that if you copy the Helvetica Neue fonts
back into the `fonts/` directory (under a `helvetica-neue` subfolder)
they will be used as the primary font instead of Roboto.

This addresses issue #23


woodrowbarlow:
here is a font comparison, for your benefit:

Hevetica Neue versus Roboto font comparison

ftbastler:
:+1:

PR #30, opened 2015-10-23
Modularization

blindman:
Hey,

Let me know what you think of this but I updated the front-end folder structure to separate the concerns of each part of the application.  Essentially, every part of the front-end has its own folder.  Let me know what you think or if you need any explanation for what is going on.  I haven't gotten around to updating the README to explain all of the files/folders but I plan on doing that soon.  A lot of the code is straight from the original `main.js` just transplanted into a different place.  I also have a branch for a configuration page but it's dependent on this code.


MichMich:
Hey Man,

Great work. This was still on my todo list. I'll look into your changes asap. It might be a good idea to update the README before I can merge this with the master branche. But more importantly, we might want to move back to the original compliments. (Although I really enjoyed your compliments ... ;))

Cheers, Michael.

blindman:
Crap, I completely forgot that I left those in there haha.  Yes, I intended to leave all of the original config options with what you had so I'll change those back.  I'll update spend some time updating the README and obviously change those compliments back to the originals.  When you get a chance, shoot me suggestions and/or thoughts you might have and I'll be happy to throw them in there and re-submit the pull request.

blindman:
I added comments to the functions as well as updated the README and reverted the config file to what you had originally.  Let me know when you're able to take a look at this.  I'd love to include the config page I have implemented but I want to make sure you're fine with this first so I can adjust where necessary.

Config Example here

MichMich:
I'll look into it asap. The config looks awesome. How are you planning to store the config settings?

blindman:
The [same way](https://github.com/blindman/MagicMirror/blob/configPage/js/config.js) as before, using a [php route](https://github.com/blindman/MagicMirror/blob/configPage/config.php) to create the config from a [template file](https://github.com/blindman/MagicMirror/blob/configPage/js/config/config.js.template).

MichMich:
Thanks for your cleaning up! :)

PR #47, opened 2016-01-12
Timeformat from HH:MMSS to HH:MM:SS

dasquiddi:


MichMich:
Could you provide a screenshot? I don't think it matches my idea of the UI.

PR #56, opened 2016-01-25
Calendar improvment

bitte-ein-bit:
This Pull request contains three calendar improvements:
* Support for multiple calendars
* Make update of table smoother and update only rows which change
* Show events currently running with endtime


MichMich:
Very nice update! Just one thing: since I will be using just one calendar, I don't want an icon. If I leave the icon blank, the calendar isn't aligned anymore. Could you modify this? - Thanks!

bitte-ein-bit:
There is an option I introduced to disable the icon. It's false if unset
and in the config.js I pushed it's set to true.

The behavior of an 'unaligned' empty icon is on purpose, because an
other calendars might have an Icon set.

Hope that's ok with you.

Am Di, 26. Jan 2016, um 15:43, schrieb Michael Teeuw:

MichMich:
My bad! I totally missed that disable function. Nice work! :) I'll commit your PR asap.

PR #57, opened 2016-01-25
Smooth time

bitte-ein-bit:
Update Time charwise and create smooth effect


MichMich:
With this update, the seconds don't appear. Could you fix this?

PR #64, opened 2016-01-28
Fix PHP Include error

bitte-ein-bit:
Mirror may be running below document root i.e http://pi.local/magicmirror/

### DIFF
--- a/index.php
+++ b/index.php
@@ -36,6 +36,6 @@
 <script src="js/news/news.js"></script>
 <script src="js/main.js?nocache=<?php echo md5(microtime()) ?>"></script>
 <!-- <script src="js/socket.io.min.js"></script> -->
-<?php  include($_SERVER['DOCUMENT_ROOT'].'/controllers/modules.php');?>
+<?php  include(dirname(__FILE__).'/controllers/modules.php');?>
 </body>
 </html>


paviro:
Thanks for pointing out!

CFenner:
a relative path would be sufficient. (include('controllers/modules.php');)

efiku:
I propose better way. (Without composer)
* Separate PHP code from HTML. 
* Make one file app.php with simple routing system and call controllers/modules from html code using `$.ajax` from jQuery.

and why [print_r](http://php.net/manual/en/function.print-r.php) in `modules.php` ?  :smiley: 

Example `app.php`
```php
<?php

$getController =  $_GET['get'] ?? ""; // using PHP7+ or isset( $_GET['get'] ) ? $_GET['get']  : "" ;


$controllers = [
  "hash"  => require( "controllers/hash.php" ),
  "helllo" => "world"
];
 
if(array_key_exists( $getController , $controllers ) ){
  header('Content-Type: application/json');
  echo json_encode($controllers[$getController]; 
 }
```
Example `hash.php`
```php
 return [
    "gitHash" => trim(`git rev-parse HEAD`)
 ]; 

```

bitte-ein-bit:
@efiku This pr fixes just php warning. The original code is from @paviro who stated in one comment that he is pretty new to php. 
Feel free to improve it but remember the code has to run on an PI (which ships with php5) and flexible. So adding a controller to the crontollerhash by hand is not a real option.

paviro:
Can't tell why I used print_r because like @bitte-ein-bit stated I have never really used PHP before. I just needed something to load my custom elements and some javascript if anyone can improve the way I did that or the whole modules system, I'd be glad to learn something new :D

PR #65, opened 2016-01-29
Smooth time

bitte-ein-bit:
Added Code to fade changing Digits, with optional switches to show seconds / fade digits.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -110,10 +110,6 @@ html {
   color: #aaa;
 }
 
-.hidden {
-  display: none;
-}
-
 .light {
   font-family: "HelveticaNeue-UltraLight", sans-serif;
 }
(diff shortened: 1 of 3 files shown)


MichMich:
I really appreciate your effort, but I don't think it's a visual improvement. With all the fading, It all becomes a bit messy, IMHO.

bitte-ein-bit:
I agree that when seconds are displayed it's messy. That's the reason I
left them out in the first request. 

But when you don't display seconds it's a really smooth effect similar
to those old displays which took some time to update. The reason I went
down this fading way, was that IMHO the original code made the mirror
restless. So I removed the seconds and added the digit-wise fading. 

If you think it would look better to just fade hours and minutes, I
could change to version. If not would adding the code as an non default
option be ok with you?

Am Fr, 29. Jan 2016, um 15:13, schrieb Michael Teeuw:

MichMich:
Yes, using an option in the config would be ok. :)

PR #70, opened 2016-01-31
Smooth time

bitte-ein-bit:
Added default config options and flipped digitFade to false as a default value. 
Added some Notes to README.

Hope that was what you had in mind.


MichMich:
Thanks! Nice work!

PR #71, opened 2016-01-31
Quotes of the day

thegunslingers:
Added a Quotes of the day feature as well as an option to display the compliments, the quotes or both.  (One note I will make, is some of the quotes that are extremely long may need more complex formatting.


js/config.js
```diff
@@ -36,7 +38,7 @@ var config = {
     calendar: {
         maximumEntries: 10, // Total Maximum Entries
 		displaySymbol: true,
-		defaultSymbol: 'calendar', // Fontawsome Symbol see http://fontawesome.io/cheatsheet/
+		defaultSymbol: 'ICAL URL Here', // Fontawsome Symbol see http://fontawesome.io/cheatsheet/
```
bitte-ein-bit: Sorry thats wrong here

thegunslingers:
Fixed that sorry, accidentally changed that when I was changing back the rest of my config file.

bitte-ein-bit:
No problem, just noticed.

Had that problem of having to change the config back too until I started
to use a my-config branch where all my local config is in. This way when
I want to fix or add something all I have to do is checkout
MichMich/master, create a branch from there and work on it. Then push
that to my github repo and create a merge request. After that I can
merge this branch in my master and rebase my config branch on my new
master. This way my config is never on my master branch and cannot be
accidentally pushed to github.


Am So, 31. Jan 2016, um 20:07, schrieb Scott Rodgers:
> Fixed that sorry, accidentally changed that when I was changing back
> the rest of my config file.


> — Reply to this email directly or view it on GitHub[1].


>



Links:

  1. https://github.com/MichMich/MagicMirror/pull/71#issuecomment-177586363

MichMich:
I like this idea, but I prefer it to be an optional module. We're still working on the module system though ...

PR #72, opened 2016-01-31
Better module loader

bitte-ein-bit:
I rewrote the module loader. It's now more dynamic and hopefully flexible. I moved the sample module to an extra git repo. After the merge you have to init the subrepro. For details see the updated README.

I changed the code so that only an `include.php` is nesessary. The rest is either used if present (`elements.html`) or has to be included in desired order.

@paviro: Would you mind splitting your extensions in separte git repos. One per Extensions?

@MichMich: Maybe we could have some sort of list of modules added to the original repo? And maybe it would be worth considering moving some code into an module as well?

Thanks


paviro:
You mean [those](https://github.com/paviro/MagicMirror-Modules) extensions? No I would not mind just have to figure out a way to do it since they all require some kind of python and/or node server running and I put them all into unified scripts...
You would need to start multiple scripts and I wanted to avoid that so that it does not get too complex with setting up, but if you have an idea or are interested to help me out we can of course change that :)

bitte-ein-bit:
Yes I mean those extensions. I think it would be possible to add a similar module loader to the python part. But I have to check and look at your code.

About the auto add. You might be right that an main.js or style.css is almost always needed. I just wanted to keep the code short. The reason I did not include them anymore but moved them to the include.php is that all is available to just load them there. So I thought why force someone to use all four files if two are sufficient. The auto include of the elements was just one way of choosing to either add an element option in include.php or just include it, if it's there but don't bother if it's not.

The list was just some idea. Let's call it an index or something like that. The reason I suggested is the following. When the modules themselves are in separate repos, no one will oversee what is available. This index could be added to the README and is partially there. Maybe some sort of section 'Available Modules' with each Module Author describing their module in a few lines.

paviro:
Would be amazing! Would simplify a lot of things. I should probably start to use [python-daemon](https://pypi.python.org/pypi/python-daemon/) in that process. 
Here is a short description of my code:
The callmonitor extensions only needs a node server, which monitors the FRITZ!Box and sends calls to the UI via socket.io
My faceregognition code is split in two. The python part does the actual regognition and is opening a UNIX socket. Right know the same node script running the callmonitor connects to that socket and sends data to the UI via socket.io. In theory I could of course run two socket.io servers for each module or we write a general scripts -> modules bridge so only one socket.io connection is needed and our bridge can take connections via socket.io and UNIX sockets or something.

Makes sense :+1: 

Yes would be nice!

MichMich:
Hi Guys,

I've got some ideas about the modules I'd like to discuss with you before we completely overhaul the system. "Unfortunately" I'm about to step on a plane for a holiday. So you guys need to have a little patience. :) 

Keep up the good work! 

Cheers, Michael.

paviro:
Alright have a good time and let us know when you are back :)

MichMich:
Yes! I'm back! :) - Let's continue with this. In mine opinion, it might be a good idea to work out a nice module API before we start building. I've got some idea's on how I would like it to be, but not enough time to really built it. 

Also, since this will be a huge overhaul, we might want to first put this in a separate branch so people can test. 

I've created a small wiki page on how this API could look like. Please let me know your thoughts: https://github.com/MichMich/MagicMirror/wiki/Module-System-Proposal. (And feel free to edit.)

paviro:
Nice work! @bitte-ein-bit I would love to work with you on this but I am not sure my knowledge on PHP and JavaScript is well enough for that. I can surely help out but probably not really code any core functions.

bitte-ein-bit:
I'll take a look next week at the wiki.

MichMich:
@bitte-ein-bit: Also check out this discussion: https://github.com/MichMich/MagicMirror/issues/80

Would you agree to close this pull request until we worked out a new API?

bitte-ein-bit:
I'll take a look there and leave my thoughts.

I wouldn't close this request. This way people will see that there is an
active redesign if they look either at the pull requests or issues tab.
Just let it hang around.


Am So, 7. Feb 2016, um 13:28, schrieb Michael Teeuw:
> @bitte-ein-bit[1]: Also check out this discussion: #80[2]


> Would you agree to close this pull request until we worked out a
> new API?


>


> — Reply to this email directly or view it on GitHub[3].
>


>



Links:

  1. https://github.com/bitte-ein-bit
  2. https://github.com/MichMich/MagicMirror/issues/80
  3. https://github.com/MichMich/MagicMirror/pull/72#issuecomment-181005076

MichMich:
Good suggestion! Will leave it open.

CFenner:
@paviro what about moving your motion detection in an own module and build around some kind of event notification so that other modules can react on a detected motion (start face recognition, fetch the newest weather data).

paviro:
As soon as there is a module system I will happily do that!

MichMich:
Since we are working on V2, I close this pull request.

PR #75, opened 2016-02-03
Hopefully better version check.

bitte-ein-bit:
Works on Windows to where by default php cannot use git. Uses .git filestructure then. If that fails to, return some static text

### DIFF
--- a/controllers/hash.php
+++ b/controllers/hash.php
@@ -1,7 +1,34 @@
 <?php
+function getHash() {
+	$hash = exec('git rev-parse HEAD', $output, $returncode);
+	if ($returncode == 0) {
+		return '0-'.$hash;
+	}
+	// No git for php? Let's try parsing git structures
+	$file = dirname(str_replace('\\','/',__DIR__))."/.git/HEAD";
+	if (is_file($file)) {
+		$refOrCommit = file_get_contents($file);
+		if ($refOrCommit !== false && $refOrCommit !== null) {
+			if (strpos($refOrCommit, 'ref') !== false) {
+				list($ref, $head) = explode(' ', $refOrCommit);
+				$head = trim($head);
+				$file = dirname(str_replace('\\','/',__DIR__))."/.git/$head";
+				return '1-'.trim(file_get_contents($file));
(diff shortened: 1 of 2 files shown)


controllers/hash.php
```diff
@@ -1,7 +1,34 @@
 <?php
+function getHash() {
```
efiku: In my opinion this part of code is not necessary why?
First of all if you are working with files in PHP, better way is using **Iterators**. Like [DirectoryIterator](http://php.net/manual/en/class.directoryiterator.php) or [SplFileObject](http://php.net/manual/en/class.splfileobject.php)

Problem with git can be solved like this:
1. Linux → **sudo apt install git -y** ( Git problem solved. )
2. **Windows → install git core and add git to env path.** ( Git problem solved )

Better PR in my opinion:
* Create MakeFile or make.sh which all necessary dependencies  or add to Readme info about dependencies ;)
* Option 2: create virtual machine form http://phansible.com/

```sh
#!bin/bash
sudo apt update
sudo apt install  php5-common php5-intl php5  apache or nginx etc..
[…]
```

event15: +1 for efiku.

bitte-ein-bit: @event15 @efiku  Sure, go ahead and improve my PR. The branch is available. Or start a new branch. But keep in mind, should work without a git command available. And bonus points for being able to use a Fallback.

To be honest. The reason for this code is that git for windows is not a real requirement and many people here definitely are not using git as a command line tool. This was just playing nice...

MichMich:
Maybe we could change this feature by allowing to refresh the screen by a command line command. Usually after a git pull, you still need to change the config file. Maybe we could do something like

    touch refresh

And let the mirror check the edit date of `refresh`

Any thoughts?

bitte-ein-bit:
I don't need to change the config. My Config is in a seperate branch which is the master branch of the PI but not published to github. So my dev machine has both branches and I just push in one way or the other. Because of that I prefer to not log in to the pi.

MichMich:
This will be fixed in V2.

PR #92, opened 2016-02-27
Initial config and oauth trial

### LINKED ISSUE #91: Google Calendar not working
Everything else is working, except my google calendar. I put in my url in the config file but it's not showing up. I'm using Midori as my browser.
nano

wcdanh2:
If you aren't the owner of a Google calendar then the private address is not available to you to download the ics.  Using Oauth2 and the Google Api you can achieve the same read permissions you would normally have.

I've added support for using the Google Api to allow use of shared Google calendars with the Magic Mirror.

Note: there are changes to the config.js  users should back up their old config and port their changes to the new one.

fixes #91 Google Calendar not working.

### DIFF
--- a/index.php
+++ b/index.php
@@ -31,11 +31,19 @@
 <script src="js/rrule.js"></script>
 <script src="js/version/version.js"></script>
 <script src="js/calendar/calendar.js"></script>
+<script src="https://apis.google.com/js/client.js?onload=checkAuth"></script>
 <script src="js/compliments/compliments.js"></script>
 <script src="js/weather/weather.js"></script>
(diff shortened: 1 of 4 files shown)


kyle-ayn:
It works!!! Thank you so much!!

wcdanh2:
looks like I need to recheck auth periodically. After the calendar runs for a while the google api stops working,  f5 fixes it for a while.  

Also I need to remove console.log messages.

MichMich:
Due to the refresh issue, I can't merge this. Can I close this pull request?

PR #100, opened 2016-03-09
Adding a config.js.sample to avoid to commit customizations of content in project

Pilooz:
Hi,
I added a config.js.sample that reflects The actual MagicMirror capabilities.
I added a .gitignore file ignoring the real config.js in the repo. I think the customisation of content does not have to be commited as a technical modification in the repo.
Before this PR, if I changed the News feed link, the repo was marked as modified and I had to commit these changes. By the way, content changes do not have to affect code.

Cheers.


MichMich:
I love this commit. It was on my to do list. But it might be good idea to show a warning in the frontend whenever somebody did not create the config file.

Pilooz:
Perhaps config file could be created by an administrative install task (ie install.php) ?

paviro:
V2 is now in an electron wrapper so no php anymore.

MichMich:
Fixed in V2: https://github.com/MichMich/MagicMirror/commit/90c62837426c4f5035d79715ae6881d2eeef7acf

PR #106, opened 2016-03-27
Added electron wrapper

paviro:
I know it is not the focus right now so feel free to ignore this pull request anyway I think it would be a nice addition to v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # MagicMirror V2
-This version of the Magic Mirror software focusses on a modular plugin system. 
+This version of the Magic Mirror software focusses on a modular plugin system.
+To start the software run `npm install && npm start`.
 
 #WARNING: THIS VERSION IS IN A VERY EARLY STAGE. IT IS NOT COMPLETED YET. PLEASE USE THE MASTER BRANCH.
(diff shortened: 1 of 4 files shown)


MichMich:
Hi @paviro, could you elaborate a little bit on the added value of the electron wrapper? I have no experience yet with this lib.

Thanks!

paviro:
I don't have much yet either however the biggest advantage I see is that there no longer is a need to install a webserver or browser which makes it easier for people who don't have much experience with shells. They only have to execute two commands to get the interface up and running. I guess in the future it should also make it possible to directly run node.js helper scripts this way but no guarantee on that, I just started reading some of their documentations.

MichMich:
It looks promising. But as far as I can see, Raspberry Pi is not yet supported. Which is the main platform we develop for.

paviro:
Well this [mirror](http://docs.smart-mirror.io/docs/supported_hardware.html) uses it and the Pi is under supported platforms, so it seems to work somehow.

paviro:
I added another commit, this will now automatically start any helper script called "node_helper.js" inside a module folder. The helper can then host a socket, it works for example for my call monitor any comments are welcome!

paviro:
If you are interested [here](https://github.com/paviro/MMM-FRITZ-Box-Callmonitor) an example module that uses this pull request. I tested it on a Raspberry Pi works as expected!

PR #107, opened 2016-03-28
Start helper_scripts as childs

paviro:
Start helper_scripts as childs instead of just loading them into the main script.

### DIFF
--- a/js/electron.js
+++ b/js/electron.js
@@ -2,6 +2,7 @@
 
 //for searching modules
 const walk = require('walk');
+const spawn = require('child_process').spawn;
 
 const electron = require('electron');
 // Module to control application life.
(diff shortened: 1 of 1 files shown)


MichMich:
Maybe only te helper scripts of the "loaded" modules should be started? This can be done by walking thru the config to see which modules will be loaded.

paviro:
Yes, I tried that but I could not figure out how to do it... I probably should start learning JavaScripts because I am missing a lot of basics, well after my final exams in school..
Problem is I was not able to load your config variable into the node.js "electron.js" script, if you happen to know how to do it I can implement it if you give me a hint!

MichMich:
This solution is ok for now. Maybe we should put it on a todo list in the readme.

paviro:
Sounds good I will add it and open a pull request.

PR #109, opened 2016-03-30
Fixed typo in node_helper and added module to README

paviro:


MichMich:
Very impressed by the speed in which you were able to update according to my changed! :)

Awesome example of using modules.

paviro:
Thank you! Means a lot to me :)
I also started porting my face recognition module but that will still need a bit of work.

MichMich:
Cool! Let me know how it goes. I still have a feature on my to do list where you can hide and show certain modules. This way your recognizer can show/hide agenda's depending on the person who is in front of the mirror.

paviro:
Looking forward to implement it! It's a lot more fun now to write modules than it was before!

PR #112, opened 2016-03-31
Hide cursor and updated electron params

paviro:


CFenner:
With the css set, can I ommit the installation of 'unclutter' on raspbian?

paviro:
Yes, exactly my idea :)

djsunrise19:
Which of these tools are necessary now: http://michaelteeuw.nl/post/83188136918/magic-mirror-part-v-installing-the-raspberry-pi

As I can see I need:
- screen rotation and some things in /boot/config.txt

But these installations are not necessary anymore:
- php, apache, unclutter, chromium / kioskmode

Am I right? Do I need to put the node script in my init. for reboot? Or does it starts automatically?

paviro:
Jup haven't tested it yet but hopefully the only thing you need is that and node start in `/etc/xdg/lxsession/LXDE-pi/autostart`.

MichMich:
I'll merge this request after I solve a newsfeed bug.

CFenner:
A browser is thou still neded right? At the moment I use midori as chromium does not work anymore.

paviro:
No its not!

PR #120, opened 2016-04-01
Added alert module

paviro:


MichMich:
Cool, I'll give it a try tomorrow.

MichMich:
It's a great start! Some feedback points:

**Messages:**
- Allow the possibility to disable the welcome message: `welcome_message: false` (this now displays a message: 'false')
- Allow the possibility to set the location of the grow popup. (I prefer top center as default.)
- Allow the possibility to set the text alignment of a message. (I prefer center as default.)
- Maybe allow a title and message. (Styling is important in this case. A lot of effort went into the Magic Mirror style/css).
- If multiple messages are fired a once, they overlap each other.

**Alerts:**
- It might be a good idea to overlay all the modules with a 50% opaque black div. (=Dim away the background). This prevents clutter.
- It looks a bit out of style. The font's and colors don't match with the rest of the mirror. If you dim away the background, you could just use a black box with white outline and white text ... Or make this configurable.
- `HIDE_ALERT` would hide any message. What if multiple modules display an alert. This should have a solid solution.

**General:**
- I agree we shouldn't reinvent the wheel, but this Alert mechanism heavily uses all kind of libraries. My focus on this version of the MagicMirror is to use a lot of `vanilla javascript`. Maybe we can work on a custom built system? (we could also replace the alert system in a later stage with a custom version ...)

With that being said: Thanks a lot for your effort! I really appreciate it! If we manage to match the visual style of the module with the rest of the Interface, I will add it as one of the default modules until we have a custom built version.

paviro:
I will try to work on the things you just mentioned. I'd love to write a new system in vanilla javascript but I just don't have the knowledge to do so sadly... That's why I mostly used other libraries.

paviro:
Alright I finished most of what you mentioned above :)
One problem I have is I can't get the fixed alert to center vertically...
I know there are some things that are not optimal and a bit messy but I guess that's sadly all I can do for now..

paviro:
Did you have a chance yet to check out the changes? I'd like to work on a autostart function but don't want to have added it in here.

MichMich:
There are still some issues, but I'm willing to commit it so you can continue. Unfortunately it has some conflicts, so I need to take a look at that first.

paviro:
I guess it's the README :D

MichMich:
Nope, a lot more ... Have to check how I can easily merge.

paviro:
I can also try to resolve it if you want to.

MichMich:
Would be great! :)

paviro:
Don't quite get what the problem is now...

MichMich:
I was afraid of that. Will look into it.

MichMich:
It's merged. For future pull requests: try to split them up. :)

paviro:
Alright I'll try :) Thank you!

PR #123, opened 2016-04-01
hide future events

CFenner:
allow the configuration of a limit in days till the events get displayed


MichMich:
Awesome!

Maybe it's a good idea to rename the property to `maximumNumberOfDays`? That way it's more clear what the value should be. I didn't try out the code yet. Are you sure you can slice an array when you put it in a loop like this?

CFenner:
Yes, that's a better name. 

I'm not that familiar with the backend stuff, maybe its a better idea to but the code in the node_helper? I saw the same `events = newEvents.slice(0, maximumEntries);` code there. 

I'm leaving the loop when I slice, so should be no problem.

MichMich:
I implemented it in the node helper since this already had some limit code. See commit: https://github.com/MichMich/MagicMirror/commit/e38dd346d9f796807ea71035c18e3533ec245ba6

Thanks for the suggestion.

PR #129, opened 2016-04-03
Added alert todos

paviro:


MichMich:
One thing I would absolutely change. (Which might be a good thing to do asap, because it's a breaking change: is use the same notification message for both types of message. `SHOW_ALERT` and then use a property 'type' to select between alert or message.

`SHOW_NOTIFICATION` is a bit confusing due to the word `notification`.

paviro:
Alright will do. Type would be `alert` and `notification `?

MichMich:
Yes. My first idea was 'alert' and 'message' because i want to stay away of notification because this is used internally. But I agree it's much clearer and used in other systems (like OSX notifications).

Also, i would still use milliseconds in this module (in stead of seconds), since all default modules use milliseconds for every timing thing.

paviro:
Okay :+1: How about `HIDE_ALERT`?

MichMich:
I think this has to go anyway, since this would also hide alerts created by other modules. Or you should rename it to HIDE_ALL_ALERTS.

An other approach could be as follows:

````javascript
var currentNotification = false;
self.sendNotification("SHOW_ALERT", {title: "Hello", message: "This is a test!", callback:function(alertIdentifier) {
    // this is called when the alert is shown.
    // the alertIdentifier is a reference to the specific alert.
     //to something with the alert identifier. For example:
    currentNotification = alertIdentifier.
}); 

setTimeout(function() {
    if (currentNotification) {
        currentNotification();
    }
}, 3000);
````

paviro:
But `HIDE_ALERT` executes `this.alerts[sender.name].dismiss()` and therefore only hides an alert if its from the same sender.

MichMich:
Ah, that's a smart solution as well! 👍

PR #135, opened 2016-04-04
Font Improvements - closes #133

### LINKED ISSUE #133: International Characters
screenshot_characters

The version of Roboto that is used right now does only include Latin characters. This helps with keeping the file size of the fonts small, but looks really odd especially if only single non-latin characters are used.

For testing purposes i have converted the font manually, but using the whole character set leads to huge files:

One solution for this would be to use the *Google Font API* which uses font subsets to minimize overhead.
But there are issues with this approach:
* The font can’t be loaded while offline
* Privacy concerns
* Unnecessary loading of information over the internet, that could be stored offline

These could be solved by caching or downloading the fonts and CSS files from the Google API, but Google explicitly states that their APIs must not be used in this manner.

Right now I'm working a tool to create subsets easily. I don't want to use fontsquirrel or similar sites because they aren't as customizable and are a bit tedious to use for subsetting.

Do you have any thoughts on the issue? This mostly affects names but also languages using other alphabets.

Domi-G:
- the "woff2" font format is now supported
- larger character set
- newer verision of Roboto
- removed SVG (deprecated) and EOT (Internet Explorer < 9) font format

## Before: 
before_commit

## After:
after_commit

In my opinion the newer Roboto version looks a lot better!


nhubbard:
Agreed. I like the new font.

MichMich:
Thank you very much!

PR #148, opened 2016-04-08
A Mishmash of Fixes

nhubbard:
I fixed the following in this:
- Markdown in ALL the documents.
- Updated StyleLint in the package.json.
- HTML Standardization. Almost forgot about that!


MichMich:
Thanks!

PR #156, opened 2016-04-10
Markdown Refresh

nhubbard:
"Damn Markdown" is no more. We now use all-Markdown tables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The following properties can be configured:
 
 Module configuration:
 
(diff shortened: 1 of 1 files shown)


MichMich:
You are a hero!

I'll update the header image tomorrow. (Smart Mirror > smart mirror). (If you can't wait that long, the PSD is in the .github folder ...)

All the module readme' still have the old table.

nhubbard:
OK, I'll fix that. :+1:

PR #165, opened 2016-04-12
Update README.md

Robiv8:
README.MD: Simply remove a comma after Hello World as there is with this an error. 

config.js.sample: only for German Language.


MichMich:
Fixed in https://github.com/MichMich/MagicMirror/commit/bc8dccd4b5c02a0b63a9bd102d2c6510a64edaba

PR #170, opened 2016-04-13
Remove ESLint

nhubbard:
ESLint wouldn't stop complaining about how my indentation was wrong even though I set it to tabs. So I removed it.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,13 +12,11 @@ First, before you run the linters, you will need to install them all **and** ins
 npm install
 ```
 
-### JavaScript: Run JSCS and ESLint
+### JavaScript: Run JSCS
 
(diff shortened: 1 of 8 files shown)


nhubbard:
I am considering shrinkwrapping our own version of the ical module, because the creator has not got back to me on a vulnerability I reported to him with a fix.

nhubbard:
Please disregard the previous comment. Any thoughts on the remainder of the pull request, @MichMich?

MichMich:
Will look into it tomorrow. Was at an event all day today. I did notice you added a config.js in this PR. That should not be the case.

MichMich:
@nhubbard, thanks for the pull request. I see you used the sample config as a default config. I prefer to let a user see the "no config" message first to guide them thru the process. Could you resubmit this pull request without the config changes (both the config file as well as the .gitignore change).

Thanks!

PR #179, opened 2016-04-15
Resubmitted ESLint Removal.

nhubbard:
ESLint was annoying me because it conflicted with JSCS and insisted that I use a different indentation. So I removed it. I also fixed what was wrong with my previous pull request. It should be better now. :+1:


MichMich:
Sorry to be a pain in the ass, but why did you add `.clockbold` to main.css if you can just use the default class: `.bold`?

nhubbard:
Oops, didn't notice that. Sorry!

MichMich:
An now it has conflicts ... 😨

MichMich:
also, I wouldn't add `require("dotenv").config();` to the config file. You should do this in js/defaults.js. This is the base config if there isn't any config available. 

One suggestion: split up pull requests to prevent pull requests getting denied on minor issues.

nhubbard:
Okay, I will close this. But I have only one issue: the `dotenv` module needs to be in both the default configuration *and* the user configuration because it will not load if it isn't required.

PR #202, opened 2016-04-19
Add units to config.js

roxasvalor:
Add a variable for selecting units in the config.js. Used for weather modules.


MichMich:
193 changed files ... this can't be correct.

PR #210, opened 2016-04-20
show tomorrow/today on full day events

CFenner:
on full day events:
- show `tomorrow` if the events is on the next day
- show `today` if the events is on the same day 

labels are localized via momentjs


MichMich:
Won't this show a specific time even in case of a full day event?

CFenner:
yes, you're right :(

PR #211, opened 2016-04-20
implements l10n

### LINKED ISSUE #191: Add l10n.js for localization
What do you think about adding something like [l10n.js](https://github.com/eligrey/l10n.js) to the vendor folder for localizing modules? Maybe even to the module class?
```
getLocals: function() {
		return {
en: "en.json", 
de: "de.json"
};
	},
```

CFenner:
fixes #191


paviro:
Nice! 🙂

MichMich:
Very well done! My compliments!

PR #229, opened 2016-04-23
Update README.md

djsunrise19:
Not necessary anymore, because we made that with "de.json" in calendar


MichMich:
Great! Thanks! :)

PR #237, opened 2016-04-24
Added option for units inside of config.js

roxasvalor:
I added an option for changing the units in config.js. I think this was in the original repository but it never made it to the V2 branch.


MichMich:
Nice one! 

Please also add it to https://github.com/MichMich/MagicMirror/blob/v2-beta/js/defaults.js since that holds the values in case the user doesn't configure it.

MichMich:
And make sure you add information about this to the documentation (Readme)

roxasvalor:
Sure thing. The changes to the defaults.js and readme.md should be in this pull request. First time ever contributing through Github but I think I have it done correctly.

MichMich:
Congratulations to your first ever merged pull request! :) Thanks for contributing to Magic Mirror! 👍

PR #238, opened 2016-04-24
Added getTranslations() to the documentation

paviro:


MichMich:
Thanks. Maybe we should note in the documentation that it isn't fully finished yet (no fallback in case a translation isn't present yet.)

paviro:
Maybe but isn't it normal to add all used translations to every file anyway even if they are not translated yet? At least I would work that way, so people know what still needs to be translated.

MichMich:
What if I add a translation in dutch to your repo (via a PR), and you add a feature that needs a translation. Then you don't have a dutch translation available yet. You want a nice fallback for that without going thru all the available translation files.

paviro:
Yes that's right. I would add it to all translation files in the fallback language but maybe thats just me 😀

MichMich:
It's a good approach, but the system should have a decent fallback mechanism as well. In addition to the fallback mechanism, I really want to have a global translation system implemented. Since a lot of words used in the modules are used more than once ("Tomorrow", "Today", "Loading ...", etc ...) - But this is not something we currently have to put into the Readme.

paviro:
Yes the global system should definitely be there! I will add a note to the documentation about the other thing.

MichMich:
Just add a reference to the issue (#191) as well, so people can read up on the current state.

MichMich:
A bit too late, but `this.translate` should be mentioned under Module instance methods. (Since it is ...)

PR #243, opened 2016-04-24
Add optional wind direction in currentweather module

roxasvalor:
Optionally add the wind direction (pulled from Openweather API) next to the wind speed in the current weather module. It is turned off by default but by adding the option "showWindDirection" in config.js it can be turned on.

### DIFF
--- a/modules/default/currentweather/README.md
+++ b/modules/default/currentweather/README.md
@@ -14,7 +14,7 @@ modules: [
 		config: {
 			// See 'Configuration options' for more information.
 			location: 'Amsterdam,Netherlands',
-			appid: 'abcde12345abcde12345abcde12345ab', //openweathermap.org API key.
+			appid: 'abcde12345abcde12345abcde12345ab' //openweathermap.org API key.
 		}
 	}
 ]
--- a/modules/default/currentweather/README.md
+++ b/modules/default/currentweather/README.md
@@ -15,7 +15,7 @@ modules: [
 			// See 'Configuration options' for more information.
 			location: 'Amsterdam,Netherlands',
 			appid: 'abcde12345abcde12345abcde12345ab', //openweathermap.org API key.
-			showWindDirection: true
+			showWindDirection: true,
 		}
 	}
 ]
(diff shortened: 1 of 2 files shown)


CFenner:
Remove showWindDirection from default cfg in README. 

Good idea.

MichMich:
Nice one! Before I merge: 
@CFenner: what do you mean with you comment?

roxasvalor:
I had added the showWindDirection variable into the default config at the top of the README. Not sure what I was thinking since none of the other config options show there except location and appid. It is fixed now,

modules/default/currentweather/README.md
```diff
@@ -14,7 +14,7 @@ modules: [
 		config: {
 			// See 'Configuration options' for more information.
 			location: 'Amsterdam,Netherlands',
-			appid: 'abcde12345abcde12345abcde12345ab' //openweathermap.org API key.
+			appid: 'abcde12345abcde12345abcde12345ab', //openweathermap.org API key.
```
MichMich: Could you remove the trailing comma?

MichMich:
Just one question (a stupid one): doesn't the cardinal direction depend on your location?

roxasvalor:
Depends on what map you're looking at. The conversion from degrees to the cardinal direction assumes that we're using the north pole in the Arctic as "north" and the south pole in Antarctica as "south". If you use maps with Antarctica on top it would be incorrect though unless openweathermap does any trickery behind the scenes.

CFenner:
Are the openweatherAPI degrees switched on south hemisphere?

roxasvalor:
I don't think so. It appears the degrees provided by the openweatherAPI are Azimuth degrees. Azimuth degrees use True North (North Pole) as North.

So I guess the answer is no. It will not switch your direction but I have sent an email to the openweatherAPI team asking them if this is truly the case.

MichMich:
@roxasvalor could you take a look at https://github.com/MichMich/MagicMirror/blob/v2-beta/modules/default/currentweather/currentweather.js#L313 and https://github.com/MichMich/MagicMirror/blob/v2-beta/modules/default/currentweather/currentweather.js#L315 ? they both say ESE.

roxasvalor:
Thanks, I found it. Check PR #308 for the fix.

PR #247, opened 2016-04-26
boldClock

nhubbard:
Make the minutes on the clock bold. Now with extra verified commits. :+1:


nhubbard:
Just realized that it made write conflicts. I'll be remaking this pull request.

PR #253, opened 2016-04-26
calendar events repeating count title for birthdayd etc

momo13292:
Adding a title with count of the repating of the event. With options for showing titles and setting the title for each calendar in the config.


MichMich:
Could you explain a little bit better what this does? I don't understand the usage. 
Also make sure you remove the nohup file.

bitte-ein-bit:
Looking at the code I guess it shows an instance count behind repeating
events in one calendar. So you enter have to enter a birthday on 27th of
April 1967 in the calendar.
 
With this patch it would show:
 
originalTitleOfEvent (49. Birthday)
 
if you add Birthday as repeatingCountTitle to the config.
 
Just guessing.... Didn't try

nhubbard:
Based on my look-over, @bitte-ein-bit is pretty accurate.

momo13292:
The usage is like @bitte-ein-bit explained. I think it's really useful to know which Birthday a Person has. Could be used for all kinds of events (annyversaries, birthdays, and everything what happens every year). Facebook Shows a notice like this in the calendar as well.

bitte-ein-bit:
IMHO it's limited to the use case of having only anniversaries /
birthday in one calendar. If I've read the code right as soon as a
calendar has mixed entries types it won't work anymore. I guess overall
it won't hurt and I see a use case.
 
Overall I'd prefer a solution which is able to parse contacts using
vcards, carddav and iCloud to generate a birthday and anniversary iCal
feed. I've done a iCloud module which is able to server birthdays so
far. I'll do an vCard version afterwards. The preferred carddav solution
seems to be more problematic, since I can't find a working carddav lib
for node. This solution has the charm that one doesn't have to manually
create all events if one's address book is properly maintained.

nhubbard:
We should probably merge this, @MichMich. We're beginning to look like this:

MichMich:
@nhubbard: lol! 😂
Still waiting for @momo13292 to remove the nohup file from the pr.

momo13292:
Sorry, removed the file. Do I have to start a new pull request?

PR #260, opened 2016-04-27
clockBold + Remove Custom CSS

nhubbard:
Don't worry, I have a reason. Git would not ignore my changes, so I removed it. **It will be re-added in a second pull request.**


MichMich:
Could you explain the use of this pull request? It makes the minutes bold? Why would that be preferred?
Also, because the custom.css is re-added, it might overwrite the custom.css file people modified.

nhubbard:
It looks a lot cooler, and it, in a controlled test that I did, was preferred by 29 of 30 people surveyed. (The last person was uncooperative.)

CFenner:
Why don't you add this as an option? Other users might like it the way it is.

MichMich:
Isn't it easier to just make a custom change in your custom.css? I agree with @CFenner. Most users will just want the current style.

djsunrise19:
As an option would be great, I think.

nhubbard:
Okay... I'll attempt this. Thanks for the feedback, guys!

PR #265, opened 2016-04-28
clockBold's Re-Incarnation

nhubbard:
This should be a better version of what some of #179 intended to do and all of what #260 intended to do.


MichMich:
There should be a way to do this without custom.css. Readding it wil overwrite customized files.

nhubbard:
There should... I'll revert my changes on my repo and re-add this pull request.

PR #286, opened 2016-05-04
Added option for using the Beaufort scale or raw wind speed

roxasvalor:
I wanted to see the exact units for wind speed so I made an option. The default is to use the Beaufort scale.


nhubbard:
Ooh, didn't even think of this! Thanks!

PR #299, opened 2016-05-07
Calendar Time Display updates

KirAsh4:
Please ignore the File Mode changes ... something weird happened when I synced. I blame the gremlins. Not sure how to ignore those mode changes.

Anyway: calendar.js has options to use either relative or absolute time display. In addition, when using `absolute`, the user has the option to set an `urgency` time frame that will force those events within that period to be displayed as relative. So events that are say in the next 5 days are displayed as `in x days` while the ones over 5 days away, are displayed with absolute dates.

README.md has been updated accordingly.


KirAsh4:
I'm revoking this as I found another error (that I created.) I'll address that in the morning.

PR #304, opened 2016-05-10
V2 beta [Calendar Module]: Time display options

KirAsh4:
Added the ability to choose between an absolute or relative time display. While in absolute mode, the user can also set an urgency time frame (from now) so that events within that time frame will be displayed as relative while everything else is in absolute time. See this post for an example:
https://forum.magicmirror.builders/topic/5/calendar-event-dates-enhancement/3#


MichMich:
Could you try removing the file mode changes from this PR?

KirAsh4:
That's how I get them from GitHub ... I tried both ways, by not touching the files and by changing them manually. For some reason, GitHub keeps messing them up.

KirAsh4:
Technically those files shouldn't be +x anyway ... but that's how I'm getting them from GH.

MichMich:
Do these changes have an effect if I don't change my config? With other words: is it a breaking change? Or are all changes optional?

KirAsh4:
Doesn't break anything. It's running fine on my end.

MichMich:
Sorry, I mean does it change the current default behaviour? Are all changes optional?

KirAsh4:
Oh yes, everything is optional. The current defaults are 'relative' and the urgency is turned off. HOWEVER, I just realized you pushed an update for the time problem last night. I do NOT have that yet on my end. If you want me to sync first and resubmit, I can do that.

MichMich:
Those changes don't seem to conflict.

PR #305, opened 2016-05-11
Update translations.js

Wilco89:
added fy.json


CFenner:
The translation file is missing.

PR #321, opened 2016-05-12
Fetch the latest Node.js without any 3rd party interferences

jordi2830:
This will get the latest Node.js from the selected branch.
I do recommend only using a stable and tested version as newer ones might break MM.
Branch can be changed by simply editing the **NODE_STABLE_BRANCH** variable.
Currently set to _6.x_ but can easily be changed to _4.x, 5.x_ or anything else desired.
(This will still need to be manually edited if you wish to upgrade to a newer Node.js version in the future (e.g. _7.x_, but it is less frequent than doing so for each minor release)

By using this setup script, Node.js will be downloaded from the official repository provided by NodeSource.

I've also fixed some typo's in the installer script.

### DIFF
--- a/installers/raspberry.sh
+++ b/installers/raspberry.sh
@@ -18,7 +18,7 @@ echo '                       \$$$$$$  |'
 echo '                        \______/'
 echo -e "\e[0m"
 
-# Define the tested version of node. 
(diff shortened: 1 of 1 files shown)


MichMich:
Thanks!

PR #325, opened 2016-05-13
Italian translation

iosonopersia:
Fixed the previous translation. Does the code of the modules take its translated strings from the MagicMirror/translations directory?


MichMich:
Thanks!

PR #335, opened 2016-05-19
Norwegian translations

RemiHansen:
Removed my earlier pull request, as I used the wrong short name for Norwegian Bokmål (nb), and also added Norwegian Nynorsk (nn). Nb/nn is the correct usage in the config file as well, so you get the day/month translated.


MichMich:
Thanks!

PR #346, opened 2016-05-29
Location feature for calendar module

KilleKat:
Adding the possibility to show also the location for an event. If
Location is set it will be shown in the next line else a '/' will be
shown. There is also the ability to set a max length to the string.
Default config is false for the location feature.

### DIFF
--- a/modules/default/calendar/calendar.css
+++ b/modules/default/calendar/calendar.css
@@ -20,3 +20,7 @@
   padding-left: 30px;
   text-align: right;
 }
+
+.calendar .location {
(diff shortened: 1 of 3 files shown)


MichMich:
Thanks!

One thing: why not just show nothing when no location is set? Displaying a `/` doesn't make much sense to me.

KilleKat:
I have change it. Also there was a littlug bug in the calendarfetcher.js

MichMich:
This PR has merge issues. Could you look at that? Also, it seems it still uses the `/` in your code.

PR #349, opened 2016-06-01
Added polish translation

matt08:
Added polish translation and references.


MichMich:
Thanks for your contribution. Unfortunately this PR has conflicts. Could you fix them? Also note that a "NE" reference should be added to the wind directions. (I need to add them to the other translations as well. It seems it is missing.

PR #372, opened 2016-07-10
Analog clock addition

KirAsh4:
Full addition of analog clock to the default one. This adds the ability to have either a digital, analog, or both displayed. The analog clock can also be customized with different faces for background. The included faces are in SVG format for size scaling, whether the clock is defined as 150 pixels or 600 pixels, the face will remain sharp and crisp.


MichMich:
Thanks man!

KirAsh4:
Ja ja. As long as I'm not too busy, or dead, I write code. :)

PR #375, opened 2016-07-17
Calendar updates - recurrence handling and monthly view

mbalfour:
The primary fix is to the handling of recurring events to make them show up correctly.  There were multiple bugs that caused recurring events to break with exception dates and specific recurrence overrides.

There is also an addition of an optional monthly calendar view that shows calendar events on a monthly display instead of a list format.


MichMich:
Could you please add some screenshots of the list format? To this conversation?

mbalfour:
image

The list format is the existing (unchanged) calendar view.  Monthly is the new view.  Here's a screenshot with both, with the startDate set to 12/03/2015 for a slightly more interesting screenshot.

NOTE:  You can't currently use both simultaneously very easily like in this pic though, since the underlying calendarFetcher uses the same maxEntries and numDays across the views.  That's something I'll look at separately a little later. :)

MichMich:
I appreciate your effort, but currently I think your calendar overview isn't generic enough to implement into the default calendar module. Main contributor @KirAsh4 is already working on some other implementations. Feel free to join in on: https://forum.magicmirror.builders/topic/82/weekly-calendar/35

I suggest you modify you calendar to be used as a 3rd party module, and add it to the 3rd-party module list: https://github.com/MichMich/MagicMirror/wiki/MagicMirror²-Modules - This way others can still benefit from your work! 

It would be great if you could send a new PR with only the recurrence handling bug fixes. 

Thanks!

PR #381, opened 2016-07-25
Brazilian Portuguese Translation 

diego-vieira:
Added Brazilian Portuguese Translation


MichMich:
Thanks for your contribution! Unfortunatly I can't commit it since it seems to remove the Greek translation from the translations.js file. Could you fix this and resubmit?

diego-vieira:
@MichMich sorry about that, should be good now.

MichMich:
Perfect! Thanks!

Issue #388, opened 2016-08-01
Language default of 'en' in config.js results in a blank screen


causztic:
Hello,

I've experienced what other users have encountered over here https://forum.magicmirror.builders/topic/412/blank-black-screen-noob/18

Basically having the option of 'en' (which is the default in the config.js.sample) results in a blank screen.  An empty string defaults it properly to English and works. Other language options works too.

MichMich:
Thanks for reporting. I'll investigate on the forum what is causing this issue. As soon as Its certain that this is a bug, I'll continue in this github issue.

MrHarcombe:
FWIW, I'm helping my son try and get this working (with a voice control module, by @dr4ke616) and we hit similar problems... but it's only for modules which use the translations part of the system - for us, initially, we just removed the default "alert" module and all seemed to be working (but any third party module that used the transations, caused the system to fail).

MichMich:
Interesting. I'll look into that.

MichMich:
@MrHarcombe is there a particular module you were using that caused the issues?

MrHarcombe:
As I say, the default "alert" module made things stop working. Having a default language of 'en' worked with the other default modules.... but I think "alert" is the only one that uses translations, isn't it?

MichMich:
I tried it with the alert module and the `en` language setting, and it seems to work fine here. Do you see any particular error in the developer tools console panel?

MrHarcombe:
The following config fails (browser [console log](https://github.com/MichMich/MagicMirror/files/394732/magicmirror-mrharcombe.c9users.io-1470083874259.txt) is attached...)
```
/* Magic Mirror Config Sample
 *
 * By Michael Teeuw http://michaelteeuw.nl
 * MIT Licensed.
 */

var config = {
	port: 8080,

	language: 'en',
	timeFormat: 24,
	units: 'metric',

	modules: [
		{
			module: 'alert',
		},
		{
			module: 'clock',
			position: 'top_left',
			config: {
				displayType: 'analog',
				analogFace: 'face-001',
			}
		},
		{
			module: 'calendar',
			header: 'UK Holidays',
			position: 'top_right',
			config: {
				calendars: [
					{
						symbol: 'calendar-check-o ',
						url: 'webcal://www.calendarlabs.com/templates/ical/UK-Holidays.ics'
					}
				]
			}
		},
/*
		{
			module: 'currentweather',
			position: 'top_right',
			config: {
				location: 'New York',
[...]

PR #390, opened 2016-08-03
Translations issue

MrHarcombe:
Regarding issue #388, I think this code handles the various permutations of translations/fallback translations and duplicate files.

If the `else if` isn't used then when the translationFile  and translationsFallbackFile are the same, then MM actually loads only the translationsFallbackFile which can cause issues later on in modules that rely on having a translation entry present, and not a tranlsationFallback entry.


MrHarcombe:
Hm, this isn't how I'd hoped it would all go - being a bit of a git novice is a pain. I just wanted to fork your "develop" branch and then submit my one change...

MichMich:
Your Git work is ok. I received just the desired change. But the change itself does not do what needs to be done. More info here: https://github.com/MichMich/MagicMirror/commit/18ae08de8f90f84dc168c9e71c3fbdf877a8a644#commitcomment-18511041

MichMich:
No, that's all be taken care of by the translator. Are you still experiencing issues? If so, please describe the issue in the original issue-post: https://github.com/MichMich/MagicMirror/issues/388

Thanks!

MrHarcombe:
It's all good - I think the author of the MMM-Voice-Control module wasn't using the singleton Translator correctly, as he was directly accessing Translator.translations instead of going through Translator.translate. This, of course, meant that in our case where the module loaded only the fallback translation then their code failed to retrieve the necessary values.

I'm going to put a PR against their module, once I'm sure I've got it working properly :)

MichMich:
Awesome! Thanks for your contribution!

Issue #399, opened 2016-08-15
[Feature] Humidity for Current Weather Module


amcolash:
Hello, I love to see the weather with my magic mirror. It has been pretty humid lately where I live and so I decided to add in humidity to the Current Weather Module. It is really a simple addition as the open weather map api already has that data provided.

I was wondering what I should do merge my additions into the code base (as I think others might like this option as well - it is configurable). Should I just make a pull request? Thanks.

MichMich:
Hi @amcolash, that would be a nice addition indeed! Feel free to send a pull request!

PR #403, opened 2016-08-18
Current humidity for the Current Weather module

amcolash:
This will add in current humidity information as discussed in #399.

P.S. sorry my editor seemed to screw up the indentation, just noticed!


MichMich:
Thanks! Is it possible for you to fix the indention in this commit? And could you share a screenshot what it would look like? Thanks for your contribution!

amcolash:
Yuck white space is a nightmare across project - every time ;)
I fixed everything up, it should look ok when viewing all changes together.

Here is a screenshot - it is not pretty, but quite useful
screenshot from 2016-08-18 09-43-27

MichMich:
Nice work. My suggestion is to use an Icon in stead of the word "Humidity" since this would require a lot of translations.

amcolash:
When I get some time in the next few days I will look into icons. I think something like this might be nice: 

image

I can make something similar on my own (so no copyright problems, references, etc).

MichMich:
Looks awesome! :)

jwk4946:
This free icon set includes humidity icons http://weathericons.io/ if that helps

amcolash:
I was just looking at that site before I checked this ticket! Thanks for the pointer, I will just stick to that one for now then to keep things similar.

amcolash:
Alright, should be all set to go! Here is a screenshot. Thanks for the support everyone - this is one of the first times I have done a pull request to a public repo.

untitled

MichMich:
Looks Nice. The only thing I'm not sure about is that it's on the next line. Not sure if that's the best place ... maybe ask some opinions on the forum?

amcolash:
It felt a bit heavy with all three measurements on the same line. I do agree though, it is a bit sparse

MichMich:
Note sure what to do with this PR. Did you ask for feedback on the forum? Also, since the water drop contains a % symbol. Maybe you could just do "100💧" ?

amcolash:
Hey Michael, I am sorry for the late response - I have been extremely busy lately. I did make a post on the forums today and am awaiting some creative insight!

cowboysdude:
Is this available yet?

MichMich:
@cowboysdude Not yet, we're still looking for a better location for the info. Suggestions are welcome.

cowboysdude:
@MichMich I like the icon you suggested.. and I'm thinking that it could be added to just underneath the temp... centered to avoid too much whitespace...    OR the entire module layout could be changed..

Get rid of the decimal part of the temp and just use straight temps such as;

56.9 should be just 57 and the humidity could be put to the right of that.... keeping the layout even ;)
That's my two cents worth.

MichMich:
Removing the decimal part would not work for Celsius temperatures. I rather have a exact temperature than the humidity.

amcolash:
Yeah I agree, that would be odd. I like the idea of being on the bottom -
most likely aligned to the right side to be out of the way. Hopefully this
weekend I can finally close this long pull request!

Also as a heads up, I was looking around in the data that we got by default
from our weather provider and it didn't seem like much other useful data
was around. We could add a min/max to that same line in the layout but that
gets repetitive if you have the forecast as well.

MichMich:
Merged. Made some visual changes.

screen shot 2016-10-01 at 14 08 54

cowboysdude:
That'll work :)  I am going to change and get rid of the decimal.... I know you guys use C but here I use F and it's odd to see the decimal point for us hahahahahha

MichMich:
@cowboysdude Make it configurable and send a PR. That way anyone kan benefit from your change. 👍🏻

amcolash:
If there are still questions about the layout, I made a new version. This one has the decimal and also keeps things a bit more compact. Thoughts?

screenshot from 2016-10-02 13-11-33

PR #415, opened 2016-08-26
 Added option to remove "more" tags in newsfeed description

yo-less:
Many feed descriptions end with with a link tag to the actual article
online (see http://www.deutschlandfunk.de/die-nachrichten.353.de.rss,
for example). The mirror news description looks strange with this tag in
the end.

I have added an option to remove the tag from the description and
updated the readme.md accordingly.


yo-less:
Will expand this to delete various tags from the beginning and ending of the description.

PR #427, opened 2016-09-06
currentweather module api url bugfix

pOakley:
Running this on Raspberry Pi 2, node version v6.5.0, magicmirror@2.0.0

The current url created has a typo that causes the module to only display default weather units (Kelvin). After checking the website (http://openweathermap.org/current), there is no "/" at this location in the url. Removing it fixes the issue and allows the user to see the specified units in config.js.


MichMich:
Thanks!

Issue #442, opened 2016-09-16
New feature (for developers?): Update header with optional information


Jopyth:
I like headers for my modules, but at the same time they need additional space. Therefore I think it is useful to be able to add tiny bits of information to the header, which otherwise would need extra space in the module content. 

An implementation example exists on my [fork](https://github.com/Jopyth/MagicMirror). I could create a clean pull request, with only the modifications regarding this `appendix` to the header (open for other name suggestions ;) ).

For example we can show, when the last data was pulled in the backend, or how many contacts from the address book have been loaded, small hints if an error occurs, etc.

![Examples](https://cloud.githubusercontent.com/assets/564777/18603277/58e0e46c-7c70-11e6-8f00-553ed6789098.png)

Maybe related losely to #236.

MichMich:
Yeah, this is something on my wishlist/todo list. Will probably add this feature next month.

MichMich:
Of course a PR is welcome, but it depends on the implementation. I'm a little picky for the core changes.

Jopyth:
Well then start picking on #444 when you have time. Or you can just use this as a basis for your ideas.

MichMich:
I've added this functionality. See: https://github.com/MichMich/MagicMirror/commit/b2a7d3584bad2752d59d2325c581f5018e875306

Example code:
````
getHeader: function() {
    return this.data.header + ' appendix';
}
````

Let me know if this works for you.

Jopyth:
Yes, that looks perfect. :)

Jopyth:
Example:

+````javascript
+getHeader: function() {
+	return this.data.header + ' <span class="fa fa-refresh"></span> ' + this.lastUpdated;
+}
+
+````

This one did not work for me inside a `header` element. However it worked inside a `div`.

Although I am wondering why [this post](http://stackoverflow.com/a/19779520), says that for example `span` elements should work. Maybe there was a different error. Can you confirm, that it is possible to include a font-awesome icon in the header? That would be awesome.

MichMich:
Not sure. It should return a string. What error do you experience?

MichMich:
If this really is an issue, I wouldn't mind moving to a dig container in stead of a header. Feel free to send a PR for this. Otherwise I can make an adjustment somewhere later this week.

Jopyth:
Never mind, my bad. I guess there must have been some other error. The `span` tag works just fine inside the `header` tag.

[...]

PR #444, opened 2016-09-17
Header appendix implementation

Jopyth:
* Does the pull request solve a **related** issue? #442 
* What does the pull request accomplish? 
Intial implementation of a header appendix, which can be defined by the module by overwriting getAppendix(). It is updated with updateDom().
It does not actually change any modules to use this, just enables future use.

Possible usages and example: see #442


andyipod1437:
Looking forward to trying this out. Thanks

MichMich:
I've added this functionality using a slightly different approach. See: https://github.com/MichMich/MagicMirror/commit/b2a7d3584bad2752d59d2325c581f5018e875306

Example code:
````
getHeader: function() {
    return this.data.header + ' appendix';
}
````

MichMich:
Please let me know if this works for you.

Issue #456, opened 2016-09-28
Limit access to localhost (security related)


Jopyth:
Are there any merits to the usual workflow, if we limit the access to localhost? For example change [this line](https://github.com/MichMich/MagicMirror/blob/master/js/server.js#L17) to 
`server.listen(config.port, config.route);` and make the default `route = "127.0.0.1"`.

Reason: More and more modules do very cool stuff (e.g. accessing Emails), however they also need the login information in the config. Currently anyone on your network (guests - if you do not have a separate guest network) can straight up load your mirror html page or even your `config.js` and read sensitive information from your config file.

So if I invite Eve to my house and she goes to 192.168.0.100:8080/config/config.js (if that is my Raspberry Pi IP) she gets the good old `Server not responding` or whatever instead of reading all our API keys, possible usernames and password for certain websites, etc...

There is probably a much better way, but for the moment this would be very easy to implement.

MichMich:
You make very valid point. But I think it might be a good idea to make this configurable. Since a lot of people access it from a external IP to test it. Maybe we should add whitelist IP's. 

Most important: never allow people you don't trust on your network! 😉

Jopyth:
Exactly that was my thought behind the `config.route` value. We would set this in the config file, so just like changing the port we can change the route. If this is set to "0.0.0.0" everything is like it is now, everyone can access it.

Something like a Whitelist would be awesome.

Still I think, to protect the unsuspecting user making the default (e.g. In the config.js.sample) to only allow access from localhost would be good. Plus of course documenting this change for anyone who is developing.

MichMich:
Will look into this if I have time. Feel free to send a PR.

Jopyth:
Would you prefer the simple all or nothing approach, or a more comprehensive one, for example with [express-ipfilter](https://www.npmjs.com/package/express-ipfilter)? With this package we can use an `ip_whitelist`, but it adds another dependency.

MichMich:
I would prefer the latter. MagicMirror² relies on a lot of dependencies already so that's not really an issue.

PR #457, opened 2016-09-29
Patch limited access

Jopyth:
Solves #456 as discussed there. 

Features:
- Warning message of default behaviour is logged, if `ipWhitelist` value is not specified in the `config`
- If you try access from a blacklisted IP you will see how to change it
- It logs if anyone tries to access your mirror, who is not allowed

From my PC:
screenshot from 2016-09-29 17 10 32

From my phone:
screenshot_2016-09-29-17-10-56

### DIFF
--- a/js/server.js
+++ b/js/server.js
@@ -16,10 +16,6 @@ var Server = function(config, callback) {
 	console.log("Starting server op port " + config.port + " ... ");
 
 	server.listen(config.port);
-	if (config.ipWhitelist === undefined) {
-		config.ipWhitelist = ["127.0.0.1", "::ffff:127.0.0.1"];
-		console.log("Warning: Missing value (ipWhitelist) from config.js, assuming default (localhost access only): " + config.ipWhitelist);
-	}
 
 	app.use(function(req, res, next) {
 		var result = ipfilter(config.ipWhitelist, {mode: "allow", log: false})(req, res, function(err) {
--- a/js/server.js
+++ b/js/server.js
@@ -20,7 +20,16 @@ var Server = function(config, callback) {
 		config.ipWhitelist = ['127.0.0.1', '::ffff:127.0.0.1'];
 		console.log("Warning: Missing value (ipWhitelist) from config.js, assuming default (localhost access only): " + config.ipWhitelist);
 	}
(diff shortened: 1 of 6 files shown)


js/server.js
```diff
@@ -10,11 +10,26 @@ var app = require("express")();
 var server = require("http").Server(app);
 var io = require("socket.io")(server);
 var path = require("path");
+var ipfilter = require('express-ipfilter').IpFilter;
 
 var Server = function(config, callback) {
 	console.log("Starting server op port " + config.port + " ... ");
 
 	server.listen(config.port);
+	if (config.ipWhitelist === undefined) {
```
MichMich: The defaults for the config can be set in: https://github.com/MichMich/MagicMirror/blob/develop/js/defaults.js that might be a cleaner option.

MichMich:
Awesome work! I've added a comment about the default settings. You might want to change/look into that. Also, is it possible for you to add information about this feature to the readme and add it to your PR?

js/server.js
Jopyth: Yes, that makes sense. However can we still display the warning message then?

MichMich:
Also, it seems like travis bot isn't happy:

````
  13:24  error  Strings must use doublequote  quotes
  25:52  error  Strings must use doublequote  quotes
````

You might want to check the code using `grunt`.

js/server.js
MichMich: Just check if it's set to a certain value, and then show a message ...

Jopyth:
Will take care of this. Maybe we can do it without the warning message. If someone wants to access the mirror from outside, the error message is pretty clear on how to fix it.

MichMich:
Awesome work! Thanks!

MichMich:
I just tried installing develop on my real mirror, and get `Access denied to IP address: ::1` with the default config. Any idea?

MichMich:
Never mind, fixed: https://github.com/MichMich/MagicMirror/commit/b267491934fcdc46e361f4876cf201a2c8785314

Jopyth:
I had `::1` at first (since that is wikipedia said would be the loopback address), but my IPv6 somehow displayed `::ffff:127.0.0.1`, so I changed it to that instead. Lets hope this works for everyone now.

PR #464, opened 2016-10-02
On the update procedure

Jopyth:
People [will run into errors](https://forum.magicmirror.builders/topic/745/switched-to-development-branch-mirror-stopped-launching).

This is mostly to start a discussion on how we can make the transition for people updating as smooth as possible. Is this helpful? Do we need the hint somewhere else, also?


MichMich:
Added info in commit: https://github.com/MichMich/MagicMirror/commit/0671e7d456296c357999068e20d1b1a8d1619939

PR #468, opened 2016-10-14
Add option to show rain amount

Jopyth:
This adds a boolean configuration option to show the estimated amount of rain in millimetres. This has been running on my mirror for a long time now. It is not a perfect prediction, but gives a rough idea, if there is going to be a **lot** of rain or if it is just a few drops.

Maybe someone else, also wants to see this information?

With the options set to `false` (same as before) and `true`:
screenshot from 2016-10-14 23 08 54


MichMich:
Thanks!

PR #469, opened 2016-10-15
added german translation for updatenotification

fewieden:


Jopyth:
Just saw, that `commit` or `commits` is added in the code, so `branch` is fine as a translation, that way it is at least consistent.

MichMich:
Thanks

PR #470, opened 2016-10-16
Change position from absolute to fixed

Jopyth:

No.

* What does the pull request accomplish?

Fixes visual display on chrome. Does not change anything when displaying with electron. Hopefully also does not change anything on other browsers.

With the `absolute` position of a hidden module (opacity is manually changed to 1)
other modules move into weird position because of the extra margin:
absolute
With the `fixed` position (opacity is manually changed to 1), the module itself moves out of the window:
fixed

Chrome is on newest Ubuntu version: Version 54.0.2840.59 (64-bit).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Solve an issue where module margins would appear when the first module of a section was hidden.
-- Solve visual display errors on chrome, if all modules in one of the right sections are hidden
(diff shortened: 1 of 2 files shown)


MichMich:
The modules switched to absolute so that they didn't take up any space (eg. other modules would move up). Is that still the case?

Jopyth:
Yes, that behavior stays the same.

### Both modules there:
hidemodule1
### Hide one with the changed code (using `fixed`), the other moves up:
hidemodule2
### It is still there (manually remove opacity), overflowing from the page, but the page does not automatically resize.
hidemodule3

MichMich:
Cool! Thanks.

PR #475, opened 2016-10-20
Deliver additional information in calendar broadcasts

RedNax67:
This PR ensures description, geo and location are part of single and repeating calendar events being broadcast. The info is available so should be part of the broadcast. Also https://github.com/RedNax67/MMM-TrafficCal needs this information to dynamically display commutes based on calendar entries.


MichMich:
Thanks!

PR #479, opened 2016-10-25
add support to enable set timezone for show in clock module.

roramirez:
Give possibilty to set a timezone into the clock Menu


MichMich:
Great work. Thanks you!

PR #488, opened 2016-10-27
Added capability to load compliments from a local file. 

BrianHepler:
Adds an optional capability to load the compliment arrays from a local JSON file. New variable "fileName" defaults to null. If present, it will attempt to load & parse the file. If it fails, the module will revert back to the array of compliments in the config.js file.


MichMich:
Nice work. I don't think "local file" is the right term, since it should be hosted on a webserver. Maybe use the term "remote file" in stead?

MichMich:
Hi @BrianHepler, will you make the above changes? Also note there is a conflict with the current changelog. If you fix both, I can merge your contribution. Thanks!

BrianHepler:
Sorry, I didn't see the notification until now. I'll get right on it.

PR #489, opened 2016-10-27
README Fix updateInterval for currentweather, weatherforecast:

roramirez:
Fix milliseconds in the example for the updateInterval option


MichMich:
Thanks!

PR #492, opened 2016-10-31
Mm fixes

aschulz90:
"_module.js:setConfig()_" currently overwrites "_this.defaults_" in a module. The first argument from "_Object.assign()_" is the target object too, resulting in "_this.config_" becoming the same object reference as "_this.defaults_" (https://googlechrome.github.io/samples/object-assign-es6/). This results in the modules default config values not necessarily being the true default values at runtime.

Now "_this.config_" gets assigned a new object with the merged values from the default config and the passed config argument instead. That way, "_this.defaults_" still holds the correct default values.


MichMich:
Thanks!

PR #498, opened 2016-11-04
Config: Add option disabled for modules config

roramirez:


README.md
```diff
@@ -92,6 +92,7 @@ Module configuration:
 | `position` | The location of the module in which the module will be loaded. Possible values are `top_ bar`, `top_left`, `top_center`, `top_right`, `upper_third`, `middle_center`, `lower_third`, `bottom_left`, `bottom_center`, `bottom_right`, `bottom_bar`, `fullscreen_above`, and `fullscreen_below`. This field is optional but most modules require this field to set. Check the documentation of the module for more information. Multiple modules with the same position will be ordered based on the order in the configuration file. |
 | `classes` | Additional classes which are passed to the module. The field is optional. |
 | `header` | To display a header text above the module, add the header property. This field is optional. |
+| `disabled` | Set disabled to `true` for not show the module. This field is optional. |
```
Jopyth: Just a language detail, maybe change this to:

> Set disabled to \`true\` **to** not show the module. [...]

MichMich:
Why not just comment out a module?

paviro:
Because it takes more time if you do it often. Would welcome this feature as well and it doesn't hurt or does it somehow? :/

MichMich:
In a good editor it takes one keypress. ;) 

It looks like the node modules are not disabled. Only the frontend. This should probably be Fixed. Remember that a config can include multiple instances of one module. (Of which one, more or all are disabled ...)

Jopyth:
Disabling the node helper would be useful, but we could scan all modules first, before starting the node helpers. This feature would also be great for the remote control, since commenting out a module also stop being able to read the config values.

MichMich:
In what case would your app need a config for a disabled module? Since it's impossible to enable it on the fly?

Jopyth:
Not on the fly, but I also wanted to enable editing the config on HD, and it would be handy to be able to "remove" a module (by disabling), but still keeping the data, in case you want to add it later again.

Jopyth:
Alternatively I can solve this by saving such disabled modules in a different file, which was my current idea.

roramirez:
Hello Guys, IMHO, I think is good add the flag, we'll can build the base for future to integrate a module config like @Jopyth say. I think is a good approach a new features.

MichMich:
@roramirez Are you planning on integrating the node_helper loading into this change?

roramirez:
Hi @MichMich, I can do it.

MichMich:
I close this PR. Fee free to resubmit a PR which includes the node_helper part.

PR #505, opened 2016-11-09
Added keys to EN and DA translations files

StefanBols:
Added "DAYAFTERTOMORROW" key to the EN and DA translation files. The key appears in other files. People will mainly copy from the EN translation file, therefor it's important that the EN translation is kept up-to-date.


MichMich:
Thanks!

StefanBols:
@MichMich: No. Thank YOU. Please make contact if you need to have more strings translated. This offer are also valid for all modules, including 3th party. Free as long it, or they stays open source :-)

PR #510, opened 2016-11-11
main.js: Fix hidden other modules

roramirez:
In commit e419701 is introduced updateWrapperStates for hide unused
regions.

The problem is, not all the modules loaded have the style.position =
'static', this cause the some modules will be hide.

This is reproduced for the next module config

```
    modules: [
        {
            module: "clock",
            position: "top_left"
        },
        {
            config: {
                feeds: [
                    {
                        title: "New York Times",
                        url: "http://www.nytimes.com/services/xml/rss/nyt/HomePage.xml"
                    }
                ],
                showPublishDate: true,
                showSourceTitle: true
            },
            module: "newsfeed",
            position: "bottom_bar"
        }
    ],

```

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -274,9 +274,7 @@ var MM = (function() {
 
 			var showWrapper = false;
 			Array.prototype.forEach.call(moduleWrappers, function(moduleWrapper) { 
-				if (moduleWrapper.style.position == "static") {
-					showWrapper = true;
-				} 
+				showWrapper = true;
 			});
 
 			wrapper.style.display = showWrapper ? "block" : "none";


MichMich:
This change does not what the function is intended to do. This way all regions will be shown. Modules will get a static position after the "show" method was called (which will be called on all modules initially). So this should work. 

You you encounter any issues of modules not showing?

MichMich:
See this post for more info: https://forum.magicmirror.builders/topic/445/looking-for-beta-testers/32

roramirez:
Hi @MichMich 
The issue is encountered using the configuration mentioned in PL. I'm using on test Firefox and Chrome in port binding by electron.

Config
````javascript
    modules: [
        {
            module: "clock",
            position: "top_left"
        },
        {
            config: {
                feeds: [
                    {
                        title: "New York Times",
                        url: "http://www.nytimes.com/services/xml/rss/nyt/HomePage.xml"
                    }
                ],
                showPublishDate: true,
                showSourceTitle: true
            },
            module: "newsfeed",
            position: "bottom_bar"
        }
    ],

````

First is showed the both modules, clock and newsfeed, pass 2 seconds and is hidden clock.

MichMich:
Ok, I'll look into it this weekend.

MichMich:
I just push a fix which should fix your issue. The behaviour isn't as it should yet, but at least your bug is gone. I continue to improve it over the weekend: https://github.com/MichMich/MagicMirror/commit/e311f170625799c2d987f2b9d07157c0d8eb441f

MichMich:
This should be fixed now. Let me know if you run into issues.

roramirez:
Oh, great!. I did think do this way the fix, I mean checking the style == "" , I did dont  know for the some modules can be have this attribute on empty

Thanks!

PR #513, opened 2016-11-13
Mm fixes

aschulz90:
The default config definition doesn't conform with the JSON specification (http://www.json.org/ , http://jsonlint.com/). 

The issues were:

- value names must be strings
- strings are enclosed in double quotes
- no trailing commas

I am currently developing a module, where i modify the config file and i need to be able to parse the config JSON-Object with JSON.parse(...). But that fails, because its not valid JSON (when config.js is based on the default config). I will probably add a workaround for this into my module, but it would be good, if the default config file uses the standard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Solve an issue where module margins would appear when the first module of a section was hidden.
-- Solved visual display errors on chrome, if all modules in one of the right sections are hidden
+- Solved visual display errors on chrome, if all modules in one of the right sections are hidden.
+- Module default config values are no longer modified when setting config values.
 
 ## [2.0.5] - 2016-09-20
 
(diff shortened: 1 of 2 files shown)


roramirez:
Hi @aschulz90 
The config statement is a Javascript dictionary instead of a JSON file definition.

Jopyth:
As @roramirez says, it is not a JSON file, or were you suggesting to switch to a JSON file configuration, instead of a javascript file?

In my opinion, one of the neat things, we can do in the `config.js`, is multiplying time (sometimes the unit is *ms*), such as 30 * 60 * 1000, which is easier to recognize as half an hour, instead of 1800000. This would not be possible in a JSON file, and there is probably some other things, too.

MichMich:
More importantly: it allows us to add closures (anonymous function) to the config in the future. So I won't be switching to JSON. (Especially since son doesn't support commenting.)

aschulz90:
No, it's not a functionality change but a formatting one. I want the config definition to be easily be parsable when externally being extracted from config.js.

My usecase is that i extract the config with a Regex and try to parse it, which fails, because of the issues stated above.

The change doesn't modify any functionality from MagicMirror.

MichMich:
@aschulz90 I understand what you wan't. But forcing JSON rules upon a javascript dictionary doesn't make sense. Especially since in the future, the config might even contain even more exotic content.

Maybe you could load the config as a javascript file and generate the JSON yourself?

PR #514, opened 2016-11-14
logger: Show line number where is called the function for the log

roramirez:


MichMich:
Did you check it in Chrome as well? I recently made a similar change, and Chrome did not like it.

roramirez:
Hi @MichMich !,  Yes, it's work

I add the shot of log into Console In Chrome
screenshot at 2016-11-14 12-39-58

MichMich:
Thanks!

PR #518, opened 2016-11-16
Module Updates

qistoph:
* Check if updates are available for non-default modules

The UI will only show one of the available updates currently. If that's an issue for you, please let me know your wishes regarding displaying multiple available updates.

Translations are only included for NL and EN.

### DIFF
--- a/modules/default/updatenotification/node_helper.js
+++ b/modules/default/updatenotification/node_helper.js
@@ -1,4 +1,7 @@
-var simpleGit = require("simple-git")(__dirname + "/../..");
+var SimpleGit = require("simple-git");
+var simpleGits = [];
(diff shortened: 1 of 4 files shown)


paviro:
Nice 👍🏼

MichMich:
Awesome work. Thanks!

MichMich:
I'm currently getting this message:
screen shot 2016-11-17 at 11 48 19

Of course we can exclude the .AppleDouble, but it might be a better idea to only check registered (in config) modules. This way unused modules won't give a message.

Is that something you can look into?

qistoph:
Yeah sure, good idea. I'll give it a shot

PR #524, opened 2016-11-18
Check updates for configured modules only

qistoph:
Fixes the request in #518 to only check for updates for configured modules.
Also checks if module has a .git folder, to skip modules installed in a different way.


MichMich:
Thanks!

PR #536, opened 2016-11-22
Log client messages to server log too

qistoph:
When troubleshooting issues I find it very useful if all logging is in one place. Having some messages in my browser window (or hidden when using electron) and others on the node server makes it harder to find bugs.

Would you be interested in a solution to log client messages to the server's log (too)?
This PR is just a PoC. If you have ideas on improving this (before mering), I'd love to hear and work on them.


MichMich:
Hi, thanks for your contribution. I need to think about it if this is something I currently want to merge. (Since there is a lot of other work we need to do on the logging part ...) Before we discuss this, I was just wondering why the jsnlog is added to the vendor folder if it's also loaded via package.json. Is there a special reason for this?

qistoph:
I understand if you need to think about it and even if you don't want to merge it (at this time or indefinitely). It was just something that is helping me out in testing and that I wanted to share.

The jsnlog.js is required at the server as well as at the client. Copying it to a location available to the client was the most obvious way to make this work. Do you see a better way?

roramirez:
Hi everybody!

Like as say @MichMich there a lot the work to do on the logging subsystem. I have some ideas and can help some stuff.

I think is best way set a flag into the system to log into the server side, like `debug`. Can be into the config.js. For other way, sometime is need add the timestamp into the serverside, because if you using pm2 or other method for run like as service, is usefull have the time and date of events.

MichMich:
I think the most important first step is to make the logger part of the module instance. This way you can log messages like: 
````
this.logger.error('an error occurred');
````
Rusulting in:
> 2016-11-23 09:05:23 MyModule: An error occurred.

This way, not only the logs contain the sender (module name) of the log message. But also are we able to enable/disable logging on a module to module basis. I thing the best config approach would be to add a `debug` property to all the module configs. Which is false by default. Additionally we can add a property to the main config section which is a object: 
```
debug {
    core: false,
    module: true
}
````
This way you can quickly enable logging for all modules and enable/disbale logging for the core messages.

The node helper should work on the same way. With the same instance property. Although it's a bit tricky since there is only one node helper for multiple module instances. In my opinion the best approach would be to enable logging for the node_helper if one or more of the instances has logging enabled.

When we have this setup in place, we can work on a way of sending logging messages from the front to the backend. The proposed jsnlog uses it's own communication system. Which, to my opinion, is a bit unnessecery, since we already have the front and backend communication in place. 

Maybe it's a good idea to discuss these idea a bit further on the forum. Or if someone feels the need to send a PR with above changes: don't let me stop you! :)

It might be a good idea though to wait with the merge into develop. Since this would push back the release of the current changes in develop. I wan't to release those somewhere coming week. 

Any thoughts?

**Comment by MichMich** [comment]
A little addition to the story above: maybe someone knows a way of making the above changes without the need of modifying all the modules. Maybe we can reference the `Log` variable to the module's instance logger. This way, no changes to the module are needed.

**Comment by roramirez** [comment]
Good points. I think because Log subsytem is not a little fix in design and code not be include in the next release.  

In forum is a good place for have input to the users on their needs it.  For other hand, in idea to set the config like 
```
debug {
    core: false,
    module: true
}

```

This can be work in cascade in every module, if is setted `debug: false` into the module use this configuration.

MichMich:
I close this PR for now. Please feel free to resubmit a PR when we have a newly setup solution as discussed above. Your contribution is more than welcome! Thanks!

PR #541, opened 2016-11-26
Add option to set bind address

roramirez:


roramirez:
@CatoAntonsen Its included into the CHANGELOG update because in different Pull Request I'm missing added. Was a mistake.

PR #543, opened 2016-11-26
clock: remove statement of repetead code.

roramirez:


modules/default/clock/clock.js
```diff
@@ -107,12 +107,6 @@ Module.register("clock",{
 		if (this.config.showPeriod && this.config.timeFormat !== 24) {
 			timeWrapper.appendChild(periodWrapper);
 		}
-		if (this.config.displaySeconds) {
```
MichMich: Are you sure this part must go too?

Issue #544, opened 2016-11-28
Customizable Timezone


VikramTiwari:
Please only submit reproducible issues. 

The time is derived using `moment-with-locale` which uses system default time. This results in proper time during development, but as soon as this code is uploaded to raspberry pi the system default (GMT) kicks in, and hence incorrect (to end user) time.

To fix this, we can use 'moment-timezone' which allows us to use moment and specify timezones as well.

nhubbard:
@MichMich I believe we can close this issue.

PR #545, opened 2016-11-28
fix timezone issue #544

VikramTiwari:
patch for: https://github.com/MichMich/MagicMirror/issues/544
- added moment-timezone file which is combination of moment-timezone and moment-with-locale
- updated entry for moment in vendor.js
- added timezone function to all moment functions in clock module

ToDo:
- update moment functions everywhere else in the defualt modules

### DIFF
--- a/modules/default/clock/clock.js
+++ b/modules/default/clock/clock.js
@@ -16,6 +16,7 @@ Module.register("clock",{
 		showPeriodUpper: false,
 		clockBold: false,
 		showDate: true,
+    timeZone: 'Americas/Los_Angeles',
 
 		/* specific to the analog clock */
 		analogSize: '200px',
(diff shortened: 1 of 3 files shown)


roramirez:
Hi @VikramTiwari ,

Your PL is backwards the HEAD of develop branch, so there some conflicts.

I thinks is not a good idea set by default a determinated location for timezone, not all people have the same timezone. 

Please, check if the issue still into the HEAD of develop branch.

roramirez:
... also, somethings I forget mention, the feature to set a timezone was add into the develop branch.

MichMich:
Also, we use a custom version of moment.js due to an annoying buffer overflow bug in the 12hour notation. So it's critical this change doesn't get overwritten.

VikramTiwari:
I think there's enough work already been done on the issue and there are a lot of extra considerations to be done before implementing an easy-going strategy as my PR. If I had more time, I would have gone through it properly. But hopefully this issues will be resolved in next release.

Feel free to close this issue and related PR.

PR #547, opened 2016-11-30
Hide private

dbahn25:
This simple code change adds a new option to the calendar module: `hidePrivate`. It prevents all events that have the iCal attribute `class:PRIVATE` from being shown in the calendar module.


MichMich:
Unfortunately your PR has some conflict. Also I'm curious what happens when an iCal feed doesn't support the `class` parameter.

dbahn25:
I think most of the conflicts are from the README and Changelog.
I have not yet tested it with iCal feeds without the `class` parameter, but the events in my iCalendar feed that are not classified as `PRIVATE` simply don't have the `class:` parameter (I opened the ical file in a text editor and the line `class:XYZ` simply doesn't exist for non-private events). For those events the class property that is checked by MM is `undefined` and the item is shown anyway.
So for feeds that don't support `class:`, it's probably going to be `undefined` for all events...

MichMich:
It seems this PR has some conflicts. If you can fix this I can merge the PR.

MichMich:
Thanks!

dbahn25:
Since the modules folder is listed in `.gitignore`, updates to the modules won't be applied when a user regularly updates mm, right? So this change is only going to be in new installations (except if one manually updates the modules folder)?

Jopyth:
@dbahn25 If what you say was true, you could not have commited your changes the way you did. So fortunately all is fine, the files are in the index already.

Also look at the [.gitignore](https://github.com/MichMich/MagicMirror/blob/master/.gitignore#L60) again, everything in the default modules is not being ignored.

dbahn25:
All right, I didn't have a closer look at the gitignore file before now. It was just that git didn't want to stage the files without `-f` and said something about gitignore. 
Thanks!

PR #548, opened 2016-12-01
Add Guide Javascript MagicMirror² Code style

roramirez:

### DIFF
--- a/.github/JAVASCRIPT_CODESTYLE.md
+++ b/.github/JAVASCRIPT_CODESTYLE.md
@@ -1,0 +1,563 @@
+# MagicMirror² Javascript Code Guide
+
+This is a guide for writing consistent and aesthetically pleasing MagicMirror
+JavaScript code.
+
+This guide is based created by [Felix Geisendörfer](http://felixge.de/) and is
+licensed under the [CC BY-SA 3.0](http://creativecommons.org/licenses/by-sa/3.0/)
+license.
+
+![Creative Commons License](http://i.creativecommons.org/l/by-sa/3.0/88x31.png)
+
+## Table of contents
+
+### Formatting
+* [Tabs for indentation](#tabs-for-indentation)
+* [Newlines](#newlines)
+* [No trailing whitespace](#no-trailing-whitespace)
+* [Use Semicolons](#use-semicolons)
+* [80 characters per line](#80-characters-per-line)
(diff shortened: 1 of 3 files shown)


.github/JAVASCRIPT_CODESTYLE.md
```diff
@@ -0,0 +1,563 @@
...
+[the opposition]: http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding
+[hnsemicolons]: http://news.ycombinator.com/item?id=1547647
+
+### 80 characters per line
+
+Limit your lines to 80 characters. Yes, screens have gotten much bigger over the
+last few years, but your brain has not. Use the additional room for split screen,
+your editor supports that, right?
+
+### Use single quotes
```
nhubbard: I would tend to disagree with this. We currently have ESLint set up to yell at us if we use single quotes, even if they are within another string.

.github/JAVASCRIPT_CODESTYLE.md
```diff
@@ -0,0 +1,563 @@
...
+```js
+var a = 0;
+if (a == '') {
+  console.log('losing');
+}
+```
+
+[comparisonoperators]: https://developer.mozilla.org/en/JavaScript/Reference/Operators/Comparison_Operators
+
+### Use multi-line ternary operator
```
nhubbard: This would disagree with our vendor files and some files of our own. We use inline ternary operators, not to be difficult, but because everyone uses it that way.

.github/JAVASCRIPT_CODESTYLE.md
```diff
@@ -0,0 +1,563 @@
...
+```js
+setTimeout(function() {
+  client.connect(function() {
+    console.log('losing');
+  });
+}, 1000);
+```
+
+
+### Method chaining
```
nhubbard: I also am a bit weary about this. It would make large modifications on other files necessary.

MichMich:
I agree with nhubbard. I'm not a big fan of enforcing all kind of rules. (Which I usually don't stick to myself either). I like the automatic check if ESLint, but that is as far as we should go, I think.

More importantly. NO ONE reads these kinds of texts. You don't want to know how many PR's and ISSUES I receive in which they didn't read the requirements.

roramirez:
@MichMich Ok.

PR #550, opened 2016-12-02
Fixed exception for modules without a position

aschulz90:
To reproduce the issue, simply add 

{
	"module": "clock"
}

as a module. You will then periodically get an exception, because the module has no content DOM. 

module_exception

I added a simple check, if the content exist, to fix that.


Jopyth:
Isn't the error coming from the module actually trying to access its own DOM? Maybe an option for modules to require an actual position would be better?

aschulz90:
The error is coming from the framework, when this.updateDom() is called in a module without a position and moduleNeedsUpdate() tries to access a non-existing DOM.

An example where it happens too and makes sense: add a calendar module without a position, because you don't want it to be shown, but still receive calendar events for another module.

MichMich:
@nhubbard Why do you think it would break other modules? Did you test it?

Jopyth:
@aschulz90 Ok, now I understand. This might be the best way to fix it then.

PR #551, opened 2016-12-02
Option to display temperatures rounded to nearest integer

olexs:
Added a config option `roundTemp` for both _currentweather_ and _weatherforecast_ modules that rounds the temperature values to the nearest integer before display. Default behaviour is unchanged (option set to <code>false</code> in defaults).

Reasoning: no-one really needs to know the temperature exact to a decimal point. The display looks much cleaner with rounded numbers, too.


olexs:
Found a way to do this a bit cleaner, will reopen once reimplemented.

olexs:
Reimplemented in a cleaner way, using the existing `roundValue` method.

MichMich:
Thanks!

PR #552, opened 2016-12-02
Option to scale the whole display using Electron's zoom factor

olexs:
Added an option to scale the whole mirror display, using Electron's `zoomFactor`. Default value is `1.0`, so default behaviour is unchanged. Allows to make everything larger (and thus more readable) on high-resolution mirror displays.


MichMich:
Nice one! Is there a way for you to fix the conflicting changelog, so I can make the merge?

olexs:
Sure, give me a sec.

MichMich:
Thanks!

PR #554, opened 2016-12-06
Auto install non-default modules

qistoph:
Another idea I've had, if you're interested.

This change allows non-default modules to be automatically installed if the git-URL is provided in the module's optional argument 'repository'.

When configuring MagicMirror on RPi, instead of my dev-machine, I would like to just copy the config.js and not have to think of installing some modules manually. See here, a solution to that.

Future improvement might be to automatically update modules, maybe depending on a setting in the configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added option `address` to set bind address.
 - Added option `onlyTemp` for currentweather module to show show only current temperature and weather icon.
 - Added option `remoteFile` to compliments module to load compliment array from filesystem.
+- Added option `repository` to modules to allow auto installation.
 
 ### Updated
 - Modified translations for Frysk.
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ Module configuration:
 | **Option** | **Description** |
 | --- | --- |
 | `module` | The name of the module. This can also contain the subfolder. Valid examples include `clock`, `default/calendar` and `custommodules/mymodule`. |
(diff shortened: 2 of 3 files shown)


Jopyth:
Since this touches an installer system: A discussion about something like this started in [this thread](https://forum.magicmirror.builders/topic/1059/something-i-m-working-on) in the forum (and in several other threads at some point). I think a few people would be interested in helping to develop this.

The reason for this is, that the install process is not streamlined across all modules (clone is often not enough), so I think an approach as simple as the one you propose, is currently not sufficient. Also it might not (yet) belong in to the core, see the quote from @MichMich in the thread mentioned:

> Make sure it's based on the module infrastructure where possible. That we you can develop and test is, without the need of directly integrating it in mm2. If It's fully up and running and tested, we can integrate it by adding it to the list of default modules.

Maybe we can use this opportunity to start a new discussion in the forum, about how to make a good installer system for the MM? This could include a similar optional option for modules, to install modules from the config automatically.

MichMich:
Yes, this would be my recommended approach. Try building one module that does what you want to accomplish. If it's streamlined an running, I can include it into the core. If any changes are nessecery to the API, let me know. Together we can built something cool, without the need of major changes to the core.

qistoph:
The auto installation of modules is converted to a module:
https://github.com/qistoph/MMM-autoinstall

Down side now is, of course, that a module (this one) is still required to be installed manually. I do understand though that proper testing would be required before integrating it as a default module.

After the git clone, an npm install is also run.

Please let me know if you have any suggestions or ideas.

PR #555, opened 2016-12-07
Check updates (in updatenotification) only for repos with remote

qistoph:
A remote repository is not always available for git repos. Developping new modules locally and storing changes in git for example. This PR fixes the updatenotifcation module to only check for updates if the git repo actually has a remote to fetch from.


Jopyth:
Very nice, I wanted to do this, too, but was too lazy. :+1:

MichMich:
Thanks!

PR #556, opened 2016-12-07
Add async callback to module when loaded

qistoph:
To enable the [autoinstall-module](https://github.com/qistoph/MMM-autoinstall) to install modules before they are loaded it was necessary to modify MagicMirror. This change provides an async function (loaded) with callback-function to modules. It is called when the module is loaded, before any other modules are loaded. The next module is loaded only after the callback function is called. This enables modules to do async operations (i.e. git-fetch and npm install) and then continue the loading of other modules afterwards.

If you see a better solution to delay loading modules, I'd be happy to adapt the autoinstall-module to that.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -66,7 +66,7 @@ var App = function() {
 	 *
 	 * argument module string - The name of the module (including subpath).
 	 */
-	var loadModule = function(module) {
+	var loadModule = function(module, callback) {
 
 		var elements = module.split("/");
 		var moduleName = elements[elements.length - 1];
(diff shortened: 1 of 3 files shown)


MichMich:
Hi @qistoph,

I understand the need for this. Since it's a big change that might cause issues which are not tested yet, I want to delay the merge of this PR until the current develop branche is pushed to master.

My aim is to do this somewhere this month.

qistoph:
Rebased the branch on the current develop branch.
Updated modules/README.md to mention "Introduced in version: 2.1.1.", instead of 2.1.0.

MichMich:
Thanks, could you also modify the changelog?

PR #558, opened 2016-12-09
Size for compliments text

roramirez:


MichMich:
maybe it makes more sense to user the standard "classes" property for this. an let people overwrite all classes (so for example also the `light` class.) Just like the hello world module?

roramirez:
@MichMich done!

MichMich:
Perfect! Thanks!

PR #560, opened 2016-12-10
make electronOptions configurable

berlincount:


js/electron.js
```diff
@@ -31,6 +31,8 @@ function createWindow() {
 		}
 	}
 
+	var electronOptions = Object.assign({}, electronOptionsDefaults, config.electronOptions);
```
olexs: Maybe move this after line 41, so that user-defined options override `kiosk`, `fullscreen` and `autohidemenubar`? Otherwise it isn't transparent that those specific options will be overridden by code.

berlincount:
Updated. I'm conflicted in how I feel about kioskmode in general, it seems unnecessarily magic.

js/electron.js
```diff
@@ -38,6 +38,8 @@ function createWindow() {
 		electronOptions.autoHideMenuBar = true;
 	}
 
+	var electronOptions = Object.assign({}, electronOptionsDefaults, config.electronOptions);
```
olexs: This is broken, because the above if-else sets properties on `electronOptions`, but only `electronOptionsDefaults` and `config.electronOptions` are merged here.

berlincount: agreed.

berlincount:
did I mention I dislike the way kioskmode is done?

roramirez:
@berlincount Yep. When after this change will be merged could be a new change to introduce how you think is better.

berlincount:
@roramirez I'd just remove it alltogether, and would use electronOptions to set the corresponding options directly when needed.

roramirez:
Well, if we'll remove the option we will break the backward compatibily with old versions  in the next release. 

Your change it seems good and great to me at now. cc @MichMich

MichMich:
This seems like a good solution. Backward compatibility is nessecery, but we might want to remove the old way from the documentation. So others will only use the new way. You might want to add some commenting in de source to explain why it's there. 

If these changes are made, I can merge the PR.

berlincount:
Like this?

MichMich:
Yeah, but if I'm not mistaken, the readme should also be updated.

berlincount:
happy? ;)

olexs:
I like this. Leave backwards compatibility in place, mark as deprecated and remove from readme, introduce a new more flexible configuration by user. Could be later extended by adding a module that warns the user when starting MM with deprecated options in config.

MichMich:
Perfect! :) Thanks!

PR #563, opened 2016-12-12
Mm fixes

### LINKED ISSUE #562: Differences between modules default values in the readme and the script file in branch develop
There are some discrepancies between documented default values in the readme and the actual script file for some modules in the develop-branch (maybe master too, haven't checked):

### calendar:

Config Value | Readme | Script-File
------------- | -------- | ----------
Urgency | 0 | 7
titleReplace | None | `{"De verjaardag van ": "",	"'s birthday": ""}`

### alert

Config Value | Readme | Script-File
------------- | -------- | ----------
display_time | 3.5 (says in seconds) | 3500 (comment says in seconds)
welcome_message | Welcome, start was successfull! | false

### currentweather

Config Value | Readme | Script-File
------------- | -------- | ----------
location | "New York" | ""
updateInterval | 300000 (10 minutes) **(actually equals 5 minutes)** | 600000
animationSpeed | 2000 | 1000

### newsfeed

Config Value | Readme | Script-File
------------- | -------- | ----------
feeds | `[ { title: "New York Times", url: "http://www.nytimes.com/services/xml/rss/nyt/HomePage.xml", } ]` |  `{title: "New York Times",	url: "http://www.nytimes.com/services/xml/rss/nyt/HomePage.xml", encoding: "UTF-8" }]`
animationSpeed | 2000 | 2500

### weatherforecast

Config Value | Readme | Script-File
------------- | -------- | ----------
location | "New York" | ""
updateInterval | 300000 (10 minutes) **(actually equals 5 minutes)** | 600000
animationSpeed | 2000 | 1000
initialLoadDelay | 0 | 2500
weatherEndpoint | variable name is `'weatherEndpoint'` | variable name is `'forecastEndpoint'`

I think that should be addressed before 2.1, but I can't say , which one is actually the correct one in every case (most likely just outdated documentation). Therefore i haven't made the changes myself and created no PR, but opened an issue instead.

aschulz90:
Fixes #562 

I have updated the READMEs with the values defined in the script-files, EXCEPT for the calendar module, where `titleReplace` was defined as this:

```
titleReplace: {
    "De verjaardag van ": "",
    "'s birthday": ""
}
```

I don't think that makes sense, that this is the default value for it (in respect to the default defined calendar) and changed it to an empty object (`{}`) instead.

### DIFF
--- a/modules/default/calendar/README.md
+++ b/modules/default/calendar/README.md
@@ -144,7 +144,7 @@ The following properties can be configured:
 			<td>When using a timeFormat of <code>absolute</code>, the <code>urgency</code> setting allows you to display events within a specific time frame as <code>relative</code>
 			    This allows events within a certain time frame to be displayed as relative (in xx days) while others are displayed as absolute dates<br>
 				<br><b>Possible values:</b> a positive integer representing the number of days for which you want a relative date, for example <code>7</code> (for 7 days)<br>
-				<br><b>Default value:</b> <code>0</code> (disabled)
+				<br><b>Default value:</b> <code>7</code>
 			</td>
 		</tr>
 		<tr>
--- a/modules/default/calendar/README.md
+++ b/modules/default/calendar/README.md
@@ -109,6 +109,13 @@ The following properties can be configured:
 				<code>
 					titleReplace: {'Birthday of ' : '', 'foo':'bar'}
 				</code>
+				<br><b>Default value:</b> 
+				<code>
+				{
+					"De verjaardag van ": "",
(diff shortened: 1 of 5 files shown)


MichMich:
This will break the default behaviour for many MM users. I don't see the harm in leaving it the way it currently is.

olexs:
README updates are definitely useful. I agree with MichMich on changing the `titleReplace` default value though.

aschulz90:
I reverted titleReplace and added the documentation for it.

MichMich:
Thanks!

PR #569, opened 2016-12-14
Added console warning for deprecated config options

olexs:
Inspired by #560. Adds a list of deprecated config options, and shows a console warning on startup when any of them are set in user's `config.js`. List initially only includes `kioskmode`, since it is marked as deprecated in code and `electronOptions` should be used instead.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -42,6 +42,7 @@ var App = function() {
 		try {
 			fs.accessSync(configFilename, fs.F_OK);
 			var c = require(configFilename);
+			checkDeprecatedOptions(c);
 			var config = Object.assign(defaults, c);
 			callback(config);
 		} catch (e) {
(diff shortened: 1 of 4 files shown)


js/app.js
```diff
@@ -58,6 +59,21 @@ var App = function() {
 		}
 	};
 
+	var checkDeprecatedOptions = function(userConfig) {
+		var deprecatedOptions = require(__dirname + "/../config/deprecated.js");
```
roramirez: Why not used into js directory?. I think the config dir is more touched by the users and they can be replace the content inside it. If remove the `deprecated.js` file can be raise a error. For me have more sense, using _deprecated_ into the core directory, may be in the future we can be put functions or other stuff to deprecated not just configuration directives.

olexs: I thought about it. But since for now _deprecated_ is limited purely to config options, and the main _js_ folder is quite full as it is, I decided to put the file into the config directory. It can always be moved later, if more than just deprecated config options are added or other reasons arise. I am open to discussion on this, of course.

MichMich:
As discussed: please put the file in the js folder. The config folder is for config files only. 
Thanks.

MichMich:
Since it seems no-one is working on this PR anymore, I close it for now. Feel free to reopen or resubmit. Your contribution is more than welcome!

roramirez:
Hey Guys,

There someone interest follow in this. I want take this for build other thing.

PR #572, opened 2016-12-17
overwrite config.js with config.json

ItayXD:
I added an option to overwrite settings set on `config.js` with settings set on `config.json`, so if a module wants to overwrite settings (I'm working on an settings panel for the MM), all the module has to do is insert the new settings to `config.json`.
I also run `git update-index --assume-unchanged css/custom.css` so `custom.css` is kept, but changes are untracked.

### DIFF
--- a/config/config.js.sample
+++ b/config/config.js.sample
@@ -74,4 +74,11 @@ var config = {
 };
 
 /*************** DO NOT EDIT THE LINE BELOW ***************/
+try {
+	var json = require("./config.json");
+} catch (e) {
+	console.log('Could not load config.json');
+	console.log(e);
+}
+config = Object.assign(config, json);
 if (typeof module !== 'undefined') {module.exports = config;}


roramirez:
Why dont parse your json output to Javascript dict?
Please read Pull Request #513

ItayXD:
Hi, this is actually not related, I may use js object as well and require
it if the problem is the use of json.
It is about the option to override the config writen in the file with a
module.

ItayXD:
The important part it `config = Object.assign(config, json);`  it might as well be:
`try {
+	var module = require("./configmodule.js");
+} catch (e) {
+	console.log('Could not load configmodule.js');
+	console.log(e);
+}
+config = Object.assign(config, module.config);`

ItayXD:
Sorry for it looking bad I'm sending it from my phone.

roramirez:
I understand your idea.  Multiple input configurations for the same config I dont see sense. This can be give a confusion the users, what file they have they will be touch and what is valid into a run system. The config.js is more powerfull than a JSON configuration.

I think if you use the extenal system to write configuration, it's should write a output as Javascript dict, this have more sense to me.

But thats is  IMOH, the last word has @MichMich

ItayXD:
If we are only keeping config.js it might be very hard to write new
settings to it using js.

ItayXD:
Any other idea to solve it is welcome

MichMich:
Sorry, We'll not be moving to a json file. I think @roramirez gave enough explanation on this matter.

ItayXD:
Okay I see you don't like the way this is done, how would you suggest
editing the config with a module?

Jopyth:
I am using this function: https://github.com/Jopyth/MMM-Remote-Control/blob/develop/node_helper.js#L412

PR #579, opened 2016-12-26
Add error handling to newsfeed

brosy:
Hi,
I have been having internet issues recently and i noticed that the newsfeed would not update when the internet was restored . 

To reproduce: 
1. Set reload interval to a low value (5seconds)
2. Disconnect internet 
3. Run MagicMirror
4. News feed will be stuck on loading
5. Reconnect internet and newsfeed will appear

Let me know if you want anything changed.

### DIFF
--- a/js/class.js
+++ b/js/class.js
@@ -41,8 +41,8 @@
 
 		return ret;
 	};
-        })(name, prop[name]) :
-        prop[name];
+})(name, prop[name]) :
+prop[name];
 		}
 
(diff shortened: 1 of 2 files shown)


MichMich:
Awesome work! Unfortunately your modifications don't pass the linters. Could you run grunt?

brosy:
@MichMich I have fixed the lint error let me know if you want anything else changed.

MichMich:
Thanks!

Issue #568, opened 2016-12-13
wrong age in calendar module...


chamaeleon82:
hi,

I´ve created google calendar as a birthday calendar. Birthday is created as an appointment (whole day, repeat every year). Start date is the birthday and year of birth. 
My problem is that the display of age is a year less. So instead of 35 years old only 34 is displayed (E.g. March 22th.).
![wrong_age_calendar_module](https://cloud.githubusercontent.com/assets/21147588/21152593/3f5a3426-c167-11e6-89d4-247588df58cf.jpg)

Does anyone know this problem?

MichMich:
That's odd. I suggest discussing this issue on the forum: http://forum.magicmirror.builders

henglert:
As I wrote in the forum: 

`--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -126,7 +126,7 @@ Module.register("calendar",{
                                repeatingCountTitle = this.countTitleForUrl(event.url);

                                if(repeatingCountTitle !== '') {
-                                       var thisYear = new Date().getFullYear(),
+                                       var thisYear = new Date(parseInt(event.startDate)).getFullYear(),
                                                yearDiff = thisYear - event.firstYear;

                                        repeatingCountTitle = ', '+ yearDiff + '. ' + repeatingCountTitle;
`

MichMich:
Please send  PR. Thanks!

roramirez:
Hey @MichMich  this issue need be closed.

PR #582, opened 2016-12-29
Fix repeating count of yearly calendar events

henglert:
#568  
The occurrence counter for a yearly repeating event e was calculated by "current year" - "e.firstYear", instead of "e.year" - "e.firstYear". That leads to a repeating event counter - 1 at the end of the year. E.g.
event e: first Date: 2015-01-02, yearly repetition
repeating count evaluated on 2015-12-31: 2015-2015 = 0 => incorrect
repeating count evaluated on 2016-01-01: 2016-2015 = 1 => correct
Fix: repeating count has to be calculated from event date instead of current date.


MichMich:
Thanks!

PR #591, opened 2017-01-03
Update CHANGELOG for the async loaded function

qistoph:
As requested in #556, I have update CHANGELOG for the async loaded function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.1.1] - Unreleased
+- Add loaded function to modules, providing an async callback.
 
 ## [2.1.0] - 2016-12-31
 


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.1.1] - Unreleased
+- Add loaded function to modules, providing an async callback.
```
roramirez: This function should will be into the next feature release  2.2.0 not the 2.1.1 We the follow the Semantic Versioning. Is a new feature not a bug fix.

MichMich: @roramirez It will be 2.1.1, 2.2.0 will be introduced when there are breaking changes (just like 2.1.0).

MichMich:
Thanks!

CHANGELOG.md
roramirez: @MichMich its ok but this its not follow the [Semantic Versioning](http://semver.org/). This a new feature.

What happend if we need release a new version with only bugfixes? The version name will 2.1.0.1? Will the first 2 always be maintained?

PR #592, opened 2017-01-03
Chartjs - with sample module

qistoph:
Working on a module to show data from a smart scale to show weight and fat percentage over time I find the need to include chart.js.

I could of course include just in my module and load it from getScripts(). Since it might be pretty handy for other modules too I would like to offer this PR.

- Add Chart.js as a vendor script, like moment.js
- Add default sample module to show the easiest use of the module
- I am working on a chart sample that uses dynamic updates too, but think that's a bit too complex for a default module. Have look at the repo https://www.github.com/qistoph/MMM-dynchart


modules/default/testchart/dynchart.js
```diff
@@ -0,0 +1,175 @@
...
+
+	// Override dom generator.
+	getDom: function() {
+		var wrapper = document.createElement("div");
+		this.ctx = document.createElement("canvas");
+		wrapper.appendChild(this.ctx);
+
+		this.myChart = new Chart(this.ctx, this.config.chartjs);
+
+		/*this.myChart = new Chart(this.ctx, {
```
roramirez: You need remove these comments lines, I dont see any sense add into the sample module.

qistoph: The dynchart is still a work in progress. Those comments will of course be removed. 

The sample included can be used as is to show static data, just like helloworld can be used to show a static message.

MichMich:
I love the charts, but this way it doesn't make sense to add it to the default modules. I think it would be a better idea to keep it as an example repository. 

Nice work though!

qistoph:
Would you be willing to add the Chart.js to the vendor scripts though? Without the default module.
That way there's no need for each module to load its own Chart.js which reduces complexity, collisions and overhead.

A different approach would be to have one module to load the Chart.js and then separate modules that use it. Which is also a bit complex imho.

nhubbard:
@qistoph I think you should resubmit your pull request with just Chart.js as a vendor addition. Otherwise, good work.

PS: this looks great!

roramirez:
There any way to add Chart.js into the packages names instead of vendor?

qistoph:
@roramirez how about not putting it in the vendor folder in git, but adding the dependency in the package.json and add a post-install script to copy it to the vendor folder:

````json
{
  "scripts": {
    "postinstall": "mkdir -p vendor/chartjs; cp node_modules/chart.js/dist/Chart.min.js vendor/chartjs/Chart.js"
  }
}
````

roramirez:
See nice. 

For me, IMHO: Have files Javascript JS into vendor script is noise. Unless is necessary if the script is patched with some bugfix or some things like that.

If we can find a clean method to add in manifiest package.json file will be better. Keep external libs versionated into the code base is something I think is need it avoid.

MichMich:
I agree with @roramirez. If you have additional questions feel free to reopen.

PR #594, opened 2017-01-04
Enabled newsfeed default module for gesture events from https://github.com/thobach/MMM-Gestures

thobach:
Enabled newsfeed default module for gesture events from https://github.com/thobach/MMM-Gestures

The newsfeed can be controlled by gestures now from a gesture sensor (see https://github.com/thobach/MMM-Gestures). This allows to show a news description upon an UP gesture and display the full news article upon another UP gesture. With a DOWN gesture the normal news ticker appears and scrolls through the titles. With a LEFT or RIGHT gesture you can manually scroll through the news titles.

Demo on: http://blog.thomas-bachmann.com/2016/02/magic-mirror-2-0-mit-gestensteuerung.html (German) / https://www.youtube.com/watch?v=V8a6UgeoYOU

### DIFF
--- a/modules/default/newsfeed/newsfeed.js
+++ b/modules/default/newsfeed/newsfeed.js
@@ -152,21 +152,21 @@ Module.register("newsfeed",{
 				}
 
 			}
-			
+
 			if(!this.config.showFullArticle){
 				var title = document.createElement("div");
 				title.className = "bright medium light";
 				title.innerHTML = this.newsItems[this.activeItem].title;
 				wrapper.appendChild(title);
 			}
-				
+
 			if (this.config.showDescription) {
 				var description = document.createElement("div");
 				description.className = "small light";
 				description.innerHTML = this.newsItems[this.activeItem].description;
 				wrapper.appendChild(description);
 			}
-			
+
 			if (this.config.showFullArticle) {
 				var fullArticle = document.createElement("iframe");
 				fullArticle.className = "";
@@ -176,8 +176,8 @@ Module.register("newsfeed",{
 				fullArticle.src = this.newsItems[this.activeItem].url;
 				wrapper.appendChild(fullArticle);
 			}
-			
-			
+
+
 
 		} else {
 			wrapper.innerHTML = this.translate("LOADING");
@@ -287,14 +287,14 @@ Module.register("newsfeed",{
 	capitalizeFirstLetter: function(string) {
 		return string.charAt(0).toUpperCase() + string.slice(1);
 	},
-	
+
 	notificationReceived: function(notification, payload, sender) {
 		Log.info(this.name + " - received event");
-		if(notification == 'GESTURE'){
+		if(notification == "GESTURE"){
 			Log.info(this.name + " - received gesture");
 			var gesture = payload.gesture;
 			// actually RIGHT, because gesture sensor is built in upside down
-			if(gesture == 'LEFT'){
+			if(gesture == "LEFT"){
 				Log.info(this.name + " - received right");
 				var before = this.activeItem;
 				this.activeItem++;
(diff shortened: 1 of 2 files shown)


fewieden:
in my opinion, the default module should stay as generic as possible, for this i would have rather a notification received for show article next and previous and keep all your gestures interpretation in your module, so other modules could use this generic functionality as well

nhubbard:
@fewieden @thobach I'm in the middle on this. If @MichMich thinks it's good, then I will approve.

roramirez:
An integration from MagicMirror to outside module this is not generic and I think is not way we need to take.

Have more sense to me add the Gestual feature to MagicMirror instead this.

thobach:
I made the notification mechanism more generic and suited towards news specific events. Thanks for all your suggestions to make this code cleaner and less coupled.

@roramirez Do you want https://github.com/thobach/MMM-Gestures integrated into MagicMirror itself? I'd consider this an option for future development, but think it's out of scope for this pull request. Thoughts?

Are there any objections to merge this change?

modules/default/newsfeed/newsfeed.js
```diff
@@ -273,5 +288,63 @@ Module.register("newsfeed",{
...
+			this.config.showDescription = false;
+			this.config.showFullArticle = false;
+			if(!timer){
+				this.scheduleUpdateInterval();
+			}
+			Log.info(this.name + " - going from article #" + before + " to #" + this.activeItem + " (of " + this.newsItems.length + ")");
+			this.updateDom(100);
+		}
+		// received "more details" the first time, so showing article summary
+		else if(notification == "ARTICLE_MORE_DETAILS" && !this.config.showDescription){
```
fewieden: this should manage the two if statements at once
```
else if(notification == "ARTICLE_MORE_DETAILS"){
	this.config.showDescription = !this.config.showDescription;
	this.config.showFullArticle = !this.config.showDescription;
	clearInterval(timer);
	timer = null;
	Log.info(this.name + " - showing " + this.config.showDescription ? "article description" : "full article");
	this.updateDom(100);
}
```

modules/default/newsfeed/newsfeed.js
```diff
@@ -273,5 +288,63 @@ Module.register("newsfeed",{
...
 
+	notificationReceived: function(notification, payload, sender) {
+		Log.info(this.name + " - received notification: " + notification);
+		if(notification == "ARTICLE_NEXT"){
+			var before = this.activeItem;
+			this.activeItem++;
+			if (this.activeItem >= this.newsItems.length) {
+				this.activeItem = 0;
+			}
+			this.config.showDescription = false;
```
fewieden: the following lines repeat also for article previous and article less details, how about a reaset helper function which does it?

modules/default/newsfeed/newsfeed.js
```diff
@@ -164,6 +167,18 @@ Module.register("newsfeed",{
 				wrapper.appendChild(description);
 			}
 
+			if (this.config.showFullArticle) {
+				var fullArticle = document.createElement("iframe");
+				fullArticle.className = "";
+				fullArticle.style.width = "100%";
+				fullArticle.style.height = "1735px";
```
fewieden: what happens on screens with less pixels?

thobach:
@fewieden Thanks for your review. I addressed all your suggestions - very nice ones ;)

Are there any objections to merge this change?

MichMich:
I'm willing to add the features to the default module since the implementation is generic (using the notification system). But the required notifications need to be documented. Could you add more info to the module's readme?

MichMich:
Also, please send your PR's to develop. not to master.

PR #599, opened 2017-01-05
Pass pointer-events on fullscreen overlay to underlaying modules

qistoph:
The fullscreen_above overlay prevents mouse actions like _hover_ to be passes to modules.

1. Make fullscreen regions (based on CSS class) not handle mouse events.
2. Make all children in fullscreen regions handle mouse events, to negate any possible side-effects of 1.

Just 1. would fix the issue I was having where my module did not receive _hover_ events. Adding 2. makes sure that modules placed in a fullscreen region still **do** receive their mouse events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.1.1] - Unreleased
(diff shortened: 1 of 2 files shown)


MichMich:
Not sure if this is a good idea. What if some module uses the fullscreen_above and wants interaction?

qistoph:
That's exactly what 2. is for 🙂

MichMich:
Whoops, missed that. Thanks!

PR #604, opened 2017-01-09
Provide example of syntax for electronOptions

aaronaxvig:
It took me a while to figure out the correct syntax so thought it might be useful to include a quick example.


MichMich:
Please send your PR's to the develop branch.

I like the change though! :)

PR #605, opened 2017-01-10
Add pm2 configuration for manager MagicMirror in script installer

roramirez:


qistoph:
Imho installing and configuring pm2 should be optional. 

If it's installed though, the message explaining how to start MM could (should?) be change too and use pm2 instead of `npm start`. 

Just my 2 cents

MichMich:
I agree with qistoph. This should be optional.

roramirez:
@MichMich @qistoph  add request  in the prompt for script installer?

PR #601, opened 2017-01-06
Russian translations2

morozgrafix:
Redo of the PR #598

**Added:**
- Russian Translations

**Fixed:**
- corrected .gitignore rules for default modules

(Reasoning behind change: when I added translation for `alert` module `git add` was telling me that `.gitignore` rule excludes files and folders under `modules/default`)

I also noticed that `.gitignore:56` has following declaration:

```
# Don't ignore the node_helper nore module.
!/modules/node_helper
!/modules/node_helper/**
```
but I don't think this directory structure exists today (also there is a small typo)

PR #607, opened 2017-01-10
Fix gitignore to not ignore default modules folder

qistoph:
`/modules/**` makes git ignore all directories and files in `/modules`, so it's not even looking in `/modules/default` for changes. `!/modules/default/**` is therefore never used.

By adding the new line `!/modules/default`, git is told to look in the `/modules/default` folder for changes and will find all sub files/dirs and match these on `!/modules/default/**`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ Temporary Items
 
 # Ignore all modules except the default modules.
 /modules/**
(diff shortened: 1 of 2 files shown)


morozgrafix:
I think I've added similar change in PR #601

PR #616, opened 2017-01-12
Add apt-get update in raspberry.sh script

alexbarcelo:
Just tried this script and realized that it is likely that it will fail because the apt index is not up to date.

The trivial fix seems to be to add an `apt-get update` before the package installation. It should be harmless also.

> Don't forget to add the change to CHANGELOG.md.

It is worth doing so? Not sure how to change it, or if it needs change at all. Shame on me :(


alexbarcelo:
Ok, seems non-trivial because just found the apt-get update... multiple ones... will double-check the expected behaviour. It failed for me the first time, also the second one (after the first manual `apt-get update`) and then a third time "worked" and did multiple updates (???)

Sorry for the noise!

installers/raspberry.sh
```diff
@@ -36,6 +36,10 @@ fi
 function version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
 function command_exists () { type "$1" &> /dev/null ;}
 
+# Update before first apt-get
+echo -e "\e[96mUpdating packages ...\e[90m"
+sudo apt-get update || exit
```
roramirez: Is not neccesarry exit here.

alexbarcelo: You are right. Just copied the next-line `install`, but it is conceivable the scenario where the packages are installed, the `update` may fail but the overall `install` process may work.

roramirez: Better prevent the fail by update in repository.  With simple apt-get update It's enough

PR #617, opened 2017-01-12
Add `apt-get update` in raspberry.sh script

alexbarcelo:
This PR is related to the (closed) #616, but there were some mishandled, my bad. EDIT: Ok, I am useless and a new pull request was indeed not needed. Sorry.

It includes the comments addressed by @roramirez. It basically performs an `apt-get update`prior to any and all `apt-get` operation (to avoid a failure under non up-to-date installations, which can be common).

Disclaimer: the execution of the installer scripts triggers extra `update`s, I don't know if they should be addressed as a whole, or left alone given the modularity of the installation. I was planning to drop the PR, but @roramirez comments made me reconsider it.


MichMich:
Cool, thanks!

PR #623, opened 2017-01-16
Init Tests

roramirez:


MichMich:
Does this mean you're going to work on a full test suite?

roramirez:
Hi @MichMich 

I've interest a work in a test suite for MagicMirror. I think is somethings we need.

I can do develop somes tests. This patch propose the use mocha for the test suite.

I would like to hear comments and if  we can work a list set of test to do.

MichMich:
Honestly, I don't have much experience with writing test. This would be a nice project to learn more.

roramirez:
We can invite to users in forums. I think is really important define the test list for will do. The test suite is a little big work, isn't something we can do it in one day or week.

qistoph:
One of the changes in 3a8d72d introduces an error in the client.

[js/module.js:419](https://github.com/MichMich/MagicMirror/commit/3a8d72db313e62729dbb6f9061415ff615b7ba0d#diff-ea3e1f30bd2d56ae767ccdfcd88fba8eR419)

`Uncaught ReferenceError: exports is not defined
    at module.js:419`

roramirez:
I see. Let me check how prevent this. If you have any suggest or aprouch is welcome :)

js/module.js
```diff
@@ -415,3 +415,7 @@ Module.register = function (name, moduleDefinition) {
 	Log.log("Module registered: " + name);
 	Module.definitions[name] = moduleDefinition;
 };
+
+exports._test = {
```
nhubbard: I think that you would do this instead:
```javascript
module.exports = {
    cmpVersions: cmpVersions
}
```

roramirez: This isn't work :(. I sent a new Pull Request with a hotfix. Can you review @nhubbard and @qistoph this?

Also, I add a post for create a set list of test for the suite. 
https://forum.magicmirror.builders/topic/1456/test-suite-for-magicmirror
I would really like have your comments.

qistoph: PR #633 removed the error for me and `npm test` works. Seems like a good solution to me.

Also noticed a minor textual improvement in `tests/functions/compare-version.js`:
"Test function ... **into** ..." could better be "Test function ... **in** ...".

roramirez: Thanks. I pushed the textual change.

PR #625, opened 2017-01-17
Replace HTML tables with Markdown tables

tosti007:
Replace the HTML tables with Markdown tables for all the README.md files in all the default modules


MichMich:
Awesome. Could you add an updated changelog file?

tosti007:
I didnt really change anything, only changed html into markdown, should i add that?

tosti007:
A sidenote: For the weather modules it wasn't possible to put the whole default icon table into the list, so I added it underneath the table instead

MichMich:
Yeah, than peace can see all the effort that is put in ... ;)

tosti007:
Fair enough, but I just found out that I didn't use the developer branch as main branch. Any idea how I can get the new changelog file without messing up this whole pull request? (yes i am a github rookie)

tosti007:
If I just edit a random line in the changelog of the master branch it won't go well with the edits of the developer branch

PR #652, opened 2017-01-25
Electron issues

roramirez:


MichMich:
Thanks! Doesn't this mean we need to update some documentation and the end message of the installer?

roramirez:
No, is not necessary. The instructions in installer message and the documentation are still working.

PR #653, opened 2017-01-25
Set configuration file by enviroment variable

roramirez:
Related proposal  by @qistoph in test suite
https://forum.magicmirror.builders/topic/1456/test-suite-for-magicmirror/7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Option to use RegExp in Calendar's titleReplace.
 - Hungarian Translation.
 - Icelandic Translation.
+- Enable ability to set configuration file by the enviroment variable called MM_CONFIG_FILE.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


roramirez:
Yes, I did  test before ;) and is working. I just remove console.log introduced.

qistoph:
Ah of course, the good old function-scope... Keeps tricking my mind, probably because of coding all kinds of languages and most having block-scope ([more details](http://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html)) 

To prevent this kind of misinterpretation, I would suggest to use this form:
````javascript
var x;
if(something) {
  x = 1;
} else {
  x = 2;
}
````

**Comment by roramirez** [comment]
Hi @qistoph 

This way seems more sense to me
```javascript
var configFilename = path.resolve(global.root_path + "/config/config.js");
if (typeof(global.configuration_file) === 'undefined' ) {
	var configFilename = path.resolve(global.configuration_file);
}
```
Let me know what you think

qistoph:
Yup, agreed

PR #659, opened 2017-01-28
fix module hidden status

Jopyth:
Did not create an issues for this, since I do not think it is necessary.

Previously the module status was changed before actually showing the module, and if lock strings prevented it from being shown, the `module.hidden` value was incorrectly `false`.


MichMich:
Thanks!

PR #668, opened 2017-01-30
missing comma in sv.json translation

Jopyth:


roramirez:
We should will check by grunt to prevent this cases. I've see it pass before few problems like this.

PR #669, opened 2017-01-30
Test e2e

roramirez:


morozgrafix:
This is a great start. I was wondering if it would be a good idea to have `_spec` in the test file names for easy identification, something like `test_name_spec.js`, or `_test` to be `test_name_test.js`

MichMich:
You guys doing awesome work! :)

PR #670, opened 2017-01-30
Translation check

roramirez:
As I mention on Pull Request #668 is need prevent errors like this. This patch add translation files to be check by jsonlint.


modules/default/alert/translations/ru.json
```diff
@@ -1,4 +1,4 @@
 {
   "sysTitle": "MagicMirror Уведомление",
-  "welcome": "Добро пожаловать, старт был успешным!""
+  "welcome": "Добро пожаловать, старт был успешным!"
```
morozgrafix: Ooops. My bad! 😞

PR #679, opened 2017-02-02
Update pt.json

EdgardosReis:
pt translation updated

### DIFF
--- a/translations/pt.json
+++ b/translations/pt.json
@@ -1,14 +1,11 @@
 {
-	/* GENERAL */
 	"LOADING": "A carregar &hellip;",
 
-	/* CALENDAR */
 	"TODAY": "Hoje",
 	"TOMORROW": "Amanhã",
 	"RUNNING": "Termina em",
 	"EMPTY": "Sem eventos programados.",
 
-	/* WEATHER */
 	"N": "N",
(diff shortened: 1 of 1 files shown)


morozgrafix:
@EdgardosReis you may want to pull latest develop branch and make your changes based on that.

MichMich:
@roramirez There is a function somewhere in MM that allows commenting in JSON files for translation purposes (to make clear what it's for). It's a bit inconvenient, that this is no longer possible. Is there a way to solve this?

roramirez:
Yes there a function to remove the comments in js/translations.js is called stripComments. For standar JSON not should containt comments. The function preparse the file before load like json.

https://www.ietf.org/rfc/rfc4627.txt
https://www.ietf.org/rfc/rfc7159.txt

The problem is for validated the json translations by Grunt not can containt comments. Should be a valid JSON file.

This point is interesting and savory to discution about what we can do. Maybe @CFenner and @nhubbard can pushed about this topic.

nhubbard:
@EdgardosReis @MichMich @roramirez See [here](https://github.com/jashkenas/docco/pull/260#issuecomment-32467210) for another point that JSON shouldn't contain comments.

PR #688, opened 2017-02-06
New Compliments category: `anytime`

Nosrac:
Added support for Compliments category: `anytime`.  Compliments inside `anytime` can be shown at any time.


MichMich:
Nice one! Thanks!

PR #689, opened 2017-02-06
add test parts of day for compliments module

roramirez:

### DIFF
--- a/tests/e2e/modules/compliments_spec.js
+++ b/tests/e2e/modules/compliments_spec.js
@@ -1,0 +1,58 @@
+const Application = require("spectron").Application;
+const path = require("path");
+const chai = require("chai");
+const chaiAsPromised = require("chai-as-promised");
+const moment = require("../../../vendor/moment/moment-with-locales.js");
+
+var electronPath = path.join(__dirname, "../../../", "node_modules", ".bin", "electron");
+
+if (process.platform === "win32") {
+	electronPath += ".cmd";
+}
+
+var appPath = path.join(__dirname, "../../../js/electron.js");
+
+var app = new Application({
+	path: electronPath,
+	args: [appPath]
+});
+
+global.before(function () {
+	chai.should();
+	chai.use(chaiAsPromised);
+});
+
+describe("Compliments module", function () {
+	this.timeout(10000);
+
+	describe("parts of days", function() {
+		before(function() {
+			// Set config sample for use in test
+			process.env.MM_CONFIG_FILE = "tests/configs/modules/compliments/compliments_parts_day.js";
+		});
+
+		beforeEach(function (done) {
+			app.start().then(function() { done(); } );
+		});
+
+		afterEach(function (done) {
+			app.stop().then(function() { done(); });
+		});
+
+		it("shows correct compliments for part of day", function () {
+
+			var hour = moment().hour();
+			if (hour >= 3 && hour < 12) {
+				compliment = "Morning test";
+			} else if (hour >= 12 && hour < 17) {
+				compliment = "Afternoon test";
+			} else {
+				compliment = "Evening test";
+			}
+
+			return app.client.waitUntilWindowLoaded()
+				.getText(".compliments").should.eventually.equal(compliment);
+		});
+	});
+
+});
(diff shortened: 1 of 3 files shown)


morozgrafix:
LGTM 👍

tests/e2e/modules/compliments_spec.js
```diff
@@ -0,0 +1,58 @@
...
+			app.start().then(function() { done(); } );
+		});
+
+		afterEach(function (done) {
+			app.stop().then(function() { done(); });
+		});
+
+		it("shows correct compliments for part of day", function () {
+
+			var hour = moment().hour();
```
fewieden: i think this test case is bad, the result will be different based on the time the test is executed.

there should be multiple test cases for each part of the day without the use of momentjs.

MichMich:
Ik think @fewieden has a valid point. Could you look into this, @roramirez?

roramirez:
Yes, I can take a look

tests/e2e/modules/compliments_spec.js
roramirez: @fewieden Why dont use momentjs?

You mean,  create  3 test case and check if is the part of day accord reflect output module content

fewieden: exactly at least 3 test cases, there can be even more to test the boundary values

fewieden:
hey @roramirez i think you misunderstood what i meant with the momentjs part, because now you will only run one of those tests based on the time of the day as well. you have to hardcode the hour for each test so each test gets executed and you can check if its working correctly for each part of the day at once

morozgrafix:
Sorry, I'm a bit confused. My understanding that this is end to end test that makes sure that correct message is displayed in electron at runtime. Since we can't control at what time Travis will be running it initial commit was covering any of 3 combinations. (Can we force set system time on Travis to simulate specific hour? I'm kind of not sure if that's a good idea anyway). I agree that we don't need to rely on moment.js and can do tests with pure JS Date function. Also agree that splitting it into 3 tests knowing that only one will run at any given point is not ideal. Hardcoding hours wouldn't work, since MM running in electron will pickup system time. Am I missing something?

tests/e2e/modules/compliments_spec.js
```diff
@@ -0,0 +1,79 @@
+const Application = require("spectron").Application;
+const path = require("path");
+const chai = require("chai");
+const expect = chai.expect;
+const chaiAsPromised = require("chai-as-promised");
+const moment = require("../../../vendor/moment/moment-with-locales.js");
```
morozgrafix: probably no longer need moment.js

roramirez:
Force the system date to check a test I think is a wrong way.  We need create a test for expected values and behavior on a test system.

The tests for date and time is a complicated are because this run a host system and at a determinated moment.

We can create a new tests for case and workflow mention by @fewieden but these need be a new tests cases. I'm not sure if possible set a date enviroment where run the test... for this we need a method to force and set date on enviroment for test case.

tests/e2e/modules/compliments_spec.js
roramirez: Removed

MichMich:
I've merged this for now. We can always improve this if we want.

MichMich:
It seems Travis is currently having issues with these tests. @roramirez Can you look into this?

roramirez:
Yes, I can take a look.  @MichMich Can you tell me what build is failed for this test? . I've seen the Pull Request in Travis but I'dont see nothing related there.

morozgrafix:
@roramirez looks like it was hitting timouts https://travis-ci.org/MichMich/MagicMirror/jobs/199172700

and this one wasn't able to locate and element: https://travis-ci.org/MichMich/MagicMirror/jobs/199173082 (that was one of my tests. I don't think it failed in the past)

roramirez:
@morozgrafix Its so stranger, now are passing the tests

qistoph:
I've seen it on Travis before with my own branch of MM. 
Maybe it happens when their server(s) are a bit busy and the electron app is taking longer to load. 

Is it an idea to keep track (URLs) of failed builds that shouldn't fail on the Trello board, so that we can look into it later with a collection of evidence?

roramirez:
Yes, is good idea @qistoph.  Can you add this there on Trello?

qistoph:
https://trello.com/c/nIrPAN3o/23-fix-failing-tests-on-travis

roramirez:
Thanks you!

PR #697, opened 2017-02-08
unit test global.version

roramirez:


fewieden:
Does it make sense to use the exact same implementation to retrieve the version?

If somehow the version returns null or whatever the test will still pass because the values are the same

PR #704, opened 2017-02-09
Added Korean Translation

APPLEH0LIC:
Added Korean Translation


fewieden:
you should checkout why this is failing you created now multiple times a pr with the sme data without any changes the error clearly points out that you have to remove the comments from your json file to pass the check

MichMich:
It's currently failing, but not due to this PR. Will look into this.

nhubbard:
@MichMich This will happen every once in awhile when Travis runs tests. It will fail the build because Git throws back an error while attempting to fetch the pull request.

PR #711, opened 2017-02-11
Check for dev console instead of counting windows

qistoph:
Update the e2e test for the "dev" argument to open the development console.

Found a way to check if the dev console is actually open, instead of just counting the number of windows.


qistoph:
Travis is failing today, see https://github.com/MichMich/MagicMirror/issues/712
Maybe you could restart the Travis job tomorrow?

Issue #712, opened 2017-02-11
Error in clock_es_spec.js


qistoph:
Travis fails today: https://travis-ci.org/MichMich/MagicMirror/jobs/200630365

Seems like the regex is looking for "sabado" while the clock displays "sábado" (with acute accent on the first a).

PR #713, opened 2017-02-11
Remove stripComments for translation files

qistoph:
As the translation files are now proper JSON (without comments), the stripComments functions are no longer needed in the loader and in tester.

### DIFF
--- a/js/translator.js
+++ b/js/translator.js
@@ -19,90 +19,10 @@ var Translator = (function() {
 		xhr.open("GET", file, true);
 		xhr.onreadystatechange = function () {
 			if (xhr.readyState == 4 && xhr.status == "200") {
-				callback(JSON.parse(stripComments(xhr.responseText)));
+				callback(JSON.parse(xhr.responseText));
 			}
 		};
 		xhr.send(null);
-	}
-
-	/* loadJSON(str, options)
-	 * Remove any commenting from a json file so it can be parsed.
-	 *
-	 * argument str string - The string that contains json with comments.
-	 * argument opts function - Strip options.
-	 *
-	 * return the stripped string.
-	 */
-	function stripComments(str, opts) {
-		// strip comments copied from: https://github.com/sindresorhus/strip-json-comments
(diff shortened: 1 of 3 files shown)


qistoph:
Travis is failing today, see https://github.com/MichMich/MagicMirror/issues/712
Maybe you could restart the Travis job tomorrow?

roramirez:
Nice.  I suggest the function in core we need to move to deprecated state. We dont know if actual users can be have custom translations files. This is the reason why I wrote in #569 yesterday.

MichMich:
Still, I'm not happy with this change. I think translation files would highly benefit from comments. Isn't there a way to make this work?

qistoph:
Hey @MichMich, what would you think of a format like this: https://gist.github.com/qistoph/d30c044cb65f0b6eba79da9b7c8bdc9f?

MichMich:
@qistoph That wouldn't fully work, since some translations are shared between multiple modules. (as a matter of fact, all translations can be used by any module.)

The reason why I prefer comments in translation files, is that you sometimes want to explain in what context a phrase is used. (Since this could affect the translation.)

qistoph:
Hmm, I see your point, however I think that your two points are actually colliding.

If translations are context specific and require comments for proper translation, than they would be really fit for reuse by other modules (and so other contexts).

What are your other considerations in choosing JSON? Why not JS, ini or a custom CSV

JSON, in the original definition, doesn't allow comments, so there's three options
1. use proper JSON, without comments,
2. use JSON with comments and custom parsing,
3. use a different format that supports comments

Which would have your preference and why?

MichMich:
This is still something we should find a good solution for.

MichMich:
I'm tempted to just leave this function in since it doesn't hurt anything. Thoughts?

roramirez:
We need retain this function. We need deprecated before to remove, there other modules have comments yet.

MichMich:
Agree. Closing this PR.

PR #716, opened 2017-02-12
Increment timeout e2e test to 20000 ms

roramirez:
Increment to 20000ms timeouts of e2e tests.

This recommendation is mentioned by @qistoph
https://forum.magicmirror.builders/post/14358

### DIFF
--- a/tests/e2e/dev_console.js
+++ b/tests/e2e/dev_console.js
@@ -22,7 +22,7 @@ global.before(function () {
 });
 
 describe("Argument 'dev'", function () {
-	this.timeout(10000);
+	this.timeout(20000);
 
 	before(function() {
 		// Set config sample for use in test
--- a/tests/e2e/env_spec.js
+++ b/tests/e2e/env_spec.js
@@ -22,7 +22,7 @@ global.before(function () {
 });
 
 describe("Electron app environment", function () {
-	this.timeout(10000);
+	this.timeout(20000);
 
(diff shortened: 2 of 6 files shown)


roramirez:
@MichMich Can you re-run this build https://travis-ci.org/MichMich/MagicMirror/builds/200750793 ?

qistoph:
Rerun it before adding the timeout. That was enough for me last time and will give us some insight

MichMich:
Doesn't seem to be working: https://travis-ci.org/MichMich/MagicMirror/jobs/200750794

qistoph:
It's not a timeout issue in the latest log. Something else is going wrong now... Weird

`The command "eval git fetch origin +refs/pull/715/merge: " failed. Retrying, 2 of 3.`

fewieden:
this happened on some builds before

roramirez:
This error is because the branch for Pull Request is not exits anymore. When Travis gets the origin for this Pull Request cannot find it.

I think we can should be increment the timeout like @qistoph say it before. We can checking what happening this week with the  tests running on Travis.

MichMich:
I'm still getting A LOT of timeout issues with travis. There really should be a better solution than increasing the timeout ...

> Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

PR #725, opened 2017-02-19
Test anytime compliments module.

roramirez:
feature introduced 79c79146a5df730cb8e8eee1cc1008639a2c21c7


fewieden:
shouldn't be the test look like
```
{
  morning: [],
  afternoon: [],
  evening: [],
  anytime: ["Anytime here"],
}
```
so you can see that anytime really has an effect, this test should always result in "Anytime here"

roramirez:
Good point. Let's do it this way.

PR #742, opened 2017-02-28
Additional Vendor Libraries

s-cotton:
Adds a few popular JS libraries into the vendor folder.

> Please send your pull requests the develop branch.
> Don't forget to add the change to CHANGELOG.md.


MichMich:
I don't really see why these should be in the default vendor folder, since none of the default modules uses these. A third party module kan easily install these themselves.

s-cotton:
If two third party modules try to include a library, is there a mechanism in place to keep from loading the library twice?  If I use bower to install, say, jquery and jquery-cycle2, in one module, and again in another, that would cause issues for sure.  Adding in common libraries could provide the mechanism to avoid these issues.  Im fine with doing bower installations in the modules I want to create, I just want to make sure that I am not causing issues by potentially including a library twice and it would be ideal to not have to do library detection inside my getScripts function.  Let me know your thoughts

s-cotton:
Maybe having these additional libraries reach out to cdnjs instead could alleviate the concerns you have over including the actual library files in your project?

MichMich:
I'm not entirely sure, but if they have the same name, I think they won't be loaded twice.

s-cotton:
Yeah, I dug through the code last night and found that if they have the exact same cdnjs name then yes, they would be blocked from adding it a second time.  That being said, if an older module used an older version of the JS library and a newer module included a newer version of the module, they would not be blocked.  I am going to take a crack at a CDN focused getScripts method setup for modules that would allow for use of CDNJS libraries with a bit more of a check in place.  I think I can use the CDNJS API to check for library versions and ensure that even if two modules request the same library, that as long as a version is provided, we could take the lowest common version and use that.

Feel free to close this one out, I'll work on that new solution and send a PR when I have it working.

PR #744, opened 2017-03-04
Workaround for missing animation on this.show() when module is alone in region

CatoAntonsen:
When you hide a module using `this.hide()` the function `updateWrapperStates` is called. This function basically checks if **any** module (wrapper) inside each region has it's `style.position` set to `static` (or blank). If no module (wrapper) is found, then the `style.display` is set to `none` and hereby hides the region.

The problem is that this prevents the animation when `this.show()` is later called.

I've added a new parameter to `option` in `this.hide()`: `hiddenWrapperStyle` that has to be set to `static`. 

If you have a module that you later want to animate when showing, you have to explicitly hide the module using this parameter, like this:

```
var options = { hiddenWrapperStyle: 'static' };
this.hide(2000, callback, options);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add  test for compliments module for parts of day
 - Korean Translation.
 - Added console warning on startup when deprecated config options are used
-
+- Added `hiddenWrapperStyle` option for `this.hide()` as an workaround for missing animation on `this.show()` when module is alone in region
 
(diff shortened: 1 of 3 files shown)


MichMich:
Thanks for the report. It feels a bit like a hack/workaroud. I think it's better to solve the issue without the need for an extra flag.

CatoAntonsen:
Yes, I agree. Have replaced it with PR https://github.com/MichMich/MagicMirror/pull/750 that hopefully should not be *that* hacky...

PR #750, opened 2017-03-05
Fixed missing animations on this.show() when module is alone in a region

CatoAntonsen:
Fix for missing animation when using `this.show()` when module is alone in region.

Replaces previously PR https://github.com/MichMich/MagicMirror/pull/744

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -249,14 +249,14 @@ var MM = (function() {
 			updateWrapperStates();
 
 			// Waiting for DOM-changes done in updateWrapperStates before we can start the animation.
-			setTimeout(function(){
-				moduleWrapper.style.opacity = 1;
+			var dummy = moduleWrapper.parentElement.parentElement.offsetHeight; 
+			
+			moduleWrapper.style.opacity = 1;
 
-				clearTimeout(module.showHideTimer);
-				module.showHideTimer = setTimeout(function() {
-					if (typeof callback === "function") { callback(); }
-				}, speed);
-			}, 0);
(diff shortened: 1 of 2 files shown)


CatoAntonsen:
Seems to break some modules. Working on it...

CatoAntonsen:
Forget to say I think it works now without breaking any of the modules I'm tested it agains...

js/main.js
```diff
@@ -245,15 +245,18 @@ var MM = (function() {
 			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			// Restore the postition. See hideModule() for more info.
 			moduleWrapper.style.position = "static";
-			moduleWrapper.style.opacity = 1;
 
 			updateWrapperStates();
 
+			// Waiting for DOM-changes done in updateWrapperStates before we can start the animation.
+			var dummy = moduleWrapper.parentElement.parentElement.offsetHeight;
```
MichMich: What does this dummy do?

CatoAntonsen:
The key for this to work is that the browser has to calculate the offset height.  We don't really use that variable so `var dummy` *could* be skipped. I just prefer to indicate that we are calling a function and not calling a method. I might be damaged after to much PowerShell at work...

CatoAntonsen:
Another note: the offset height is not important. But that the browser has to calculate it gives us the result we want.

If you know a better way, I'll happily update the PR.

CatoAntonsen:
@MichMich Do you still think this is a hack?

MichMich:
Thanks!

PR #756, opened 2017-03-07
compliments.css

cowboysdude:
Css file for compliments module


MichMich:
This could just go into your custom css file.

PR #759, opened 2017-03-08
Add note to allow all IP addresses. ipWhitelist configuration directive.

roramirez:

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ The following properties can be configured:
 | --- | --- |
 | `port` | The port on which the MagicMirror² server will run on. The default value is `8080`. |
 | `address` | The ip address the accept connections. The  default open bind `::` is IPv6 is available or `0.0.0.0` IPv4 run on.  Example config: `192.168.10.100`. |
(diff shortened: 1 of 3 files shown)


BerndKohl:
Thumbs up! I was looking for this "feature" only recently witouth success!

fewieden:
I don't think that this hint is valuable. Developers can implement a whitelist on different ways, but people without knowledge go back and allow all ips, which is in my opinion the complete opposite what the intention to implement this feature was. Stuff like this is why botnets like mirai can overtake so many devices.

roramirez:
It's possible add extra Warning. But a lot people have [problem with ipWhitelist](https://forum.magicmirror.builders/search?term=ipwhitelist&in=titlesposts&sortBy=relevance&sortDirection=desc&showAs=posts), we keep need the things simple

Also we can adds a note to https://forum.magicmirror.builders/topic/1326/ipwhitelist-howto

fewieden:
Yes, people had a lot of trouble with this feature because it isn't explained very well in the readme. But it isn't a help either to avoid it.

MichMich:
It might be a good idea to show a working in the node console when starting with a full whitelist.

roramirez:
Add warning in console seems good for me. Also @fewieden  tell what you thing should be we'll add in README

fewieden:
I guess the link to the tutorial of @mochman is a good starting point. The link he references in it, is probably to complicated for most people and a hint that they should add as few ips as possible to the whitelist.

PR #769, opened 2017-03-11
Add in single line news items

amcolash:
This pull request is to allow for the option to disable wrapping of news items to multiple lines. For my current mirror, there is not a lot of space at the bottom and so this can help to clean up things (if desired).

This change adds in 2 new configuration options to the newsfeed module: `wrapTitle` and `wrapDescription`. Each is set to `true`, so things are exactly as they were before the pull request unless a user wants this functionality.

I have attached screenshots of an example (sorry not the exact same news article, but it gets the point across).

Single Line:
singleline

Wrapped News:
wrapped

EDIT:
One thing to note, I added in the single line style into `main.css` as I thought it might be useful for other modules to have the ability to specify wrapping without needing to have the style in different places. If that doesn't seem like it makes sense, I can change it to a module-specific style.


MichMich:
Awesome work! Is it possible for you to solve the conflicts?

amcolash:
Fixed

Issue #694, opened 2017-02-07
Midori on Pi Zero doesnt support ES6 Arrow Functions (line 51)


Mar-85:
Please only submit reproducible issues. 

**Platform**: Pi Zero - latest Midori

**Node Version**: 7.4.0

**MagicMirror Version**: Latest MM2

**Description**: Midori is the only browser that reliably and without issues/slowdown can run on the pi zero with MM2. It unfortunately doesnt support ES6 Arrow Functions and causes errors in the pm2 logs and wont load the module. I had this issue with a 3rd party module Profile Switcher and @tosti007 had kindly reverted back to the regular way of using functions and it works perfectly. 

**Steps to Reproduce**: Use Midori to try run the latest compliments, there will be an error in the log and compliments will not start.

**Additional Notes**: Wondering if there is a way to not use these functions at least until Midori is updated as there are quite a few users of the Pi Zero now.

Mar-85:
Changed the line to the following at all is back to working now.
                if (this.config.remoteFile != null) {
                        this.complimentFile(
                                function(response){
                                        this.config.compliments = JSON.parse(response);
                        });
                }

fewieden:
thats not working correctly because in the function this has a new context, you have to save the var this first  like it is in line 57

Mar-85:
oooh sorry! I had that in mine, forgot to add it in there above in the copy and paste. Thanks for noticing!

MichMich:
Interesting. I thinks it's a good point to discuss if we fully switch to ES6 or complete remove all ES6 functions to support older versions.

nhubbard:
@MichMich We could use a middle ground. Convert the project to all ES6, add Webpack with Uglify, Concat and Babel, and then run it through there.

MichMich:
@nhubbard I thought about this a while back, but it would make the project MUCH more complex. Since you end up with compiled files and source files. And while this might be solvable, it is a HUGE issue for issue for people starting with development.

Mar-85:
@MichMich @nhubbard that does sound as though it would make it quite a bit more complex. I was thinking more along the lines of removing ES6 functions or just labelling the modules to say whether they are compatible with older ES6 functions or not. At least that way those in the know would easily be able to remove them before installing...though I suppose this second way doesnt make it very end user friendly.

Suppose the answer would revolve around what users you are catering for.

MichMich:
[...]

PR #770, opened 2017-03-12
Calendar module excluded events filter

amcolash:
This is a simple filter that will hide certain calendar events from showing up on your mirror. A use case would be like hiding all birthdays from the mirror. It just does a simple `includes` check, so no fancy regex patterns. Should be useful for others!


fewieden:
depends on which way we want to go #694 includes is an es6 feature

amcolash:
Gotcha, I just did a quick search on W3Schools (I am a Java dev for the most part, where we have `containsIgnoreCase`). A quick search returns a [stackoverflow answer](http://stackoverflow.com/questions/1789945/how-to-check-whether-a-string-contains-a-substring-in-javascript) with a bunch of options we could use. `indexOf` would be outdated, but works on < es6. I have no real preference as I was just interested in simple string matching. I don't think I am the one to make that call.

modules/default/calendar/calendar.js
```diff
@@ -38,7 +38,10 @@ Module.register("calendar", {
 			"De verjaardag van ": "",
 			"'s birthday": ""
 		},
-		broadcastEvents: true
+		broadcastEvents: true,
+		excludedEvents: [
```
roramirez: Can be better is use [] instead of [""]

amcolash: ah, right. thanks for that! will add that to the pull branch

amcolash: fixed with commit 5466e1b

MichMich:
Nice work! Is it possible for you to fix the conflicts?

amcolash:
I probably messed up the CHANGELOG on one of the other pull requests because I merged them before all of the changes were taken upstream.

MichMich:
Thanks for your contributions!

Issue #753, opened 2017-03-06
Possibility to add '°C' to temperature values


BerndKohl:
It would be very much appreciated if there was a setting allowing the configuration of the displayed temperature values in modules weather forecast and current weather.

Currently only ° is display (in current weather) or nothing at all (in forecast).

It would like to have °C displayed after the temperature. e.g. 8°C or 45°F.

Speedy-one:
Just open the currentweather.js and search for 

`temperature.innerHTML =" " + thies.temperature + "&deg;";`

and add an **C** after &deg;

`temperature.innerHTML =" " + thies.temperature + "&deg;C";`

BerndKohl:
Thanks you.
But wouldn't that be overwritten when this module gets an update?
I would prefer a configurable solution...

amcolash:
See #773. Once pulled it should be in develop branch. Otherwise you can use the patch until merged.

BerndKohl:
Thanks a lot, that sounds very promising.

amcolash:
Glad to help!

PR #773, opened 2017-03-13
Add in temperature degree label to Current Weather module

amcolash:
This pull request solves #753. It simply adds a label to the temperature if the configuration option is set.


MichMich:
Guess whet I'm about to ask: could you fix the conflicts? 😂 Thanks!

amcolash:
At least all of these conflicts are with the changelog and not the code!

Issue #181, opened 2016-04-16
After updating, weather goes all wonky


Intecpsp:
After my weather updates I see this at the top:

![screen shot 2016-04-16 at 11 17 10 am](https://cloud.githubusercontent.com/assets/1164012/14581842/f2dacecc-03c4-11e6-878f-63a86d87e807.png)

The Weather Forecast looks like it is ignoring CSS (overlapping the calendar) and the header containing the wind and sunset somehow gains random text. Just moved to v2 today (reformatted SD card).

roxasvalor:
+1 I am having this problem as well.

MichMich:
Can you check which module generates the random string (by disabling all others.) and resubmit a screenshot? Thanks!

roxasvalor:
I tried disabling all but the current weather module and it doesn't seem to be the cause. I'm adding in other modules now until I find the problem one.

The weird thing is that the error shows up between the sunset image and the sunset time.

Intecpsp:
Leaving it a bit longer looks like this:

![screen shot 2016-04-16 at 1 20 27 pm](https://cloud.githubusercontent.com/assets/1164012/14582538/04a4d1a0-03d6-11e6-8fe9-fcbf8a112429.png)

I'll try turning everything else off and letting it sit a bit now.

roxasvalor:
I've narrowed it down to running the clock module and the current_weather modules at the same time.

![screenshot](https://cloud.githubusercontent.com/assets/18503539/14582607/c35891e2-03cf-11e6-8d75-c1755474c299.png)

Don't read too deep into the wind speed and direction.

Intecpsp:
I actually found it with just the 2 weather modules

![screen shot 2016-04-16 at 1 39 07 pm](https://cloud.githubusercontent.com/assets/1164012/14582630/bbe5a66c-03d8-11e6-9711-b18f595996b4.png)

Intecpsp:
Leaving it alone for a bit longer it goes back to normal.

![screen shot 2016-04-16 at 1 51 02 pm](https://cloud.githubusercontent.com/assets/1164012/14582688/51227fb0-03da-11e6-8c64-1564fadd1d63.png)

MichMich:
I just pushed an update. Could you give it a try?

Intecpsp:
I'm stuck at "Loading weather ... "

MichMich:
Sorry, stupid mistake. Please try again.

Intecpsp:
[...]

PR #775, opened 2017-03-14
Npm for vendors

roramirez:
This Pull Request change to use npm to install vendor and get out from the code base for these external libraries. 

For first work is include in this Pull Request only font-awesome.  The old process keep clean and work the same way.  Only is added a trigger when is use the command "npm install" to install too in the directory called vendor.


roramirez:
Thanks @fewieden Is already added here.

MichMich:
Nice! One issue: we have a modified momentjs version in the vendor folder due to a bug (12h formatting screwed up). So we might need to use a fork for this.

MichMich:
See https://github.com/MichMich/MagicMirror/issues/181
Not sure if this issue is still present in the most recent moment.js version.

fewieden:
I'm currently using it with a rpi3 timeFormat 12, the clock and currentweather looking normal. Can you remember the edge cases when this error occured exactly?

MichMich:
Are you using roramirez pull request? 

More info can be found in the mentioned issue.

fewieden:
yes im using his fork and branch.

the only info i could find was rpi and timeformat set to 12. Still looking good so far.

fewieden:
So I ran it for couple of hours and didn't come up with this issue so far. Would be great if others would test it as well, as it seemed to not appear for everyone in the past as well.

MichMich:
It might be fixed in new Raspbian version. I'll just approve this commit, and hope no-one starts to cry ... ;)

PR #787, opened 2017-03-20
test env requst http://localhost:8080

roramirez:
This test expect get 200 HTTP code on get request to
http://localhost:8080


nhubbard:
@MichMich Can you re-run the build for Node v6? It seems a build error occurred and Travis didn't catch it.

PR #788, opened 2017-03-21
Support full screen 

flo80:
Changes / new features:
- Added meta tags for fullscreen mode on iOS 
- Add script for "click to toggle fullscreen" in standard browsers

PS: I am new to github & open source development, so I think I based this on the wrong base :(


MichMich:
Thanks! I've changed the base branch for you.
Before I commit this merge, could you explain a little bit more about what it does and what the benefits are? Maybe add some screenshots for the interaction? 

Also note that the following line:
````
<script type="text/javascript" src="js/screenfull.js"></script>
````
... should be put at the bottom of the index file.

And please add an modified CHANGELOG to your commit.

Thanks!

**Comment by flo80** [comment]
This mainly has two features:

1. For iOS devices, it enables the "add to home screen" feature, which allows running MagicMirror in full screen on an iPad. (To enable this, I added the meta tags in index.html. It should also work on Android but I can't test it since I don't have an Android device). This mainly will be relevant for users not using MagicMirror as intended but in server-only mode and use a tablet to display their dashboard. 

2. It also simplifies to put the dashboard into fullscreen mode in a desktop browser (also mainly applicable to server-only mode). The user just has to click anywhere on the dashboard and it toggles into fullscreen mode. In fullscreen mode, a click toggles back to normal mode. This is implemented by the new scripts. (Note: I had to also change the CSS to fix a graphical issue with a white border in WebKit). 

I don't think, screenshots will help as nothing visually changes. 

I am not sure if these features are relevant for a lot of users. But for my server-only setup with an iPad (and occasional desktop use), it is a nice-to-have. 

PS: I now rebased my changes to the develop branch so it should merge well.

**Comment by paviro** [comment]
```
var b = document.getElementById("fullscreen");
		document.addEventListener("click", function() {
			screenfull.toggle(b);
			window.dispatchEvent(new Event("resize"));
		});
```
This is not a good idea... There are modules with touch support, adding a click callback on the body element will break all of them!

roramirez:
Good point @paviro . So, this should be configurable by config.js and the behavior be off by default.

nhubbard:
@roramirez @paviro @MichMich We could make it an optional setting.

flo80:
Thanks for your comments. I took out the fullscreen toggle scripts and made them a separate module (see https://github.com/flo80/MMM-toggleFullscreen). The only thing left in this PR are the meta tags for iOS devices.

MichMich:
Great solution: using a separate module. Merging the meta tags ...

Thanks for your contribution!

PR #789, opened 2017-03-22
Added option to ignore old items in the news feed module

42SK:
Some RSS feeds from major news sites contain headlines that are rather old.
To prevent the news feed module from displaying outdated items over and over again, this pull request adds an `ignoreOldItems`option to the module. If this option is set to `true`, the module will only display headlines that have been published within the last 24 hours. The precise time interval can be adjusted via the `ignoreOlderThan` option.

_This PR does not solve any related issues._


MichMich:
Nice one! Thanks!

Issue #797, opened 2017-03-26
Unable to select timezone


xy86:
Not able to choose timezone. 

Tried:
- timezone: Etc/GMT+01
- timezone: Etc/UTC+02

....for Oslo - Norway.

The clock disappears whenever a timezone-value is added (besides "null" or missing parameter).

Any help appreciated :)

MichMich:
Merged. Closing this issue.

roramirez:
This issue I think need re-open, there a little thing I'm under investigation about mentioned in 
review https://github.com/MichMich/MagicMirror/pull/803#pullrequestreview-30370685

MichMich:
Reopened.

MichMich:
@roramirez Can I close this one?

MichMich:
Closed due to inactivity.

sushant96702:
Hi i have installed magic mirror recently and i wanna know that how can i change timezone i am from india 
please give me step by step instruction
any help would be appreciated

sdetweil:
the clock settings are in the base OS, not in MagicMirror

so, change your system time

sushant96702:
thanks but how do i change time format to 12 hours as on rasp pi i have 12 hr time format but time on magic mirror is still 24 hrs

sdetweil:
@sushant96702  edit the ~/MagicMirror/config/config.js and change the timeFormat to 12 from 24

```
        language: "en",
        timeFormat: 24,                   // < -- this
        units: "metric",
```

sushant96702:
i have already done this

sdetweil:
please post the config.js. note that the config is case sensitive.. timeformat is NOT the same as timeFormat

it should read
```
timeFormat: 12,
```
works for me

sushant96702:
[config.txt](https://github.com/MichMich/MagicMirror/files/4786702/config.txt)




i have turned into txt because git doesnt support .js files

sushant96702:
[config.txt](https://github.com/MichMich/MagicMirror/files/4786702/config.txt)




i have turned into txt because git doesnt support .js files

sdetweil:
@sushant96702  thanks

I put this in for my config.js, by using cp from my Downloads (no edits)
and got this result  (can't post pictures here), working as expected
https://www.dropbox.com/s/3efaky0vf4q5k3x/Screenshot%20at%202020-06-16%2008-03-40.png?dl=0

sushant96702:
[...]

PR #803, opened 2017-03-28
Fix timezone issue

johanhammar:
This PR fixes [GH-797](https://github.com/MichMich/MagicMirror/issues/797) using moment-timezone-with-data.js instead of moment-timezone.js

### DIFF
--- a/vendor/vendor.js
+++ b/vendor/vendor.js
@@ -9,7 +9,7 @@
 
 var vendor = {
 	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
-	"moment-timezone.js" : "node_modules/moment-timezone/moment-timezone.js",
+	"moment-timezone-with-data.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
 	"weather-icons.css": "node_modules/weathericons/css/weather-icons.css",
 	"weather-icons-wind.css": "node_modules/weathericons/css/weather-icons-wind.css",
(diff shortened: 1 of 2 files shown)


johanhammar:
Cleaned up the PR a bit. I think this solves GH-797 as it seems to work for me. Please test before merging or let me know (or decline the PR) if it is not correct.

MichMich:
@roramirez Does this work for you?

fewieden:
I tested it with `Europe/London` and worked for me

johanhammar:
Great. Your suggestion sounds reasonable. @MichMich Let me know if I should update the PR with the above.

johanhammar:
@roramirez Do you have any advise on how to write a test for this?

vendor/vendor.js
```diff
@@ -9,7 +9,7 @@
 
 var vendor = {
 	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
-	"moment-timezone.js" : "node_modules/moment-timezone/moment-timezone.js",
+	"moment-timezone-with-data.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
```
roramirez: This should be keep the original name moment-timezone.js We dont if other module if use the vendor called moment-timezone.js

roramirez: This change is a bug introduced into commit a579b12448a99d1ba8f4bebd1ebb8a8397f3c406

PR #810, opened 2017-04-01
Bugfix introduced into bb08fe8113e37a6a44dc3ab4ca64fcf8a3ec200e

roramirez:
This is work on Pull Request #803 and help of @johanhammar and @fewieden

### DIFF
--- a/vendor/vendor.js
+++ b/vendor/vendor.js
@@ -9,7 +9,7 @@
 
 var vendor = {
 	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
-	"moment-timezone.js" : "node_modules/moment-timezone/moment-timezone.js",
+	"moment-timezone.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
 	"weather-icons.css": "node_modules/weathericons/css/weather-icons.css",
 	"weather-icons-wind.css": "node_modules/weathericons/css/weather-icons-wind.css",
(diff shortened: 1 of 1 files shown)


johanhammar:
Great, then maybe we can close/decline #803? 

I would really like to add a test for this but am finding it hard to mock moment.js.

roramirez:
Sometimes It's really hard work with dates in Javascript ;)  

I think the good approach will be create a new pull request with test for that. I'll can help you with that.

For other hand, there are a stranger behavior I mentioned in review https://github.com/MichMich/MagicMirror/pull/803#pullrequestreview-30370685

roramirez:
Hi @johanhammar 

There something weird for this case if we use moment-timezone in the test suite. You can see more information about this in 

 * https://forum.magicmirror.builders/post/19276
 * https://forum.magicmirror.builders/post/19279

If you are interested to push on something on this will be great

johanhammar:
I'll take a look at this later this week

johanhammar:
Hi @roramirez 

There seems to be a problem with the moment module  which doesn't seem to load correctly. I don't know the details of how modules are loaded (especially when running tests). 

I've attached a screenshot showing moment not beeing loaded when running the test.
magic_mirror

roramirez:
Hi @johanhammar 

Are you setting nodeIntegration ?

For me this  is the stranger case, if I set nodeIntegration give me the same error like you show in the picture before.

PR #812, opened 2017-04-01
TL;RL cheat notes for IPv4 ipWhitelist when PI listening on IPv6

doug-kline:
Hopefully save a google or support request --  while still using IPv4 addresses on MagicMirror network, need to set full address by using ::ffff: as prefix to IP in config.  e.g.  ::ffff:192.168.5.1 and not 192.168.5.1

Need to know when to use the ::ffff: or not or if PI running IPv6 or IPv4 ... see full explanation:

https://forum.magicmirror.builders/topic/1326/ipwhitelist-howto


doug-kline:
Sorry to take your time, I should have caught the spaces.

PR #814, opened 2017-04-02
Option to wrap events in calendar

amcolash:
I have a lot of events that I would like to wrap to multiple lines. This behavior is more ideal than cutting the title early with ellipsis. I have added in this option (disabled by default). It can be enabled with `config.wrapEvents`. The option will cut whole words that fall past boundary. If the word is greater than the maxLength, it will take its own line.

Here are before/after screenshots:

**no wrap**
without_wrap

**with wrap**
with_wrap

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 ### Added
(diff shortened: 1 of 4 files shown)


cederstrom:
Sweet feature! Great work!

@MichMich Any info on when this will be released?

PR #832, opened 2017-04-06
Add missing dependency.

MichMich:


roramirez:
This will be going into develop branch? Where is need it this dependency?

MichMich:
No, into Master as well. Just did a clean install which failed because of this missing dependency.

PR #833, opened 2017-04-06
Docker - Do not replace configuration

javiersigler:
Do not replace configuration in docker entrypoint.


khassel:
Hi, same Problem with the modules. May you can add the "n" also in the unmount_modules section?
Thanks,
Karsten.

PR #850, opened 2017-04-14
Default modules set

roramirez:
Add test for modules/default/defaultmodules.js


undergroundauth:
Thanks this works
You guys are amazing!!

undergroundauth:
app is running fine but getting black screen please help

batzyn:
Ok I install everything but only I sell the clock and the calendar of the messages and the news but install the modules of email and the weather but nothing comes out on the screen is black only with clock and calendar and news that I can do?

justjim1220:
Need to see your code

batzyn:
Code change config.js but nothing happens

roramirez:
Sorry guys this is a Pull Request. You should to move the conversation another place. In case the comment is related on review of this change add here.

If need support for any troubles go to forum https://forum.magicmirror.builders/

Issue #871, opened 2017-04-24
The wind directions can be confusing in written form.


Arkkimaagi:
The abbreviations for wind directions do not work similarly in Finnish as in English.

| English         | Finnish compass direction |
|-----------------|---------------------------|
| North           | pohjoinen                 |
| Northnortheast  | _no word_                 |
| Northeast       | koillinen                 |
| Eastnortheast   | _no word_                 |
| East            | itä                       |
| Eastsoutheast   | _no word_                 |
| Southeast       | kaakko                    |
| Southsoutheast  | _no word_                 |
| South           | etelä                     |
| Southsouthwest  | _no word_                 |
| Southwest       | lounas                    |
| Westsouthwest   | _no word_                 |
| West            | länsi                     |
| Westnorthwest   | _no word_                 |
| Northwest       | luode                     |
| Northnorthwest  | _no word_                 |

At least I do not know words for the minor directions in Finnish. So the translation is a bit gibberish as it assumes that people understand shorthand for something not being used in Finland.

This can be confusing in English too, how many does actually understand that the letter N means that wind blows from North towards the observer in South. (On a normal map from up to down)

In my mirror project I have solved this by adding a feature of showing the wind direction with a little arrow pointing where the wind is blowing to. So, if you'd see `3N` normally, I'd see 3 and a small arrow pointing down, as the wind is coming from north and blowing towards the observer, towards south.

![image](https://cloud.githubusercontent.com/assets/1191667/25357795/db90aada-2947-11e7-9111-d69fd71a01d4.png)

I will make a pull request of this.

Arkkimaagi:
One happy coincidence with my mirror project is that my mirror happens to be on a wall where I can see it from direction South towards North, makes translating the arrow direction to my home extra handy.

Arkkimaagi:
It seems that this has been merged into development. So, if I understand correctly, this issue can be closed now?

Arkkimaagi:
Closing the issue as it was merged.

PR #872, opened 2017-04-24
Support showing wind direction with an arrow

Arkkimaagi:
As described in https://github.com/MichMich/MagicMirror/issues/871 this feature lets user use arrow instead of an abbreviation for wind directions. This can be especially helpful on languages like Finnish that do not have all the compass directions English has.

If the user toggles config for current weather like this: `showWindDirectionAsArrow: true` they'll see the arrow like this:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add test of match current week number on clock module with showWeek configuration.
 - Add test default modules present modules/default/defaultmodules.js.
 - Add unit test calendar_modules function capFirst.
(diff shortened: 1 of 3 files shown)


MichMich:
Thanks for this contribution!

Could you please update the readme and changelog within this PR? Also it looks like `this.config.showWindDirectionAsArrow` hasn't got a default value in the default config. This isn't strictly necessary, but helps a lot for future contributors.

Arkkimaagi:
Sure, it will take a couple of days to do so as my calendar is full right now.

Arkkimaagi:
Is this what you had in mind? Did I miss anything?  :)

PR #873, opened 2017-04-25
ADD fixme wanted where the day if sunday for test number of week

roramirez:
The test  "shows week with correct number of week of year" to match does not work on sunday. This Pull Request comment this test and add comment there labeled FIXME

### DIFF
--- a/tests/e2e/modules/clock_spec.js
+++ b/tests/e2e/modules/clock_spec.js
@@ -121,10 +121,12 @@ describe("Clock module", function () {
 		});
 
 		it("shows week with correct number of week of year", function() {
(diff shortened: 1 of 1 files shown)


roramirez:
Hi @johanhammar If you are interesting... this test does not work in Sunday.

johanhammar:
Hi @roramirez 

I think it would be better if the week could be mocked. That would minimize errors like "does not work on sundays". Any ideas on how to mock the week?

roramirez:
The problem I can see is  the current-week-number start Week at monday and moment in Sunday.

fewieden:
Should we use moment for week number in production then? I think most users rely on monday as week start.

Issue #839, opened 2017-04-09
Some calendars won't load


fewieden:
Lately a lot of users complain about that some of their calendar files wont be loaded anymore in v2.1.1

https://forum.magicmirror.builders/topic/2221/after-upgrade-v2-1-1-multiple-calendars-not-displaying
https://forum.magicmirror.builders/topic/2208/calendar-shows-no-entries-after-mm-update

coderaaron:
So, here is what I'm getting when I do `npm start dev`

```Launching application.
Create new calendar fetcher for url: https://outlook.office365.com/owa/calendar/<REDACTED>@<REDACTED>/S-1-8-117873446-3978456252-3857600756-3375899348/reachcalendar.ics - Interval: 300000
Create new news fetcher for url: http://www.nytimes.com/services/xml/rss/nyt/HomePage.xml - Interval: 300000
Whoops! There was an uncaught exception...
TypeError: curr.exdates[i].toISOString is not a function
    at Object.ical.objectHandlers.END (/home/pi/MagicMirror/modules/default/calendar/vendor/ical.js/node-ical.js:44:44)
    at Object.handleObject (/home/pi/MagicMirror/modules/default/calendar/vendor/ical.js/ical.js:267:41)
    at Object.parseICS (/home/pi/MagicMirror/modules/default/calendar/vendor/ical.js/ical.js:308:20)
    at Request._callback (/home/pi/MagicMirror/modules/default/calendar/vendor/ical.js/node-ical.js:11:24)
    at Request.self.callback (/home/pi/MagicMirror/node_modules/request/request.js:188:22)
    at emitTwo (events.js:106:13)
    at Request.emit (events.js:191:7)
    at Request.<anonymous> (/home/pi/MagicMirror/node_modules/request/request.js:1171:10)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
MagicMirror will not quit, but it might be a good idea to check why this happened. Maybe no internet connection?
If you think this really is an issue, please open an issue on GitHub: https://github.com/MichMich/MagicMirror/issues```

I will do some more investigating when I have time but I thought I'd report it in case anyone with more experience knows off the top of their head what the problem is.

coderaaron:
After a little more digging, it looks like this is caused by recurring events that have more than one `exdate` (exception date)...I started debugging the ical.js and node-ical.js, but need to do some work so I can't go any further today

DutchDynamics:
[...]

PR #875, opened 2017-04-25
Latest version of ical.js and fix for EXDATE formats

DutchDynamics:
The format of EXDATE is different between google and exchange. This is causing the exclusion on recurring appointments not to work correctly and in some cases the mirror is unable to load the calendar

### DIFF
--- a/modules/default/calendar/vendor/ical.js/ical.js
+++ b/modules/default/calendar/vendor/ical.js/ical.js
@@ -1,319 +1,423 @@
-(function(name, definition) {
+(function (name, definition) {
 
-/****************
- *  A tolerant, minimal icalendar parser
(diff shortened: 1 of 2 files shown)


MichMich:
@roramirez It this ok with the work you're currently doing?

roramirez:
Well ... I think we need approve  only if include some tests. In the past we change something and affect the calendars, maybe was rrule or other related. 

I tested the change proposed (https://github.com/MichMich/MagicMirror/issues/839#issuecomment-297214548)  for set the ORDINAL_BASE lower date and the tests of this library they broke. I think not this way to fix this.

MichMich:
@DutchDynamics Are you able to add tests?

MichMich:
Closed due to inactivity. Feel free to reopen with tests.

Issue #876, opened 2017-04-25
Improve translation flexibility


Arkkimaagi:
String joining can get us only so far regarding translations. In some cases it would be easier to reorder words with the translation strings.

I'd suggest that we'd add support for something like string literals that are coming in future version of js.

For example calendar could support Finnish way of saying the `RUNNING` a bit better
en.js:
```js
	"RUNNING": "Ends in {timeUntilEnd}",
```
fi.js:
```js
	"RUNNING": "Päättyy {timeUntilEnd} päästä",
```
Notice the `{timeUntilEnd}` variable there? That will be replaced with the variable during translation.

This would be rendered in Finnish as:
```
Päättyy viiden minuutin päästä
```
Where moment.js returns `viiden minuutin` as we have five minutes left.

Here is a screenshot from my mirror (where I have dropped the capitalization with css):
![image](https://cloud.githubusercontent.com/assets/1191667/25406371/1ba1a2d8-2a0f-11e7-8397-6fac496515aa.png)

Of course we'd want to keep the translator backwards compatible, we'd want also this to work just as it does now so that all the current implementations do not break:
```js
	"RUNNING": "Ends in",
```
It should fallback to current approach resulting 
```
Ends in five minutes
```

I have a working sample of this thing with an example for this calendar thing only, but support could be built for all translations in future to combat rigid string joining problems.

Arkkimaagi:
Seems that this has been merged into develop, maybe this issue can be closed? Or do you close these only when merged into master?

roramirez:
Hi @Arkkimaagi you should be close this issue.

Arkkimaagi:
Closing the issue as it was merged.

PR #877, opened 2017-04-25
Improve translation flexibility

Arkkimaagi:
As I have described in https://github.com/MichMich/MagicMirror/issues/876 the translation system could be improved with better flexibility in word reordering.

With this code, support for flexible translations can be added with a case by case approach. This change is backwards compatible and should not break existing translations. (As far as I know)

I have provided a sample translation within the default calendar module that solves Finnish language problem with the word order.

If this merge request is accepted, moving from code like this:
```js
this.translate("RUNNING") + " " + moment(event.endDate, "x").fromNow(true)
```
to code like this
```js
this.translate("RUNNING", {
  fallback: this.translate("RUNNING") + " {timeUntilEnd}",
  timeUntilEnd: moment(event.endDate, "x").fromNow(true)
})
```
Would allow translation like this:
```js
"RUNNING": "Päättyy {timeUntilEnd} päästä",
```
to result:
```
Päättyy viiden minuutin päästä
```
while falling back to old behaviour with the `fallback` option given in the variables.

The new translator function checks for the variables object and if it exists, it tries to replace all the variables within the given translation template with ones found in the variables. If the translation template does not contain any variables in curly braces, it checks if there's an fallback option in the variables and uses that to generate the old style translation.

This also fixes the Finnish translation for `RUNNING` events. Instead of saying something like "Ongoing in 5 minutes" like the current translation says, it is fixed to "Ends in 5 minutes" in Finnish.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add test of match current week number on clock module with showWeek configuration.
 - Add test default modules present modules/default/defaultmodules.js.
 - Add unit test calendar_modules function capFirst.
+- Add support for writing translation fucntions to support flexible word order
 
 ### Updated
 - Added missing keys to Polish translation.
 - Added missing key to German translation.
+- Added better translation with flexible word order to Finnish translation
 
 ### Fixed
 - Fix instruction in README for using automatically installer script.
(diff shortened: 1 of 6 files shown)


MichMich:
Awesome work! It does need some documentation though. Could you update the Readme and Changelog? Thanks!

Arkkimaagi:
I will make my suggestion for updating these files, but it will take a couple of days to do as I'm quite busy atm.

Arkkimaagi:
Anything else missing? Please feel free to edit the documentation to your liking, I tried. ;)

PR #883, opened 2017-04-30
Update currentweather to support indoor temperature

jasonyork:
This is an enhancement to allow the `currentweather` module to show indoor and outdoor temperature side-by-side if you have another module that can supply the indoor temperature value.

I'm currently working on a new module to interface with the Vera MiOS home automation system (https://github.com/jasonyork/MMM-mios) with the idea that it could emit the indoor temperature from the thermostat (or other temperature sensing device)

Below is a screenshot from my working implementation.

image

Looking forward to your feedback and ideas.  Thanks.

### DIFF
--- a/modules/default/currentweather/currentweather.js
+++ b/modules/default/currentweather/currentweather.js
@@ -25,6 +25,7 @@ Module.register("currentweather",{
 		lang: config.language,
 		showHumidity: false,
 		degreeLabel: false,
+		showIndoorTemperature: false,
 
 		initialLoadDelay: 0, // 0 seconds delay
 		retryDelay: 2500,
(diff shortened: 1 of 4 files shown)


modules/default/currentweather/currentweather.js
```diff
@@ -203,6 +204,17 @@ Module.register("currentweather",{
 		temperature.innerHTML = " " + this.temperature + "&deg;" + degreeLabel;
 		large.appendChild(temperature);
 
+		if (this.config.showIndoorTemperature && this.indoorTemperature) {
```
roramirez: You should initialized `this.indoorTemperature` variable

jasonyork: Good catch.  Updated.

Issue #891, opened 2017-05-17
[Suggestion] Move custom.css to /config/


JanLoebel:
Hi,

what do you think about moving the "_/css/custom.css_" file to "_/config/custom.css_" and only load it if it exists? This would be great to have the possibility to use Docker and only specify the local config- and module-folder.

Any suggestions?

Best regards,
Jan

roramirez:
There not possible to mount css directory into volume of docker?

JanLoebel:
Not really, because in the css directory there is also the _main.css_ which you would have to copy in the mounted local directory.

MichMich:
What could be an option, is to make the main.css path configurable. With the default set to the current location. This way, you can modify it in your config, and it won't be a breaking change.

Feel free to send a PR for this. Make sure to update the CHANGELOG & README accordingly.

PR #895, opened 2017-05-19
#891 Added ability to change the custom.css path.

JanLoebel:
#891 Added ability to change the custom.css path.

As discussed I've done the simple step to customize the path to the custom.css stylesheet.

### DIFF
--- a/js/loader.js
+++ b/js/loader.js
@@ -35,7 +35,7 @@ var Loader = (function() {
 				// This is done after all the moduels so we can
 				// overwrite all the defined styls.
 
-				loadFile("css/custom.css", function() {
+				loadFile(config.customCss, function() {
 					// custom.css loaded. Start all modules.
 					startModules();
 				});
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add test for check if exits the directories present in defaults modules.
 - Add calendar option to set a separate date format for full day events.
 - Add ability for `currentweather` module to display indoor temperature via INDOOR_TEMPERATURE notification
+- Add ability to change the path of the `custom.css`.
 
 ### Updated
 - Added missing keys to Polish translation.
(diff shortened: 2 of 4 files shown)


js/loader.js
```diff
@@ -35,7 +35,7 @@ var Loader = (function() {
 				// This is done after all the moduels so we can
 				// overwrite all the defined styls.
 
-				loadFile("css/custom.css", function() {
+				loadFile(config.customCss, function() {
```
roramirez: If set a wrong path for this configuration variable all system is crashed. It's to fragile ...I thing can be a handler in case of error. 

So, this can be refactored to add a test case

JanLoebel: ? "all system is crashed"? I've tested it and it works beside the notification in the console like any other resource which could not be loaded, e.g. from a module. 

Tested with:
```
config : {
  customCss: "/iDoNotExists.css"
}
```
Can you please provide an example? If we want to check the path for existing we should add that to the "loadFile" method itself. But I don't see any advantages to the current approach to log the error to the console.

nhubbard: @roramirez @JanLoebel The mirror will most likely freeze up if it attempts to load a non-existent file from C++ (the underlying language in Node). We should have a try-catch case to figure out the status or use `fs.exists()`.

MichMich: I agree. It should give some feedback that it can't find the file on the provided path.

Other than that: Awesome contribution!

JanLoebel: Hi there,

sorry for the late response. I had a look again at your suggestions and I'm still not sure if I get it right.

- fs.exists() can not be used at that point in the loader.js because it can not require NodeJs modules, it's javascript only. So the result would be a bigger refactoring of the loader or the place the custom.css is loaded.

- The internal "loadFile" method in the loader.js is already appling some sort of error handling:
```
stylesheet.onerror = function() {
	console.error("Error on loading stylesheet:", fileName);
	if (typeof callback === "function") {callback();}
};
```
So there will already be logged in the console that the file could not be found.

What would your solution look like?

roramirez: The style of catch error on stylesheet in loadFile looks good.

PR #913, opened 2017-06-07
Added more documentation to weather modules

alberttwong:
Added more documentation (specifically examples for common changes) to change weather a different unit like Fahrenheit.

### DIFF
--- a/modules/default/currentweather/README.md
+++ b/modules/default/currentweather/README.md
@@ -15,6 +15,7 @@ modules: [
 			// See 'Configuration options' for more information.
 			location: "Amsterdam,Netherlands",
 			locationID: "", //Location ID from http://openweathermap.org/help/city_list.txt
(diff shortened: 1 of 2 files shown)


rejas:
Aren't those units taken from the general section of the config file? So that the docs should be clear to change them there and not in the config of the module?

MichMich:
What @rejas says is correct. By default, the general config will be used. You can overwrite this if you want. It might be good to explain that in this readme. You current change gives the idea that this is the only way to set the unit type.

alberttwong:
I would vote for more explanation in the readme because it's confusing and most variables are localized to the module.

MichMich:
Is this something you will change? Otherwise I'll close this PR for now. Cheers!

roramirez:
Should be add more explain into documentation sections not into the sample configuration.

MichMich:
I close this PR due to inactivity. Please feel free to reopen.

Issue #922, opened 2017-06-18
Feature: Allow for server to be run remotely instead of on localhost


QNimbus:
As the title says. I would like to make some changes to allow for the node server to be run on a different server/network location than the electron client. Specifying the server ip/hostname would be a configurable option with a default of 'localhost'.

The server would be started by running `npm serveronly` on the remote location. The app (client) will decide if it needs to start the server (e.g. config file specifies `localhost` as the server location) or if it will direct the electron client to connect to the remote ip/hostname.

I would be more than happy to create a PR for this feature if the community supports this feature

ezar:
+1

rejas:
+1

MichMich:
It might be a good idea to investigate the amount of work that needs to be done before you start this overhaul. If this is doable with a few minor changes, feel free to send a PR.

roramirez:
Sound good for me +1. Today you can set the address where the server can start. Only I see the part is missing is say to client app (electron) address where to get the data, maybe ws could be hard part of this work.

QNimbus:
It will be more work than I had anticipated because the client and server are more intertwined than I had originally thought. However it is possible and I would still like to have a shot at doing this if everyone is on board with this idea/feature.

To get the config from the server to the client (and to prevent having two separate config files) I found a really neat way of using a cookie (that gets removed immediately) to get the config from the server node application to the electron client. Inspired by the [NPM clientconfig package](https://github.com/henrikjoreteg/clientconfig).

I will try to submit a PR as soon as possible but I have quite a few other responsibilities that keep eating away at the time I can invest in this.

ezar:
+1

QNimbus:
I will try to write  some automated tests for this implementation - the feature itself still needs to be tested by users/developers since it's now in the `develop` branch.

Can I close this issue or do we still need this open since it hasn't been properly tested yet?

roramirez:
The ideal to me is close this after wrote the tests. Because a new feature with automated tests we be sure will work like as we expected.

MichMich:
Closed due inactivity (and since it probably already works on the dev branch :))

PR #924, opened 2017-06-25
Added clientonly script

QNimbus:
See issue [Feature: Allow for server to be run remotely instead of on localhost #922](https://github.com/MichMich/MagicMirror/issues/922)

This PR introduces a `clientonly/index.js` script similar to the `serveronly/index.js` script which allows a user to run only the client part (electron) of the MagicMirror application.

The client is started in a similar fashion as the server. By running `node clientonly --address [address-of-server] --port [port-of-server]` you can connect to an already running server instance. The config of the server is passed via a HTTP cookie to the clientonly script which in turn executes the electron client as a child process and passes the config via an environment variable. This ensures that client always uses the same config as the server. (So editing the config.js file client side will not affect the config of the server or client, it is effectively ignored. Otherwise you would end up with multiple config definitions that need to be maintained.

I've tried to keep the changes to the existing source as minimal as possible - this is just a first attempt at implementing this feature. If the community has suggestions that improve or otherwise help implementing this feature, I would like to hear it!

### DIFF
--- a/clientonly/index.js
+++ b/clientonly/index.js
@@ -6,7 +6,7 @@
 (function () {
 	const cookie = require("cookie");
 
-	var config = { };
+	var config = {};
 
 	// Parse command line arguments, if any
 	var addressIndex = process.argv.indexOf("--address");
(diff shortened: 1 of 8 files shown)


MichMich:
Thanks! Can you explain why it uses a cookie?

QNimbus:
@MichMich I chose a cookie to get the server side config to the client because I think it's the leanest way of getting the config over to the client in the same transaction (http.get) instead of having to define a new route (`/config`) on the Express server and sending over the config as a JSON object.

I have to admit it's less verbose than using a separate route, so if that's an issue I could rewrite the PR to use a separate route (RESTfull) and get the config over in that fashion.

Is there another way you think this could be handled better?

QNimbus:
As an unrelated note - I wanted to change the `clientside/index.js` file to also allow for environment variables to be used to get the server ip/hostname and port instead of just using the command line parameters.

Do you think this is something that would improve this feature? If so, I will push the necessary changes to the `client-only-pr` branch.

MichMich:
The cookie solution feels a bit like a hack. I think a separate route is cleaner. It would be great if you could update this PR. (including the mentioned env variables changes).

FYI: I'll release a new master branch July 1st. Since your PR is a high impact change, I'd like it to be tested for a while before I merge it into master. So I'll merge your PR after I release the new master. Is that ok with you?

QNimbus:
@MichMich Okay, I will implement a separate route to get the config from the server. (Along with the env variables)

Also agree on the fact that this PR constitutes a major change and should not be in the next `master` release. I can also try to make some unit tests to verify correct functioning of the clientside-only code.

QNimbus:
Changes have been implemented as requested; I still need to look at making some test cases to properly unit test the clientonly functionality.

I do agree that it looks a lot less 'hacky' by using the Express route instead of the cookie. Thanks for the advice!

clientonly/index.js
```diff
@@ -0,0 +1,98 @@
+/* jshint esversion: 6 */
+
+"use strict";
+
+// Use seperate scope to prevent global scope pollution
+(function () {
+	const cookie = require("cookie");
```
rejas: can be removed i guess?

package.json
```diff
@@ -52,6 +52,7 @@
   "dependencies": {
     "body-parser": "^1.17.2",
     "colors": "^1.1.2",
+    "cookie": "^0.3.1",
```
rejas: Not necessary anymore right?

js/electron.js
```diff
@@ -47,7 +47,7 @@ function createWindow() {
 
 	// and load the index.html of the app.
 	//mainWindow.loadURL('file://' + __dirname + '../../index.html');
-	mainWindow.loadURL("http://localhost:" + config.port);
+	mainWindow.loadURL(`http://${config.address}:${config.port}`);
```
rejas: Instead of ` can " be used like everywhere else?

rejas:
Nice idea and work. Now one can make a MagicMirrorAsAService site :-)

package.json
QNimbus: Correct, I will fix this in the morning! Thanks for catching that!

clientonly/index.js
QNimbus: Correct, I will fix this in the morning! Thanks for catching that!

js/electron.js
QNimbus: Back ticks are used because this is a templated string; in other words, you have to use those back ticks otherwise you get a syntax error

QNimbus: Or did you mean to say to not use templated strings at all? I feel it improves the readability of the code, but if it's against the coding guidelines for this project, I have no problem changing it.

rejas: No, you're right, templating is the right thing to do, didnt notice that (but I am of course not the maintainer to inforce any other rule)

QNimbus:
I rebased the client-only-pr onto the new 'develop' branch

MichMich:
Merged into develop ... let's start the testing! :)

Issue #776, opened 2017-03-14
Calendar 24 hours


olikami:
Infos

**Platform**: Raspberry Pi 3

**Node Version**: v6.10.0

**MagicMirror Version**: v2

**Description**: The calendar shows times as am/pm instead even if timeformat is set to 24.

**Steps to Reproduce**: Display any calendar with the calendar module.

roramirez:
I can see, this feature is not include into calendar module. I mean, the variable set `timeFormat` in global configuration variable isn't have effect the format how the  times are showing.

The config for `timeFormat` of **calendar module** have totally different behavior than global variable.

It's not difficult implement the behavior expected. I think this issue can be open until a contributor (or new) can add this.  

Add label as "first contribution" or "easy commit" can be an idea.

Richard238:
Same issue here. Added event to Google calendar for 26/03/2017 at 22:35
MM calendar shows it, incorrectly, as 10:35PM
MM `timeFormat` is set to 24
MM clock shows 24hr format as expected

roramirez:
HI @Richard238 
This is something we knowed. You can read in my previous comment.
The issue is open at someone can resolve it. If you or someone else are interested to fix go ahead with a Pull Request.

amcolash:
There was a previous issue on this #412. It looks like the easiest way to do this is by changing your language configuration to `en-gb`. I am not sure if there is a need to actually build this functionality out.

Just for those interested in git and version control, check out https://git-scm.com/book/en/v1/Getting-Started
A pull request is basically a request to the repository owners to include new features/functionality/fixes that are not currently in the codebase.
https://yangsu.github.io/pull-request-tutorial/

amcolash:
This is a hackish way of working things out. It won't work for languages other than english (not sure if there are a lot of non-english users). Moment has these formats associated with languages, so it might be a bit tricky to have it work internationally.

`calendar.js`

```
// Set locale.
if (config.language === 'en' && config.timeFormat == 24) {
  config.language = 'en-gb';
}
moment.locale(config.language);
```

Richard238:
@amcolash That works perfectly, thank you! :-)

johanhammar:
The problem is that moment.js defines the longDateFormat as "LT : 'h:mm A'" for en (United States, the default locale), which is what is used by the calendar()-function. 

There are different approaches on how to fix the problem:
1. Don't use moment calendar()-function, instead implement the functionality in the Calendar Module.
2. Use something like

[...]

PR #926, opened 2017-06-26
Calender respects timeformat

QNimbus:
In reference to issue #776, made changes to calendar.js to respect timeformat config option if it is used


DanielWeeber:
24h format does not get displayed correctly for me. It says (language:de timeFormat:24) "Morgen um 04:00" (Tomorrow at 04:00) whilist the appointment is at 4pm (16:00).

Fixed by setting case 24 to "LT: "hh:mm" line 84 commit 3404ebb

Fabian1976:
Same here. It doesn't work for me either (language:nl timeFormat: 24. You would think at least dutch would be tested).

If I change the timeFormat in calander.js to capital HH (like ytmytm mentioed) it does work!

PR #938, opened 2017-07-07
Cleanups

rejas:
Electron with its browser engine doesnt need those prefixes


rejas:
I see. Then it would be helpful to document the supported browsers somewhere and add more prefixes when needed. Is there a consensus on what browsers and version should be supported?

MichMich:
As @roramirez said, these prefixes are needed for other browsers. We officially only support Electron, but stuff like this makes it possible for others to use it as well. I don't think we should add more official supported browsers.

Issue #940, opened 2017-07-08
Calendar times incorrect since update


mc-landin:
Since the update to the latest PI version my calendar times are incorrect. The timezone is set correctly and set to 24H.

Yet, all times after noon are set 12 hours too early.

Tuxdiver:
Problem is in calendar.js:
```
diff --git a/modules/default/calendar/calendar.js b/modules/default/calendar/calendar.js
index 9ab864b9..672a47b 100644
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -81,7 +81,7 @@ Module.register("calendar", {
                case 24: {
                        moment.updateLocale(config.language, {
                                longDateFormat: {
-                                       LT: "hh:mm"
+                                       LT: "HH:mm"
                                }
                        });
                        break;
```

should fix the problem

QNimbus:
You are correct - was sloppy testing on my behalf. The PR should fix this problem. Thanks for reporting!

roramirez:
@MichMich Please, close this issue :). Thanks @QNimbus

PR #943, opened 2017-07-09
Fix issue #940

QNimbus:
Fix for issue 940 - time was incorrectly displayed in a 12-hour fashion regardless of the 24 hour clock preference in config.js


roramirez:
Hi @QNimbus, thanks for you contribution.

Seams good for me but ... Can you provider it a test for this case?

Issue #961, opened 2017-07-28
Issue: When using 'excludedEvents' these count towards 'maximumEntries'


QNimbus:
**Platform**: Raspberry Pi 3

**Node Version**: node v8.2.1

**MagicMirror Version**: MagicMirror v2.1.3-dev

**Description**: 

As the title says. When configuring the calendar module where you define certain events to be excluded, these are counted towards the maximum entries to be shown. This means the calendar will show less than `maximumEntries` even though there are more entries in the range of `maximumDays` to display.

So when you have `maximumEntries` at 6 and `maximumNumberOfDays` at 28, and the first 5 events are in the `excludedEvents` array, only 1 event gets shown. It's even worse when all 6 entries are in the `excludedEvents` array. In this case no event will be shown and also no `No upcoming events` message is displayed. (Since the events array is not empty if you disregard the `excludedEvents`)

**Steps to Reproduce**: 

Use a calendar with more than `maximumEntries` events in the future within the `maximumNumberOfDays` window. Specify `maximumEntries` with a small value (e.g. 6), expecting at least 6 events to be shown. Now make sure a couple of these events get excluded by defining the `excludedEvents` array. You will see that there will no be less than 6 events shown.

**Expected Results**:

To show all non-excluded events up to a maximum of `maximumEntries` within the `maximumNumberOfDays` window.

**Actual Results**:

Excluded events count towards the `maximumEntries` total which means if a certain number of events get excluded, the calendar will show less than `maximumEntries` of events even though there are more events to be shown. In the case where no events are shown it doesn't display the `No upcoming events` message since the array which is tested for length === 0 (`if (events.length === 0)`) is not actually empty.

**Configuration**: 

```
		{
			module: "calendar",
			position: "top_left",
			header: "<i class=\"fa fa-plane\" aria-hidden=\"true\"></i> Schedule",
			config: {
				titleReplace: {
					NIZ: "Niet inzetbaar",
					OSV: "Ouderschaps verlof",
					KOV: "Vrij (korte opdracht)",
				},
				excludedEvents: ["Reisverlof", "EN ROUTE", "AANMELDEN", "AFMELDEN"],
				calendars: [
					{
						symbol: "",
						fadePoint: 0.1,
						maximumEntries: 6,
						maximumNumberOfDays: 28,
						url: "https://calendar.google.com/calendar/ical/[censored]/basic.ics"
					}
				]
			}
		},
```

**Additional Notes**: 

[...]

PR #962, opened 2017-07-28
Fix calendar.js bug

QNimbus:
Fix bug where excludedEvents count towards maximumEntries displayed


QNimbus:
In reference to issue #961

PR #972, opened 2017-07-29
Add warning color when are using full ip whitelist

roramirez:


MichMich:
Could you add an updated changelog? Thanks!

PR #973, opened 2017-07-30
Add log when clientonly failed on starting.

roramirez:


MichMich:
Could you add an updated changelog? Thanks!

PR #977, opened 2017-08-01
Use before after vendors

roramirez:
Use the hooks `before` and `after` in the vendor_spec instead of use `beforeAll` and
`afterAll` in the e2e test to prevent failure in CI.


roramirez:
Good @MichMich!, We've need this patch in our codebase.

PR #979, opened 2017-08-03
Fix this.file path

CatoAntonsen:
I made a small fix to make sure `this.file(filename)` don't return double hash'es ( / ). 

An alternative solution would be to skip manually adding the hash, but I'm not 100% sure `this.data.path` allways ends with a hash.


CatoAntonsen:
@MichMich , seems to be a problem with the tests? What do I do now?

roramirez:
Can you add an test into the testsuite to include this change?

nhubbard:
@CatoAntonsen You might want to close this pull request and re-open it. The test failure was not your fault; it was the vendor suite which was recently disabled by @roramirez.

roramirez:
... Or merge your branch with `upstream/develop` and push it again

PR #982, opened 2017-08-05
Fix Documentation in  roundValue function

roramirez:


fewieden:
Did you test this? parseFloat should create a number and toFixed just changes the amount of decimals, but its still a number?! not a string

roramirez:
Yep. I'll include some tests.

roramirez:
@fewieden The tests are ready. 

Did you see this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

fewieden:
Thanks for the link, I really thought it's returning a number.

PR #986, opened 2017-08-08
Change to install Roboto fonts by npm

roramirez:
Remove the Roboto fonts into source code or project. These are install by npm.


QNimbus:
If I do a fresh install of MagicMirror (and `npm install`), the Roboto fonts will not load because the filenames in the `roboto.css` file does not match the actual font file names.

Example from roboto.css : 
`url("node_modules/roboto-fontface/fonts/roboto-condensed/RobotoCondensed-Light.ttf") format("truetype");`

Actual filename is : 
`/node_modules/roboto-fontface/fonts/roboto-condensed/Roboto-Condensed-Light.ttf`

I just wanted to verify that others can reproduce this issue as well before I proposed a fix for this... 

@roramirez Can you confirm that your fonts do not get loaded as well?

roramirez:
@QNimbus You're right! My mistake :(

There a way to make a test for the css?

QNimbus:
@roramirez I've made an attempt at an e2e test case and added it to the [PR #986](https://github.com/MichMich/MagicMirror/pull/998). I think I inadvertently also found the reason why the vendor_spec.js test case isn't working at this moment (hit and miss) and I will try to test my theory; if it works I will submit a PR for that as well...

roramirez:
I stay expectant :O,

Do you mean this?
https://github.com/MichMich/MagicMirror/pull/1000/files#diff-a7d779483335cc417b43ff18051866adR45

PR #991, opened 2017-08-15
Temperature scale show

DanielWeeber:
Show Scale depening on config flag.
default is "false", so not shown - will not change the current behavior expect you set the config flag.

We could also only set true or false and show celsius and/or fahrenheit depending on the country.


nhubbard:
@DanielWeeber Can you fix the linting errors in the weather forecast file?

https://travis-ci.org/MichMich/MagicMirror/jobs/264845495#L1499
https://travis-ci.org/MichMich/MagicMirror/jobs/264845495#L1450

DanielWeeber:
Hopefully they are fixed now. I'm not sure. All of this is new to me ;)

nhubbard:
@DanielWeeber All good now!
@MichMich How does this look to you?

modules/default/weatherforecast/weatherforecast.js
```diff
@@ -141,11 +142,25 @@ Module.register("weatherforecast",{
...
+				if(this.config.scale == "F" || this.config.scale == "f") {
+					maxTempCell.innerHTML += " °F";
+				}
+			}
 			maxTempCell.className = "align-right bright max-temp";
 			row.appendChild(maxTempCell);
 
 			var minTempCell = document.createElement("td");
 			minTempCell.innerHTML = forecast.minTemp;
+			if(this.config.scale == "C" || this.config.scale == "c") {
```
roramirez: I consider more elegant to use a lower or upper function case over two conditioanl for a letter.

DanielWeeber:
I don't know why the Travis CI check failed. Some checks within the Clock module failed, but I did not change anything there. Could anyone help me to pinpoint this issue?

DanielWeeber:
Only removed a comment so that the Travis-CI runs again. Now it passed... strange.

fewieden:
Instead of creating another config option I would use the global unit config option like in currentweather

https://github.com/MichMich/MagicMirror/blob/master/modules/default/currentweather/currentweather.js#L195-L208

DanielWeeber:
I thought of that also, see my first post. However you all like. I would prefer adding a config label (as also in currentweather "degreeLabel") and defaulting it to false so that current configuration does not get changed.

Issue #950, opened 2017-07-17
Issue: Empty ipWhitelist array does not allow any host to access MM


QNimbus:
When you set `ipWhitelist = []` in the config file, it should allow access from everyone to the MM instance according to the documentation. Instead with an empty whitelist (`express-whitelist` module) it doesn't allow any host to access the MM instance.

I think this could be solved by replacing the following line in `server.js` 

`var result = ipfilter(config.ipWhitelist, {mode: "allow", log: false})`

with

`var result = ipfilter(config.ipWhitelist, {mode: config.ipWhitelist.length === 0 ? "deny" : "allow", log: false})` 

This will result in `ipWhitelist = []` in the config file to allow any and all hosts to access the MM instance.

QNimbus:
To prevent some confusion during configuration (users customizing their `config.js` file), I've changed the default `address` which is used to specify addresses/interfaces where the server listens on, to listen on all interfaces by default. 

If a user changes the default ipWhitelist (which by default only allows localhost) he/she won't be hampered by the fact that they might not have the `address` property configured in their local config.

qistoph:
From a security perspective I would opt to prevent users from accidentally opening up access to their mirror. Listening only on, or allowing access from localhost woudl be wise from that perspective.

Take a look at [Mirai](https://en.wikipedia.org/wiki/Mirai_(malware)) for example.

Raspberry did a major change regarding these issues, by [disabling SSH by default](https://www.raspberrypi.org/blog/a-security-update-for-raspbian-pixel/).

Therefore I would suggest to keep the default to limited access.

QNimbus:
I agree with you - I will try to come up with a more elegant solution, and send a PR with reference to this issue.

roramirez:
There a log when is open a full ipWhitelist.

qistoph:
@QNimbus, any updates?

@roramirez, what do you mean?

roramirez:
When you are using a full IpWhitelist, MM alert you with a  warning log entry in console.
https://github.com/MichMich/MagicMirror/blob/develop/js/server.js#L30

qistoph:
My concern is that the default install & run will leave you with an application listening on all network devices on any IP (include IPv6). Given the increase in IoT devices being abused I think this is something to try and prevent as software developers.

So I think the default install & run should at least prevent one of the two; 

- listen only on the localhost interface, or 
- listen only on the localhost IPs 127.0.0.1 and ::1

[...]

PR #1004, opened 2017-08-25
More secure defaults.

qistoph:
Related to #950. More secure suggestion.

Without config, listen only on looback interface. In sample config
listen on any interface, but use an IP whitelist.

### DIFF
--- a/config/config.js.sample
+++ b/config/config.js.sample
@@ -9,10 +9,11 @@
  */
 
 var config = {
-	address: "", // Address to listen on, can be
-	             // "localhost", "127.0.0.1", "::1" to listen on loopback interface
-	             // another specific IPv4/6 to listen on a specific interface
-	             // "", "0.0.0.0", "::" to listen on any interface
+	address: "", // Address to listen on, can be:
+	             // - "localhost", "127.0.0.1", "::1" to listen on loopback interface
+	             // - another specific IPv4/6 to listen on a specific interface
+	             // - "", "0.0.0.0", "::" to listen on any interface
+				 // Default, when address config is left out, is "localhost"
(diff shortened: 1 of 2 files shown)


MichMich:
Maybe the comments should reflect what the default value is when no adds is given?

qistoph:
@MichMich, something like this? dfcbfad9a3c41df49d55f9a1907e5db6bfd82f05

config/config.js.sample
```diff
@@ -9,7 +9,11 @@
  */
 
 var config = {
-	address: "localhost",
+	address: "", // Address to listen on, can be:
```
roramirez: I think we need keep the localhost value for config sample... and only add comments here.

qistoph: Okay, sounds fine to me

qistoph:
I've been working on some test cases for these configuration options, but keep running into issues...

https://github.com/qistoph/MagicMirror/tree/issue950_2

It seems that starting the app from the global namespace of mocha, leaves traces that influence later test cases. So I've tried to run them in a sandbox, but that is starting to be quite some work...

https://github.com/qistoph/MagicMirror/tree/issue950_sandbox

Do you maybe have any suggestions @roramirez ?

config/config.js.sample
MichMich: Will you update this in the PR?

roramirez:
@qistoph why not use as e2e tests?

roramirez:
I pushed the fix
It's need set a valid file for defaults if not app get a exception and the tests will be fail
https://github.com/qistoph/MagicMirror/pull/1

We need now fix the conflict with version tests.

:)

PR #1011, opened 2017-09-01
Fix on 991

qistoph:
Change 5568e0c2adc73aa20bebb5bfc47df65fea273b05 for #991 introduces JavaScript error, because it's calling toUpperCase() on a boolean value (default: false).

This PR adds a check to see if it's a string.


qistoph:
I have added a commit to make the config of weather forecast more in line with current weather. It uses the configured units (metric, imperial, default) to determine the sign.

On a side note, it's "x degree Celcius", "x degree Fahrenheit", but "x Kelvin". So the degree-sign should not be included for Kelvin. I took this into account for the Weather Forecast module, but did not change it in the Current Weather module.

Issue #1017, opened 2017-09-07
Weather forecast module doesn't show up


rhodso:
**Platform**: Raspberry Pi Model 3b
**Node Version**: Latest
**MagicMirror Version**: Javascript

**Description**: 
As discused in: https://forum.magicmirror.builders/topic/4781/problem-with-weather-forecast/16?page=1 , the Weather forecast module doesn't show up. It has never showed up ever since my install 2 days ago

**Steps to Reproduce**: Set up on a Pi using the automatic installer
**Expected Results**: Weather forecast section
**Actual Results**: Weather forecast section not there
**Configuration**: https://pastebin.com/cf3eHaHM
**Additional Notes**: Attempted the 3h fix from Mykle1, still didn't work

sfmailand:
It looks to me like you haven´t put a valid API key into the config, which would cause it to be unable to pull weather (and therefore not be able to display it)

rhodso:
Nope, that's not it.
I've tried it with both keys I own and I generated another key.

Config file: https://pastebin.com/xdAwUHnD
WeatherForecast.js file: https://pastebin.com/H6qkhbVg

I verified that all 3 keys I own work by using the Current Weather Module, which shows up evey time I have a valid key

Negendahl:
Im having this issue too, i located the problem not being with the API, but the /daily part of the URL. 
This is my "original" URL created by weather forecast (dosent work - Invalid API key error)
http://api.openweathermap.org/data/2.5/forecast/daily?id=2613766&units=metric&lang=en&cnt=7&APPID=APIKEY

This URL works (only removed the "/daily" part of the URL):
http://api.openweathermap.org/data/2.5/forecast?id=2613766&units=metric&lang=en&cnt=7&APPID=APIKEY

I cant get to to work though the module though, but the problem is not with the API key.

Hope it helps :) 

/Mick

nhubbard:
@MichMich Could you take a look at this? I'm currently inundated with school, so I can't do it.

MichMich:
I just did a clean install of the MagicMirror master branch and used @rhodso's config file using my own API key. The weather forecast appears as expected.

Just to be sure:

- Which branch are you running?
- Are you running the latest version? (`git pull` to update).
- Did you try removing the node_modules folder and run `npm install` again?
- Do you see any errors in the Electron console?
- Are you sure used the correct API key?

Let me know!

acejacek:
[...]

PR #1024, opened 2017-09-16
Hack fix issue #1017

roramirez:


nhubbard:
@roramirez I wonder what is causing the tests to fail on three out of four platforms... none of them are related to the weather module. Maybe there are errors in the console that we can't see.

roramirez:
Yes, this is something we need to fix. Its random cases

kmcrawford:
I'm not sure this is a fix. That API returns data in 3 hour increments vs a daily.  So you would have to calculate the low and high across all 3 hour increments for a day to see the daily high & low.

nhubbard:
@kmcrawford Send me an email at nhubbard@users.noreply.github.com with your email. I will send you an API key that was grandfathered into daily endpoint access.

MichMich:
What does this fix do? I don't understand it's purpose.

MichMich:
Hopefully I have time tomorrow to start working on a new weather integration to overcome the API issue. Before I do: did anyone already started working on this?
My plan is to make one new module that replaces both the current weather as well as the forecast module. It will have multiple sources (user can choose).

kmcrawford:
I haven't yet started on it.  But your approach would be the same approach I would do.  These apis are pretty simple and straightforward.

nhubbard:
@MichMich Which service would you plan on using? Dark Sky is free unless you make 1,000 calls in a day. I've never used Apixu, but it sounds just as promising. In a perfect world, we would make it pluggable (i.e. bring your own weather provider) but that adds lots of complexity.

MichMich:
@nhubbard. Thats sort of the plan: multiple providers. More can be added in the future. 

If you like to contribute to the discussion, Please join Slack.

kmcrawford:
What is the Slack channel?

I was bored in a meeting to made the changes in a new branch.

https://github.com/kmcrawford/MagicMirror/tree/new-api

MichMich:
https://forum.magicmirror.builders/topic/4759/magicmirror-in-slack-we-ll-see-you-there

roramirez:
@MichMich The purpose if the openweather api return an  401 code when you use a  new account for forecast daily will to  change the other endpoint where is work.

MichMich:
FYI: https://github.com/MichMich/MagicMirror/commit/ef172592b8b1657a92ceea5a56b6b0629abf6046

roramirez:
Look good!. But If we'll use forecast I would maintain the behavior included in this change.

MichMich:
@kmcrawford I just pushed a fix that should solve the three hour increment. Could you give it a try?

PR #1029, opened 2017-09-20
Add APIXU support

kmcrawford:
Since openmaps api doesn't work for forecast for new users, this will allow users to continue or  use open maps and use apex


kmcrawford:
I disagree with some of these linter warnings...

PR #1030, opened 2017-09-21
Support for apixu

kmcrawford:
Since openmaps api doesn't work for forecast for new users, this will allow users to continue or use open maps and use apex.


MichMich:
Thanks for this contribution. What might be a good idea, is to rename `apiXuKey` to `apiKey`. This way, if we want to start using other providers in the future, we won't get a whole bunch of different variables. Is it possible to change this?

@roramirez What do you think of this (temporary) solution?

kmcrawford:
That's cool, it just needs an ENUM so the system knows which api to use, vs trying one api and getting a 403, then trying the other.  

Ex: apiType: "APIXU" //Default OPENWEATHERMAPS

MichMich:
That's exactly what I'm thinking ... I'm planning to make one module for both the forecast as the current wether (configurable to choose what it displays), which has multiple providers you can choose from. I hope to create a new branch with the new module soon ....

kmcrawford:
Well, the code is done.. CI keeps failing on intermittent issues.

MichMich:
FYI: https://github.com/MichMich/MagicMirror/commit/ef172592b8b1657a92ceea5a56b6b0629abf6046

roramirez:
@MichMich I think this is a lot changes for the module, the chance to have a lot issues is bigger too. I cant see it as temporaly solution. I think the temporal soluction is #1024 to before merge with the new module as you are building.

Maybe @kmcrawford after than can provider a module provider for Apixu

kmcrawford:
#1024 is not a working solution.  The api is a 3 hour increment, if you run it you will the days are not aggregated.

MichMich:
Probably not what you hoped after your huge amount of work, but I won't be merging this PR since we are working on a more long term fix. We might be able to reuse some of your work in the new weather module. Closing this PR for now, let me know if you want to discuss.  (Also, feel free to join us on slack to work on the new module.)

PR #1032, opened 2017-09-25
Fixed weatherforecast module, issue #1017

Sp4M:
weatherforecast daily api (openweather) is not free anymore. We can still use 5d/3h api, this patch the module to use the new api. (issue #1017)


MichMich:
Thanks for this contribution. The problem to this is that users who have a valid API key (like myself) cannot use the long term forecast anymore. It should be optional to use the alternative forecast. Additionally, I don't really understand the use of `iconTableOrdered`, could you explain?

MichMich:
Fixed in: https://github.com/MichMich/MagicMirror/pull/1024

Sp4M:
You can use the alternative forecast just like the previous one with this patch.
Basically all the "3h parts" are aggregated for each days. 
The min/max of the day is the global min/max of ths parts. 
This idea is not possible for icons, so iconTableOrdered solve this problem by defining what icon to use after aggregation : it assumes that you want to see the worst case of the day (if it's all sunny with a big thunder at 12h, I'll prefer to see the thunder icon fot this day). With this in the config, everyone can choose his own worst case.

PR #951, opened 2017-07-17
Fix for issue #950

QNimbus:
Changed 'server.js' to allow an empty ipwhitelist to allow any and all hosts instead of none as mentioned in the documentation

Issue #1036, opened 2017-09-29
ipwhitelist not working


t4rif4ri:
**Platform**: RPI 3 with newest MM build, installed and updated yesterday.

**Description**:  Until yesterday my mirror was working with an older build version. The mirror was accessable trough the internet via my configured ddns service. 
After the update im getting the error, that the ip was blocked. 
I tested with ipwhitelist [] but it isnt working. Additionally i entered 0.0.0.0 to adress wo get the ipwhitelist: [] working, but it didnt fix the problem.

How can i make the mirror accessable from every ip adress?

MichMich:
Are you running the develop branch or the master branch?

t4rif4ri:
actual Branch is master

MichMich:
Could you give the develop branch a try?

    git checkout develop
    git pull
    npm install

t4rif4ri:
Ok, this problem is solved, no blocking anymore in shell console, but cant connect anyway. Browser output: (111) Connection refused

MichMich:
So, is it solved? Or isn't it? :D If it isn't solved, what message do you see in your node output when trying to connect?

t4rif4ri:
Output is not showing any blocked ips, but the connection got refused when i open the link in the browser. Ports are open, like when it was working.

roramirez:
There was  fix into the `develop` branch about this issue. Please, check it out
https://github.com/MichMich/MagicMirror/commit/45d597ac49d52e1728b9d40d95b0a7acf2945cbd

So, this issue is really fixed.

As @MichMich said

```
cd MagicMirror
git checkout develop
git pull
rm -frv node_modules
npm install
```

Make sure you can access from the browser and where is running MagicMirror. Are differents IPs?, I mean MM and Browser

t4rif4ri:
mirror is accessable from intern network. Also within the mirror itself. From outside i get the connection refused error shown in my browser. Tried different port settings, but cant connect via dyndns from outside. Other ports like webserver on port 80 is still working and accessable from outside.

roramirez:
... I presumed this is configuration issue or other not related with MagicMirror itself. Did you set [`address` to `'0.0.0.0'`  in your config.js](https://github.com/MichMich/MagicMirror#configuration)?

What show you the command `netstat -putan` ?

nhubbard:
@t4rif4ri We don't (officially) support DynDNS. I would recommend taking your case to the forums.

@MichMich close?

[...]

PR #1042, opened 2017-10-01
Updated readme about default value for `address` configuration

roramirez:
This change was introduce before to make more secure the MagicMirror.

You can find the commit in  https://github.com/MichMich/MagicMirror/commit/2f05228d

And Pull Requests related #950 and #951


roramirez:
Reference issue #1036

PR #1058, opened 2017-10-13
Gracefully Shutdown Module node_helpers on exit/SIGINT (Fixes #1056)

### LINKED ISSUE #1056: No method to gracefully shutdown modules
I have developed a few modules that spawn other processes and I could not find any existing method to gracefully exit when stopping or restarting the MagicMirror.  Because of the way Electron handles SIGINT commands, it is not as simple as adding a `process.on('SIGINT', function...)` to the node_helper.  Unless you are running `server_only`, this never gets called.

I would like to propose: when exiting (e.g. SIGINT, CTRL+C, pm2 stop, or pm2 restart), check each `node_helper` for a `stop` function, and call it if it exists; similar to the `start` function.  This will allow module developers to implement a "stop" function if necessary to clean up their modules.

Working prototype here: https://github.com/shbatm/MagicMirror/commit/9935d2fccb4206c2949b43cc118aad6789303f12 

* Works when using Electron (`app.on("before-quit"`) or server-only (`process.on("SIGINT"`)
* Prevents modules stalling by force-quitting after 3 seconds if the `stop` functions are taking too long.

Please let me know your thoughts or if there is already a better way to do this.  I can start a PR if this is something that would be useful.

shbatm:
Fix for Issue #1056: When exiting MagicMirror server (e.g. SIGINT, CTRL+C, pm2 stop, or pm2 restart), check each node_helper for a `stop` function, and call it if it exists. This will allow module developers to implement a "stop" function if necessary to clean up their modules, close any connections, and stop any child processes.

To implement in a module, add a `stop` function:

**Example:** 
````javascript 
stop: function() { 
  console.log("Shutting down MyModule"); 
  this.connection.close(); 
} 
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Add option to use [Nunjucks](https://mozilla.github.io/nunjucks/) templates in modules. (See `helloworld` module as an example.)
 - Add Bulgarian translations for MagicMirror² and Alert module
+- Add graceful shutdown of modules by calling `stop` function of each `node_helper` on SIGINT before exiting.
 
 ### Updated
 
--- a/js/app.js
+++ b/js/app.js
@@ -236,6 +236,33 @@ var App = function() {
 			});
 		});
 	};
+
+	/* stop()
+	 * This methods stops the core app.
+	 * This calls each node_helper's STOP() function, if it exists.
+	 * Added to fix #1056
+	 */
+	this.stop = function() {
+		for (var h in nodeHelpers) {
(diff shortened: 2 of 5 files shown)


MichMich:
It might be a good idea to add a default implementation in https://github.com/MichMich/MagicMirror/blob/develop/js/module.js

shbatm:
Does there need to be a default in `module.js`? This would only be called in `node_helper.js` and I thought  `module.js` was for the front-end.

shbatm:
Added default implementation into `modules/node_modules/node_helper/index.js`.  This looked like the appropriate place, please advise if it should be somewhere else.

MichMich:
Sorry. You were correct. That's the appropriate file. Thanks for adding it!

PR #1053, opened 2017-10-10
Delete large package-lock.json files

roramirez:

PR #1060, opened 2017-10-14
Revert "Delete large package-lock.json files"

nhubbard:
Reverts MichMich/MagicMirror#1053. There is a clear reason to keep these files. They may be large in line length, but it isn't best practice to remove these files... they will re-appear later sometime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,3 @@ Temporary Items
 *.orig
 *.rej
 *.bak
-
-# all package-lock.json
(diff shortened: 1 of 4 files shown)


MichMich:
According to https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md:

"This file is intended to be committed into source repositories, and serves various purposes ..."

Reverting the removal of the file as @nhubbard suggested.

PR #1080, opened 2017-11-15
some new features (resubmission)

slametps:
* truncated description option and how long (feed description from bloomberg/engadget for instance, is really a long description. So, it need to be truncated)
* per feed reload interval support


MichMich:
Thanks! Merged!

PR #1111, opened 2017-12-30
added option of decimal comma for temp values for default weather modules

TTigges:
This adds the option of using a comma instead of the point for the decimal mark in temperature values for both default weather apps: currentweather and weatherforecast.

I added the line _"decimalComma: false,"_ to _"defaults"_ (after language, as I saw it fit there). Having the default value _"false"_, the changes are not visible until the line _"decimalComma: true,"_ is added in the _config.js_.

The lines that serve the temperature values are surrounded by **if (decimalComma)** waiting for _"true"_ to replace the decimal point with a decimal comma, leaving the initial line if it stays _"false"_:

`if (this.config.decimalComma) {`
     `temperature.innerHTML = " " + this.temperature.replace(".",",") + "&deg;" + degreeLabel;`
`}`
`else {`
_stays as it was before_
`}`

**.replace(".",",")** does the trick.

I think this update is of value to a lot of people, seeing as how more countries/languages use the decimal comma instead of the decimal point, while still leaving the original decimal mark for those that don't notice or care (also looking at the total number of people using the decimal point, which might be higher ...).

Regards,
Torben

For more information about the usage of decimal comma vs. decimal point, see: https://en.wikipedia.org/wiki/Decimal_mark#Countries_using_Arabic_numerals_with_decimal_point


MichMich:
Thanks for your contribution. If you allow me to be a bit picky: it might be a better idea to use a `decimalSymbol` config option with a default of ‘.’ Instead of the `decimalComma` setting. 

This way *any* char can be used as a decimalSymbol. 

Also: don’t forget to add an updated changelog.

TTigges:
Thanks for the feedback. I think that going for the input option might be a bit too complicated for some users, seeing as there are mainly only points and commas and not a lot of people know their typography details. But then again, you’re right, some people might want Arabic or Persian decimal marks.

I consider the true/false option to choose between point and comma to be safer and still better than having just the point. It's about keeping it simpe as something like an input option has the potential for a lot of weird stuff to happen.

Anyway, I changed it back to a one-liner:
`temperature.innerHTML = " " + this.temperature.replace(".", this.config.decimalSymbol) + "&deg;" + degreeLabel;`

But I added at least a fallback if the value is left empty, so there's no chance of 100°C ;-)
`if (this.config.decimalSymbol === "") this.config.decimalSymbol = ".";`

I also added a line to the develop-branch CHANGELOG.md. Let me know what you think.

MichMich:
Works perfect! Thanks for your contribution!

TTigges:
You're welcome. Working on more. ;)

PR #1116, opened 2018-01-01
DOM creation notifications in cases of async template rendering

ConnorChristie:
Problem: The current way the `DOM_OBJECTS_CREATED` notification is sent out is after all doms are processed but not fully rendered in cases of async template doms.

I propose a solution to send a notification when that module's dom has been fully updated. This change does not break anything and only adds a notification to when a module's dom has been fully rendered.

This issue has surfaced for me because I was trying to render a chart on a div rendered by a template and when `DOM_OBJECTS_CREATED` was called, my dom had not actually been loaded yet since it was still being processed by the templating library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.2.3] - 2018-01-03
+
+### Added
+
+- Add system notification `MODULE_DOM_CREATED` for notifying each module when their Dom has been fully loaded.
+
 ## [2.2.2] - 2018-01-02
 
 ### Added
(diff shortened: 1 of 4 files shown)


MichMich:
Thanks! Could you add an updated `CHANGELOG.md`?

ConnorChristie:
I updated the changelog but the build seems to have timed out on one of the e2e tests...

MichMich:
Merged! Thanks! :)

shbatm:
Just pulled an update to the develop branch and found that this commit had broken my version of the MMM-Carousel module.  The module responds to a "DOM_OBJECTS_CREATED" message to start the module transitions; however, after this update, "DOM_OBJECTS_CREATED" is sent too early; before the DOM is actually populated with the result of the `module.getDom()` call and every function which depends on this modules DOM being in place fails.

Changing my module to respond to the new "MODULE_DOM_CREATED" notification fixes the problem, but it this could have a large impact on any existing modules out there which don't use templating and rely on the DOM_OBJECTS_CREATED notification to start doing whatever they're going to do.

ConnorChristie:
@shbatm I have opened an issue to track this bug that has been introduced. I will look into modifying the updated flow to call the original `DOM_OBJECTS_CREATED` synchronously after all non-templated DOMs are rendered

PR #1127, opened 2018-01-07
Set only 7 version of node to run tests in Travis CI

roramirez:


E3V3A:
Why?

MichMich:
@E3V3A Unfortunately a lot of Travis test fail due an timeout. Because we are currently running tests on 4 versions, almost all Travis test fail. By limiting the amount of versions we test, we hope to limit the number of failed travis tests. Of course this is a bad workaround. But for now it a welcome change. If you want to contribute to QA, make sure to join our slack channel (where we discussed this option).

PR #1128, opened 2018-01-07
PowerSaving without a PIR Sensor

RpDp-git:
I have written a piece of code in python that puts the display to sleep if a device(preferably phone or smart watch) is not connected to the network. See :  https://github.com/RpDp-git/PiDisplaySleep


roramirez:
I can't understand the change here. Are you looking add to the module list?

Please add the module in Wiki page:

https://github.com/MichMich/MagicMirror/wiki#3rd-party-modules

RpDp-git:
Sorry. I think I posted in the wrong place.

Issue #1140, opened 2018-01-15
DOM_OBJECTS_CREATED being fired off too early


ConnorChristie:
Due to https://github.com/MichMich/MagicMirror/pull/1116 the original `DOM_OBJECTS_CREATED` event is being sent to early causing some existing modules not be able to access DOM objects.

**Affected Modules**:
- MMM-Carousel

Root cause analysis:
> Just pulled an update to the develop branch and found that this commit had broken my version of the MMM-Carousel module. The module responds to a "DOM_OBJECTS_CREATED" message to start the module transitions; however, after this update, "DOM_OBJECTS_CREATED" is sent too early; before the DOM is actually populated with the result of the module.getDom() call and every function which depends on this modules DOM being in place fails.
> 
> Changing my module to respond to the new "MODULE_DOM_CREATED" notification fixes the problem, but this could have a large impact on any existing modules out there which don't use templating and rely on the DOM_OBJECTS_CREATED notification to start doing whatever they're going to do with the assumption their DOM objects are in place.

shbatm:
Thank you for looking into this.

MMM-ip appears to be affected as well.

shbatm:
Not sure if this should be a separate issue or not, but it's related to the same commit that originated this bug:
* `this.sendNotification` now fails when called from another module without a `sendTo` parameter:

```js
main.js:88 Uncaught TypeError: Cannot read property 'notificationReceived' of undefined
    at sendNotification (main.js:88)
    at Object.sendNotification (main.js:488)
    at Class.sendNotification (module.js:367)
    at Class.broadcastEvents (calendar.js:529)
    at Class.socketNotificationReceived (calendar.js:106)
    at module.js:245
    at r.<anonymous> (socketclient.js:32)
    at r.emit (index.js:133)
    at r.onevent (socket.js:270)
    at r.MMSocket.self.socket.onevent (socketclient.js:26)
```

MichMich:
I merged a pull request yesterday. Did this solve this issue?

ConnorChristie:
If it was the PR by me then it wouldn’t fix this second issue, that seems to be a different problem.

@shbatm Would you be able to look into why it’s throwing that exception with the sendTo added?

shbatm:
It doesn't appear to be the `sendTo` parameter causing the issue after further investigation. Something with the change to `for (var module of modules) {` from 
```js
for (var m in modules) {
          var module = modules[m]; 
```
[...]

PR #1143, opened 2018-01-17
Fix for #1140 - sendNotification module errors after #1116

shbatm:
Revert change to `js/main.js`'s `sendNotification` function back to original method.

```js
for (var module of modules) {
```
This ES6 method does not behave well with the `modules` array when `config.js` contains disabled modules. It results in an empty `module` variable for every disabled module, which causes an error when the function tries to send a notification to an undefined module.

Reverting back to the original:
```js
for (var m in modules) {
    var module = modules[m];
```


ConnorChristie:
Looks good, thanks for opening this PR!

PR #1146, opened 2018-01-20
Put article iframe in front of modules

cederstrom:
Before this change the article was brought up in its ifram in fullscreen and you could still see the other modules in front of it


MichMich:
Thanks! :)

thobach:
@cederstrom Nice contribution! I now made it scroll as well - do you want to test it? https://github.com/MichMich/MagicMirror/pull/1161

PR #1150, opened 2018-01-23
add event location if it exists

bonaime:
- add event location if it exists
- event location could have they own css but don't know how to do it...
#346 

https://forum.magicmirror.builders/topic/382/calendar-location


MichMich:
Thanks, I like the idea, but I don't think it should be on by default. Additionally the implementation could be a bit cleaner by using classes indeed. Check out the forum find out how this is done.

Please refer to http://forum.magicmirror.builders for support and troubleshooting.

bonaime:
I will see if I could improve #346

PR #1176, opened 2018-02-11
Forecast - Max Days

relm923:
Forecast module:

When `config.forecastEndpoint = "forecast"` `config.maxNumberOfDays` is not properly respected. Ex: `config.maxNumberOfDays = 3` 5 days are rendered because the API maxes out at 39 3 hour entries.

This change does not cap the API call and instead checks the max while parsing the response.


MichMich:
Just to be sure. Did you check this with both a high setting, als well as a low setting? Because I think the &cnt property on the api does make a difference. It's not that they always give back same number of entries, right?

relm923:
The `/forecast` API defaults and maxes out at `cnt=40` so capping during parsing has the same effect. I can't confirm the `/forecast/daily` behaves similarly since I don't have access to it (new openweather account) and their docs are not clear.

MichMich:
I think it's important to test it with one of the older account. I'll check on slack if someone is willing to test this (I'm a little limited on time ...)

relm923:
I'm happy to work with someone on Slack too but it appears the invite limit was reached and I can't sign up

MichMich:
Hm, thats a good point. Maybe try the forum to check with someone?

E3V3A:
@relm923  Any updates? 

@MichMich 
Since it's dev branch perhaps it could be merged anyway, and people would complain in case something broke. Not ideal, but since nothing happened here since 1+ month... Also having it in the dev branch, allows easier testing.

MichMich:
@E3V3A Done. Could you do some testing? Since I'll be releasing 2.3.0 in a few days ...

PR #1190, opened 2018-03-01
made the module "this" instances into a table

E3V3A:


MichMich:
Might be a good idea to do the same for this.data.

PR #1193, opened 2018-03-03
Added Feels Like and Windspeed in KMPH for currentweather module

iampranavsethi:
Added two new configuration options for currentweather module. 

`showFeelsLike`: Shows how it actually feels like (wind chill or heat index) 
`useKMPHwind`: adds an option to see wind speed in Kmph instead of just m/s or Beaufort


MichMich:
Please send your PR's to the develop branch and don't forget to add an updated Changelog and modified documentation.

PR #1201, opened 2018-03-11
 Fixed to work in Midori 0.4.3

parnic:
Was getting an error trying to run this on Midori:
http://localhost:8080/modules/default/compliments//compliments.js @53: SyntaxError: Unexpected token '>'
so I replaced it with the old-style syntax. I use Midori instead of Chromium for performance reasons on my Pi Zero running Raspbian Jessie.

### DIFF
--- a/modules/default/compliments/compliments.js
+++ b/modules/default/compliments/compliments.js
@@ -55,7 +55,7 @@ Module.register("compliments", {
 
 		var self = this;
 		if (this.config.remoteFile != null) {
-			this.complimentFile((response) => {
+			this.complimentFile(function(response) {
 				this.config.compliments = JSON.parse(response);
 				self.updateDom();
 			});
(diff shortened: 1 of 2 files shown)


modules/default/compliments/compliments.js
```diff
@@ -55,7 +55,7 @@ Module.register("compliments", {
 
 		var self = this;
 		if (this.config.remoteFile != null) {
-			this.complimentFile((response) => {
+			this.complimentFile(function(response) {
 				this.config.compliments = JSON.parse(response);
```
fewieden: Just to let you know, this is not a correct port to es5. In order to work the same as the current version, you also have to replace this with self in this line, otherwise you cannot use the external file

parnic: Thanks! I closed the pull request when the build failed and I realized it probably wasn't working properly, I just wasn't using a compliment file so it didn't show up locally.

PR #1202, opened 2018-03-12
Currentweather module updates

iampranavsethi:
v2.4.0
- Added two new configuration options for currentweather module.

`showFeelsLike`: Shows how it actually feels like (wind chill or heat index)
`useKMPHwind`: adds an option to see wind speed in Kmph instead of just m/s or Beaufort

Updated `CHANGELOG.md` and `README.md` for currentweather module.


E3V3A:
Nice!

PR #1209, opened 2018-03-14
update node stable to 9.x

### LINKED ISSUE #1208: Increase the stable node version in installation script
In the `dev` branch the current instalaltion script is using the `6.x` as the [stable Debian node](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions) branch. This is very out dated and should be changed to `9.x`. 

[Here](https://github.com/MichMich/MagicMirror/blob/develop/installers/raspberry.sh#L85) is the line. 
```bash
	NODE_STABLE_BRANCH="6.x"
```

That is:
```bash
curl -sL https://deb.nodesource.com/setup_9.x | sudo -E bash -
```

E3V3A:
update node stable to 9.x in installation script. 
This would fix #1208.


fewieden:
I suggest to set it to a LTS version, in that case 8.x

PR #1248, opened 2018-04-02
null check for notification removal

### LINKED ISSUE #1240: Uncaught TypeError: Cannot read property dismiss of null (alert.js)
I keep getting random events where the error is:  

**`Uncaught TypeError: Cannot read property 'dismiss' of null`**

alert_bug

The error is coming from the Alert default module on Line 113, in [alert.js](https://github.com/MichMich/MagicMirror/blob/develop/modules/default/alert/alert.js#L113):

```js
	hide_alert: function(sender) {
		//Dismiss alert and remove from this.alerts
		this.alerts[sender.name].dismiss();  // <<========= LINE  113 !
		this.alerts[sender.name] = null;
		//Remove overlay
		var overlay = document.getElementById("overlay");
		overlay.parentNode.removeChild(overlay);
},
```

@MichMich  
Does this have to be `null` or can it also be set to `""` or something else?
Any ideas how to fix or work around this?

I'm surprised nobody else has reported this issue, so I want to eliminate that it is a problem with my module.

E3V3A:
Make sure there is something to remove, before we attempt to remove the notifications. 
- This fixes #1240

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Fixed issue where wind chill could not be displayed in Fahrenheit. [#1247](https://github.com/MichMich/MagicMirror/issues/1247)
+- Fix issues where a module crashes when it tries to dismiss a non exsisting alert. [#1240](https://github.com/MichMich/MagicMirror/issues/1240)
 
 ### Updated
 - Updated italian translation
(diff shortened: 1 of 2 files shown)


MichMich:
Could you be so kind to add a Changelog? 😘

paviro:
Faster than your bot! 😅
I probably should have added this to begin with. Guess I thought one wouldn’t try to close a notification without having opened one. Would definitely add a check nowadays anyway 😄

E3V3A:
I was doing this from web interface and it seem that the user (me!) didn't know how to properly add the commit and change the CHANGELOG. Also the CHANGELOG seem more comprehensive for each release, so please edit it yourself, this time. Sorry for the inconvenience.

PR #1257, opened 2018-04-08
Fixed to work in Midori browser

parnic:
The Midori browser doesn't support new JS features. Midori is the only browser that I've found that's performant enough to be run on a Raspberry Pi Zero. MagicMirror v2.2.2 worked fine with Midori, but changes added in 2.3.0 broke it.

Compliments was broken in v2.2.2 as well, but I never made a pull request for my local changes at the time.


fewieden:
I'm not a fan of putting old syntax in the master branch. Spreading `var self = this` everywhere etc.

What I did for one of my modules, I created a specific `midori-supported` branch, which basically does what you did. I would prefer to do the same for MagicMirror

parnic:
As a point of reference, MM2.2 worked fine. MM2.3 just displays a black screen. So it's not like there's "tons" of old syntax or `self = this` everywhere. It's just a question of whether the project wants to be something that just works everywhere or would prefer to maintain multiple branches just for syntactical concerns. I can stop submitting pull requests and just do the fixup locally if the project wants to drop support for the Pi zero.

fewieden:
We had a discussion here on github once if we support midori or not, as it is not officially supported, at the end it is in the hands of michael. My goal is not to stop you from contributing to this project

MichMich:
Feel free to send a PR. I won’t work on midori myself, but it’s ok to merge. Please keep the PR’s compact (no major overhaul) and don’t forget the changelog.

PR #1275, opened 2018-04-27
Added option to calendar module that colors only the symbol

john3300:
This adds a new config option, 'coloredSymbolOnly', to the calendar module that makes only the symbol be colored, and not the whole line.  I found having the whole line be colored was too much when everything else on the screen was white.  So coloring just the symbol still gives me the visual indication that I need, but with a smaller impact on the whole.

Before and after screenshots are attached.
before
after

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -31,6 +31,7 @@ Module.register("calendar", {
 		hidePrivate: false,
 		hideOngoing: false,
 		colored: false,
+		coloredSymbolOnly: false,
 		calendars: [
 			{
 				symbol: "calendar",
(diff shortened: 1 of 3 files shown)


modules/default/calendar/calendar.js
```diff
@@ -135,14 +136,19 @@ Module.register("calendar", {
...
+			if (this.config.colored && !this.config.coloredSymbolOnly) {
 				eventWrapper.style.cssText = "color:" + this.colorForUrl(event.url);
 			}
 
 			eventWrapper.className = "normal";
 
 			if (this.config.displaySymbol) {
 				var symbolWrapper = document.createElement("td");
+
+				if (this.config.colored && this.config.coloredSymbolOnly) {
```
fewieden: This should be only `this.config.colored` in your case if `coloredSymbolOnly` is false but colored is true, it will not be colored, but should be

john3300: The case of `colored` is true and `coloredSymbolOnly` is false is already handled on line 139 and will mark the entire `tr` row with color, which means the symbol `td` cell will also implicitly have color.

This line catches the case where `colored` is true and `coloredSymbolOnly` is also true, and adds the color to just the `td` cell.

PR #1302, opened 2018-06-03
Ability to fetch compliments from a remote server

idoodler:
Compliments module now can fetch compliments from an external server


MichMich:
Cool! Thanks.

PR #1303, opened 2018-06-03
Add regex filtering to calendar module

kjb085:
Relatively simple changes, just adding regex support to the advanced filtering feature of the default calendar module


kjb085:
Need to rebase from the development branch

PR #1312, opened 2018-06-08
Translation fix

sten82:
just added translation for "feels" in current weather module. It was hardcoded. Added translation for de/en. All other languages have a fallback to english if key is not provided.


sten82:
Just saw this was fixed few days ago.

PR #1347, opened 2018-07-07
Abillity to toggle news article in fullscreen

cederstrom:
Not much to say about this. It's a new notification which enables me to toggle current news article in fullscreen. This is practical if you, as I am, using physical buttons to interact with the Magic Mirror.


cederstrom:
@MichMich I thought this was such a small change so I did not update the changelog. I think that the update to `modules/default/newsfeed/README.md` is enough.

Do you want me to add a post in the changelog anyway?

PR #1364, opened 2018-07-26
Develop Autostart into half screen (F11 keystroke)

hoofta:
I am looking to have MMM auto start into half of the screen on the right to keep chromium open on the right side. I know F11 is the keystroke but I couldn't figure out how to add that into the mm.sh script.


MichMich:
This can be accomplished via the config file. Please check the documentation or refer to http://forum.magicmirror.builders for support and troubleshooting.

PR #1372, opened 2018-08-05
Calendar: Absolute dates do not show absolute even if getRelative and urgency are set to zero

jannekalliola:
Calendar module shows upcoming events with relative "today" and "tomorrow" even if the config defines all dates to be absolute - with the following configuration parameters:

getRelative: 0
timeFormat: "absolute"
urgency: 0

This patch adds an extra if to check whether timeFormat is absolute for events closer than 48 hours.

Further, it might make sense to refactor the whole time formatting function, as it has checks against both configurable and non-configurable values, and it uses different set of conditions for full day events and events with time frame.


jannekalliola:
Failed with the branches somehow.

jannekalliola:
@MichMich Please delete this pull request as faulty.

Issue #1368, opened 2018-08-02
UpdateNotification module contains unlocalizable word "commits"


balassy:
The `modules/default/updatenotification/updatenotification.js` file contains the following lines:
```
var subtextHtml = this.translate("UPDATE_INFO", {
  COMMIT_COUNT: this.status.behind + " " + ((this.status.behind == 1) ? "commit" : "commits"),
  // ….
});
```

Even though the localization files contain an entry for `UPDATE_INFO`, the word `commit` or `commits` always appears in the displayed message.

I recommend removing the words, and substituting only the number into the `COMMIT_COUNT` placeholder. Note however that this may be a breaking change for some existing localization message.

**MagicMirror Version**: 2.4.1

**Expected Results**: Non-English words never appear in localized messages.

**Actual Results**: The word `commit` or `commits` always appear in the update notification message.

Thank you for creating and maintaining this project open source!

fewieden:
What is your suggestion to replace it with, as commit is the technical term for it?

balassy:
Yes, "commit" is the technical term for it, however in some languages that word may change its form when used in plural or in this particular sentence. I can also imagine that a translator can find another word for this term that is less technical for an average user. 

My recommendation is to leave it to the languages and the localization files how they want to use this word and substitute only the `this.status.behind` value into the `COMMIT_COUNT` placeholder.

MichMich:
Feel free to send a non-breaking PR for this.

balassy:
I've just submitted PR #1383 to change this behavior.

rejas:
Since PR #1383 got merged this I think this issue can be closed, @MichMich @balassy ? :-)

balassy:
Yes, this can be closed, PR #1383 fixed the issue. Thank you.

PR #1383, opened 2018-08-18
Making the word "commit" localizable in the UpdateNotification module message

balassy:
This PR changes the UPDATE_INFO message used by the UpdateNotification module: the COMMIT_COUNT placeholder does not contain the word "commit" or "commits" any more, so language files can independently localize them.

Instead of introducing a logic to use the singular or the plural form of the word ("commit" or "commits"), I added the word "commit(s)" to the language files.

Note that some language files already contained this word in some form, that's why not all language files are affected by this change. I can confirm that the Hungarian localization is perfect after this change.

This PR may fix Issue #1368 .

Thanks for reviewing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Mixup between german and spanish translation for newsfeed.
 - Fixed close dates to be absolute, if no configured in the config.js - module Calendar
+- The word "commit" in the message of the UpdateNotification module can be localized.
 - Fix for weatherforecast rainfall rounding [#1374](https://github.com/MichMich/MagicMirror/issues/1374)
 - Fix calendar parsing issue for Midori on RasperryPi Zero w, related to issue #694.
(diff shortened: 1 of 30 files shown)


MichMich:
I appreciate your work, but I find the commit**(s)** not very elegant. Maybe it's better to add two keys to the translation files: `COMMIT` & `COMMITS`. Since the translation system uses fallbacks, these only need to be available in the English translations.

Languages that need localisation can overwrite these.

balassy:
I don't understand how fallbacks could help here, since most languages already have this message localized, and if I understand correctly the translation system can only fallback between languages, but not between resource keys.
Anyways, I created two different resource keys in the language files. What do you think?

MichMich:
Valid reason. I think the only thing left to do is to modify the changelog entry so it reflects a bit better what it does now ... Can you make that change so I can merge it?

balassy:
I've updated the branch and changed the CHANGELOG entry. What do you think?
Thank you.

MichMich:
Thanks! Merged!

balassy:
Thanks for merging!

Issue #1400, opened 2018-09-08
Gzip encoded calendar urls fail to load


ubertao:
**Platform**: Raspberry Pi 3B+, Electron 2.0.0

**Node Version**: v9.11.2

**MagicMirror Version**: v2.4.1

**Description**: 
Some calendar servers (e.g. icloud.com) responses are gzip encoded, calendar module won't load them.

**Steps to Reproduce**: 
Add "https://p33-calendars.icloud.com/holidays/cn_zh.ics" to calendar module.
(Or any url which response is gzip encoded.)

**Expected Results**:
Calendar module displays the calendar events.

**Actual Results**:
Calendar always displays "loading".

**Configuration**: 
```json
{
    module: "calendar",
    header: "China Holidays",
    position: "top_left",
    config: {
        calendars: [
            {
                symbol: "calendar-check-o ",
                url: "https://p33-calendars.icloud.com/holidays/cn_zh.ics"
            }
        ]
    }
}

```

MichMich:
Fix merged.

PR #1401, opened 2018-09-08
Pr calender gzip

ubertao:
Patch calendar fetcher to support gzip encoded calendar.
Solves issue #1400.


MichMich:
Thanks for the PR. It seems the CHANGLOG is based on the current Master version and is missing all entries that are currently in de develop branch. Could you use the develop version in stead?

MichMich:
Thanks.

PR #1405, opened 2018-09-16
Multi-line compliments

ubertao:
Allows multi-line compliments by supporting `\n` (newline) character in compliment text.

For example, in your `config.js`:
```JavaScript
{
    module: "compliments",
    position: "lower_third",
    config: {
        compliments: {
            anytime: [ 'Blackbird singing in the dead of night\nTake these broken wings and learn to fly.' ]
        }
    }
}
```

Will be shown in two lines:
```
Blackbird singing in the dead of night
Take these broken wings and learn to fly.
```

This is ideal for lyrics or poems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 *This release is scheduled to be released on 2018-10-01.*
 
 ### Added
+- Support multi-line compliments
 - Simplified Chinese translation for "Feels"
 - Polish translate for "Feels"
 - French translate for "Feels"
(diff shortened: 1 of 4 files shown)


MichMich:
Thanks for your contribution. Using innerHTML isn't preferred since this could be used to fire malicious code. (IE when an external compliments file is used). Is there a way to refactor this?

ubertao:
Forgot the fact that compliments content can be loaded from an URL. Patch submitted.

MichMich:
Maybe a simpeler approach: how about we just add the `white-space: pre-line;` style to the wrapper? Then no further coding is nessecery.

ubertao:
@MichMich I'm not familiar with CSS, it's definitely simpler.

MichMich:
Awesome work! Thanks! :)

PR #1345, opened 2018-07-05
Adding Font Awesome Version 5

groszejc:
Refer to the following [Issue #1344](https://github.com/MichMich/MagicMirror/issues/1344)

This pull request will give the user the added flexibility to use Font Awesome V5 icons as well as maintaining Font Awesome V4. 

Add FA5 folder to MagicMirror/vendor/node_modules/ file path:
[font-awesome_5.zip](https://github.com/MichMich/MagicMirror/files/2168355/font-awesome_5.zip)

Add FA5 to MagicMirror/vendor/vendor.js file [Line 16]:

`var vendor = {
	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
	"moment-timezone.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
	"weather-icons.css": "node_modules/weathericons/css/weather-icons.css",
	"weather-icons-wind.css": "node_modules/weathericons/css/weather-icons-wind.css",
    "font-awesome.css": "node_modules/font-awesome/css/font-awesome.min.css",
	"font-awesome5.css": "node_modules/font-awesome_5/css/all.css",
	"nunjucks.js": "node_modules/nunjucks/browser/nunjucks.min.js"
};`

Add to styles list in the MagicMirror/modules/default/calendar/calendar.js file [~Line 51]:

`	// Define required scripts.
	getStyles: function () {
		return ["calendar.css", "font-awesome.css", "font-awesome5.css"];
	},`

This would give the user the ability to select symbols from either FA4 or FA5 within the config.js file, just formatting them correctly as shown below:

```
{
    symbol: "calendar", //FA4 format
    symbol: "fas fa-baseball-ball", //FA5 format
},
```

I'm unsure how to update the changelog? First pull request, so bear with me...

Issue #1344, opened 2018-07-04
Upgrade to Font Awesome Version 5


groszejc:
I would like to have font awesome version 5 installed within the following file path for additional icons: 

MagicMirror/vendor/node_modules/

This would also need to have the MagicMirror/vendor/vendor.js file updated to reflect the upgrade as well:

var vendor = {
	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
	"moment-timezone.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
	"weather-icons.css": "node_modules/weathericons/css/weather-icons.css",
	"weather-icons-wind.css": "node_modules/weathericons/css/weather-icons-wind.css",
	**"font-awesome.css": "node_modules/font-awesome_5/css/all.css",**
	"nunjucks.js": "node_modules/nunjucks/browser/nunjucks.min.js"
};

MichMich:
I'm not sure how this would affect existing modules. Any idea?

groszejc:
I have needed to use the upgraded font awesome symbol specification in my calendar module shown below:

`{
						symbol: "fas fa-calendar-alt",
						url: "webcal://www.calendarlabs.com/templates/ical/US-Holidays.ics",
                        color: "#ffffff"
					},`

The default modules I use have not been affected: Clock, Current Weather, Weather Forecast and News Feed. Unsure of the effects of the Compliments, Hello World or Alert modules.

MichMich:
It seems that fa5 isn't backwards compatible to fa4. This means that if we upgrade it, some icons stop working for mirrors that are currently configured. What might be an option, is to make the version of font awesome (in the calendar) configurable, with fa4 as the default.

groszejc:
Doing some testing, it looks like we could have both FA4 and FA5 icons available by adding another line to the MagicMirror/vendor/vendor.js file:

var vendor = {
	"moment.js" : "node_modules/moment/min/moment-with-locales.js",
	"moment-timezone.js" : "node_modules/moment-timezone/builds/moment-timezone-with-data.js",
	"weather-icons.css": "node_modules/weathericons/css/weather-icons.css",
	"weather-icons-wind.css": "node_modules/weathericons/css/weather-icons-wind.css",
    **"font-awesome.css": "node_modules/font-awesome_4/css/font-awesome.min.css",
	"font-awesome5.css": "node_modules/font-awesome_5/css/all.css",**
[...]

PR #1427, opened 2018-10-04
Added Font-awesome 5

P-Storm:
Added font-awesome 5. I updated the default calendar to use 5 with shims to 4. Updated example.

Didn't remove font-awesome 4 for backward compatibility. 

Fixes: #1345 & #1344 .

Possible change: remove font-awesome 4 completely and substitute it with the shim.


MichMich:
Thanks for your contribution! Sorry for the slow response. I've been on a holiday for a while.

PR #1434, opened 2018-10-07
Added in screenshot for the newfeed module

shade34321:
Added in screenshot to the news feed module as the documentation states it wants.


MichMich:
Thanks for your contribution! Sorry for the slow response. I've been on a holiday for a while.

PR #1483, opened 2018-12-14
Always display the degree symbol in the Weather Forecast module

balassy:
This PR changes how the Weather Forecast module displays the values to be consistent with the Current Weather module.

The Current Weather module always displays the degree symbol (&deg;) after the temperature value:
current-weather

However the Weather Forecast module displays only the number:
before-scale-false

By changing the `scale` option to `true` one can configure the module to display "&deg; C":
before-scale-true

After this PR the Weather Forecast module always displays the degree symbol (&deg;):
after-scale-false

And the Celsius or Fahrenheit symbol if the `scale` option is set to `true`:
after-scale-true

This PR also adds documentation about the existing `scale` option to the README of the Weather Forecast module and fixes some typos there.

Thanks for creating and maintaining this project!


MichMich:
Thanks for the PR. Sorry it took me so long to merge. Had to fix a linting issue first.

balassy:
Excellent, thank you!

Issue #1503, opened 2019-01-03
Change weatherforecast and currentweather unit display


fwitte:
Hi everybody,
I just started building/configuring my own MM and found a bug using "default" for the units. The degree symbol should not be displayed. It is degree Celsius and degree Fahrenheit, but not degree Kelvin.

**weatherforecast**
https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/weatherforecast/weatherforecast.js#L145-L158

**currentweather**
https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/currentweather/currentweather.js#L200-L213

https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/currentweather/currentweather.js#L221

I suggest changing the display for both modules like this:

```
var degreeLabel = "";
if(this.config.scale) {
	switch(this.config.units) {
	case "metric":
		degreeLabel += "&deg; C";
		break;
	case "imperial":
		degreeLabel += "&deg; F";
		break;
	case "default":
		degreeLabel = " K";
		break;
	}
}
```

**currentweather**
```
temperature.innerHTML = " " + this.temperature.replace(".", this.config.decimalSymbol) + degreeLabel;
```

**weatherforecast**
```
maxTempCell.innerHTML = forecast.maxTemp.replace(".", this.config.decimalSymbol) + degreeLabel;
minTempCell.innerHTML = forecast.minTemp.replace(".", this.config.decimalSymbol) + degreeLabel;
```

If these changes are okay with you, I will open two PR for this issue.

fewieden:
I will also change it for the new weather module.

fewieden:
Just checked you can disable it in the new weather module by specifying config option `degreeLabel: false`

fewieden:
Handled in https://github.com/MichMich/MagicMirror/pull/1510 for the new weather module.

PR #1505, opened 2019-01-03
Remove degree symbol for Kelvin scale

fwitte:
As discussed in #1503, temperature displays are e. g.

- 25 °C,
- 112 °F or
- 360 K. 

Also, I changed the appropriate source code for both, currentweather and weatherforecast, in order to align with each other.

### DIFF
--- a/modules/default/currentweather/currentweather.js
+++ b/modules/default/currentweather/currentweather.js
@@ -201,13 +201,13 @@ Module.register("currentweather",{
 		if (this.config.degreeLabel) {
 			switch (this.config.units ) {
 			case "metric":
-				degreeLabel = "C";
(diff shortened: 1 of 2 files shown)


MichMich:
Can you please add an updated `CHANGELOG.md`? Thanks!

fwitte:
@balassy Thanks for your suggestion, I did not consider that being a problem. I will adjust the changes to match the changes applied to the new weather module. The behavior will then be:

`scale/degreeLabel: true` -> "°F", "°C", "K"

`scale/degreeLabel: false` ->"°", "°", ""

balassy:
@fwitte Thank you very much, Francesco, I appreciate it.

Issue #1504, opened 2019-01-03
Weather module forecast not working with openweathermap


fwitte:
Hi again,
looking into the weather modules I stumbled across the different configuration keys for essentially the same purpose in the currentweather (`degreeLabel`) and the weatherforecast (`scale`) modules. I think, it would be more intuitive using the same key for both modules. Personally, I prefer the weatherforcast version.

**weatherforecast**
https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/weatherforecast/weatherforecast.js#L28

**currentweather**
https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/currentweather/currentweather.js#L30

Is this change applicable, as this is a change in the API? I will open a PR, if this change is desired, the implementation should then be postponed to the next major release.

Have a nice day!

MichMich:
I agree, but it's important that it's not a breaking change. So the old configuration should still work. Feel free to send a PR to the develop branch.

MichMich:
Also note that we are working on a new weather module which will replace both current weather modules. It's included in the latest release.

fwitte:
Thanks for your hint!

Just one more question: I can not make the forecast work, can you provide a working example or give me a hint, where my error might be? Thank you and have a nice day!

This is my code, it works with `type: "current"`. Providing the `initialLoadDelay` parameter, does not change the behavior.
```
		{
			module: "weather",
			position: "top_right",
			header: "Vorhersage",
			config: {
				type: "forecast",
				degreeLabel: true,
				location: "mylocation",
				locationID: "mylocationid",  //ID from http://www.openweathermap.org/help/city_list.txt
				apiKey: "myweatherid"
			}
		},
```

MichMich:
@fewieden Any idea?

fewieden:
The problem is that you need to specify another weatherendpoint for openweathermap forecast like `weatherEndpoint: "/forecast/daily"`

fewieden:
If is set the initial load delay to 30 seconds it is delayed by 30 seconds. Be aware of that you need to specify th timeframe in milliseconds like: `initialLoadDelay: 30000`

More informations here: https://github.com/MichMich/MagicMirror/tree/master/modules/default/weather#general-options

fwitte:
Thanks for your help, still does not work. I dug into the code and found the following:

[...]

PR #1488, opened 2018-12-27
Weather refactoring

fewieden:
- Fixed openweathermap
- Fixed darksky
- Implemented forecast for both
- added indoor values
- documentation
- ...

current
forecast

PR #1506, opened 2019-01-04
Fix openweather forecast in new weather module, fetch daily data

fwitte:
See #1504 

Example

```
{
	module: "weather",
	position: "top_right",
	header: "Wetter",
	config: {
		type: "current",
		degreeLabel: true,
		location: "Oldenburg",
		locationID: "2857458",
		apiKey: "myapiKey"
	}
},
{
	module: "weather",
	position: "top_right",
	header: "Vorhersage",
	config: {
		type: "forecast",
		degreeLabel: true,
		weatherEndpoint: "/forecast",
		location: "Oldenburg",
		locationID: "2857458",
		apiKey: "myapiKey"
	}
},
```
looks like this now:
grafik

Note that for the forecast to work with a **free API-subscription** from openweathermap, you must provide `weatherEndpoint: "/forecast"`! `weatherEndpoint: "/forecast/daily"` returns error 401, as it is used for the 16-day forecast and requires the Startup-subscription at minimum!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Updated
 
 ### Fixed
+- Weather forecast now works with openweathermap in new weather module. Daily data are displayed, see issue [#1504](https://github.com/MichMich/MagicMirror/issues/1504).
 
 ## [2.6.0] - 2019-01-01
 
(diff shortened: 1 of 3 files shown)


vincep5:
I use openweathermap (the free version) and have access to forecast/daily.  I must be grandfathered in.  In any case, if the project wants to support both /forecast/daily and /forecast, then these commits will not work with /forecast/daily.  It does not have forecast.rain["3h"]

fewieden:
As mentioned in my initial pull request https://github.com/MichMich/MagicMirror/pull/1488#issuecomment-450208710 there was currently no support for newer api keys in the forecast.

This mr broke now the support for the old api keys.

PR #1525, opened 2019-01-09
Fix conflict between font awesome versions as per #1522

### LINKED ISSUE #1522: Font Awesome icons not working after v2.6.0 upgrade
Hi @MichMich 

Great work as ever getting another update out - hope you managed to celebrate the New Year whilst completing the release!

I have uncovered an issue with the Font Awesome 5 support introduced in v2.6.0

**MagicMirror Version**: v2.6.0

**Description**: 
When amodule includes `"fontawesome.css"` (i.e. font awesome 4) and another module (e.g. the default calendar) includes `"font-awesome5.css", "font-awesome5.v4shims.css"` then unexpected results can occur depending on which order they are included in `config.js`.

**Steps to Reproduce**: 
A: Modify `config.js` file in a v2.6.0 install so it only includes definitions for the default calendar module and then MMM-Strava (in that order). In this instance `fa-hashtag` and `fa-trophy` display as expected, but `fa-arrows` (a renamed icon) does not.

B: Modify `config.js` file in a v2.6.0 install so it only includes definitions for the default calendar module and then MMM-Strava (in that order). Change the default config for calendar to use `symbol: "volleyball-ball"`. In this instance `fa-hashtag` and `fa-trophy` display as expected, but `fa-arrows` (a renamed icon) and `fa-volleyball-ball` (a new icon) do not.

C: Change `config.js` to include definitions for MMM-Strava then the default calendar module (in that order) and all icons appear.

D: Change `config.js` to omit the default calendar module (i.e. only include MMM-Strava) and all icons appear.

**Expected Results**: 
All fa4 and fa5 icons should appear in all instances

**Actual Results**: 
If the _fa5_  (e.g. calendar) module is omitted from `config.js`,  the _fa4_ module works as expected.
If the  _fa5_  (e.g. calendar) module is added to `config.js` before the _fa4_ module, the config for the _fa5_ module cannot use new fa5 icons AND the _fa4_ module cannot use renamed icons

**Additional Notes**: 
The release notes for v2.6.0 suggest the new FA5 functionality is backwards compatible. However, it feels like this is either a bug or a breaking change. 

A workaround is to release an update to the _fa4_ module so that any icons are renamed per the [font awesome 4 upgrade notes](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes). However, to do so means the third party module will depend on v2.6.0 (i.e. `requiredVersion` must be increased to `2.6.0`) thus preventing existing users who choose not to upgrade from taking advantage of improvements to a module.

Also, if another _fa4_ module is added to the config after the upgraded _fa5_ modules, the problem recurs.

ianperrin:
Proposal to fix #1522 

* Remove font awesome 4 from dependencies in vendor/package.json
* Add custom font-awesome.css to allow multiple css to be included for one vendor i.e. both the full font awesome css (all.min.css) and the v4-shim css (v4-shims.min.css) for backward compatibility
* Modify vendor.js so font awesome 5 (with backwards compatibility) is used for all modules
* Revert font awesome references in default calendar module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed unhandled error on bad git data in updatenotiifcation module [#1285](https://github.com/MichMich/MagicMirror/issues/1285).
 - Weather forecast now works with openweathermap in new weather module. Daily data are displayed, see issue [#1504](https://github.com/MichMich/MagicMirror/issues/1504).
 - Fixed analogue clock border display issue where non-black backgrounds used (previous fix for issue 611)
+- Fixed compatibility issues caused when modules request different versions of Font Awesome, see issue [#1522](https://github.com/MichMich/MagicMirror/issues/1522). MagicMirror now uses [Font Awesome 5 with v4 shims included for backwards compatibility](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#shims).
 
 ### New weather module
 - Fixed weather forecast table display [#1499](https://github.com/MichMich/MagicMirror/issues/1499).
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -51,7 +51,7 @@ Module.register("calendar", {
 
 	// Define required scripts.
 	getStyles: function () {
-		return ["calendar.css", "font-awesome5.css", "font-awesome5.v4shims.css"];
+		return ["calendar.css", "font-awesome.css"];
 	},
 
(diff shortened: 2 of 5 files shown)


MichMich:
Nice work. Could you add an updated changelog reflecting the fix?

MichMich:
Whoops, didn't want to close the PR. Sorry. Can you send a new PR (including the changelog)?

ianperrin:
> Nice work. Could you add an updated changelog reflecting the fix?

Done

MichMich:
Thanks!

Issue #1377, opened 2018-08-10
Install script error with pm2 autostart


pyrosmiley:
Was gonna only make a forum post, but I've run into this with fresh Stretch installs on two separate pis, a 3 and a new 3b. Updated everything then immediately ran MM install script on each, has happened the same each time.

**Platform**: Rpi 3 & 3b, brank spankin' new & updated Raspbian install. Node v9.11.2

**Node Version**: 9.11.2

**MagicMirror Version**: 2.4.1

**Description**:On fresh raspbian installs on each machine, I ran the MM install script, which I have not had issues with before. It actually ran just fine as far as I could tell the first time, but after a reboot of the machine, I noticed the following: MM was not autostarting, I could *start* it from a terminal, but electron never actually opened, and anytime I tried to start it with pm2, the pm2 daemon was not running, and MM was not saved as something to watch.

 It turns out that the `sudo su -c "env PATH=$PATH:/usr/bin pm2 startup linux -u pi --hp /home/pi"' line was in fact the culprit. when I ran that manually, it gave back an `Unknown platform /init name` error, with the explanation that `PM2 detected systemd but you precised linux`. So I manually ran `pm2 startup` and the command it gave me to input instead was `sudo env PATH=$PATH:/usr/bin /usr/lib/node_modules/pm2/bin/pm2 startup systemd -u pi --hp /home/pi`. Everything worked fine after that.

**Steps to Reproduce**: Fresh raspbian install, `sudo apt-get update && sudo apt-get upgrade -y`, then execute MM install script from `curl`.

**Expected Results**: I expected it to work like it's supposed to

**Actual Results**: pm2 startup config error

**Configuration**: N/A

**Additional Notes**: I am not sure if an updated version of pm2 or Raspbian or whatever has altered the command for the setup that's used in the script, but it didn't work until I replaced `linux` with `systemd`. I originally had assumed that my pi3 was the culprit (it was one of the ones that ran into the crazy CPU errors with electron, so I know it's already a special boy) but then the same thing happened with a 3b that was fresh out of the box.

Here is the full console output wherein:
1. I attempt to run startup command, and it fails.
2. Startup command successfully set up
3. MM pm2 config successfully loaded

```
pi@mirror:~ $ sudo su -c "env PATH=$PATH:/usr/bin pm2 startup linux -u pi --hp /home/pi"
[PM2] Init System found: systemd
-----------------------------------------------------------
 PM2 detected systemd but you precised linux
[...]

PR #1531, opened 2019-01-12
Fixing raspbian installation script

fdahms:
Hi,

i tried to install MM2 on a fresh raspbian 3b+ with "2018-11-13-raspbian-stretch-full.zip". But the automatical installation script does not work. After google the problems, i find the solutions and add it to the installation script. I tested it severall times with a fresh image and it it works out of the box.
Like a installation script should work

Fixing issue #1377


MichMich:
Awesome! Thanks!

Issue #1462, opened 2018-11-11
Not install on arm64


vvzvlad:
**Platform**: 
```
orangepi@orangepizeroplus2:~$ grep . /etc/*-release
/etc/armbian-image-release:# PLEASE DO NOT EDIT THIS FILE
/etc/armbian-image-release:BOARD=orangepizeroplus2-h5
/etc/armbian-image-release:BOARD_NAME="Orange Pi Zero Plus 2"
/etc/armbian-image-release:BOARDFAMILY=sun50iw2
/etc/armbian-image-release:VERSION=5.65
/etc/armbian-image-release:LINUXFAMILY=sunxi64
/etc/armbian-image-release:BRANCH=next
/etc/armbian-image-release:ARCH=arm64
/etc/armbian-image-release:IMAGE_TYPE=stable
/etc/armbian-image-release:BOARD_TYPE=conf
/etc/armbian-image-release:INITRD_ARCH=arm64

orangepi@orangepizeroplus2:~$ uname -m
aarch64
orangepi@orangepizeroplus2:~$ getconf LONG_BIT
64
orangepi@orangepizeroplus2:~$ dpkg -s libc6 | grep Arch
Architecture: arm64
Multi-Arch: same
orangepi@orangepizeroplus2:~$ npm -v
6.4.1
orangepi@orangepizeroplus2:~$ node -v
v10.13.0
orangepi@orangepizeroplus2:~$ 

```

**Node Version**: 
v10.13.0

**MagicMirror Version**: 
2.5.0, master

**Description**: 
Unable to install MM on the arm64 platform, npm install fails with:
`Error: Failed to find Electron v1.7.0 for linux-arm64 at https://github.com/electron/electron/releases/download/v1.7.0/chromedriver-v1.7.0-linux-arm64.zip`

Same problems:
https://forum.magicmirror.builders/topic/6377/error-on-clean-installation-using-npm-install-electron-chromedriver-1-7-1
https://forum.magicmirror.builders/topic/8786/rpi3-npm-install-failed-to-find-electron-v1-7-0-for-linux-arm64
https://forum.magicmirror.builders/topic/6926/mm-installation-error-by-electron-arm64/2

**Steps to Reproduce**: npm install

```
orangepi@orangepizeroplus2:~/MagicMirror2$ npm install
[...]

PR #1537, opened 2019-01-14
up spectron version to 4.0.0(fix arm64 install)

vvzvlad:
Fix error in #1462 issue, not install MM on arm64 platform(orangepi, etc). 

According to the advice in the issue, I raised the spectron version to 4.0.0 (previously it was 3.8.0) which fixed the installation problems. However, I do not know how it will affect other platforms. On Raspberry PI, it runs and works correctly, but that's all I can check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Updated
 - Bumped the Electron dependency to v3.0.13 to support the most recent Raspbian. [#1500](https://github.com/MichMich/MagicMirror/issues/1500)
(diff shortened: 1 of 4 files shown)


MichMich:
Hi, Unfortunately Travis can't run the checks using these changes. Not sure what's the issue since I'm not using spectrum myself. Any Idea how to solve this? (Can't merge without this issue being fixed.)

vvzvlad:
I ran npm install and included the updated package-lock.json in the PR. It turns out that package-lock.json can not be fixed manually.

MichMich:
Unfortunately it's still not passing. Any idea?

MichMich:
Closed due to inactivity. Feel free to reopen.

PR #1549, opened 2019-01-21
Currentweather dev

weirdkid:
Added a new config option called showExtraWeatherInfo because I wanted to 
display temp and feelsLike and nothing else. The onlyTemp option was the only 
way to turn off the extra weather info above the temp, but it also blocked the feelsLike.

onlyTemp still works, but it should be considered deprecated, since the same
behavior/results can be achieved by setting showExtraWeatherInfo and 
showFeelsLike to false.


MichMich:
Ins't it a better idea to create booleans for all the elements? I'm nut sure the current proposal makes it much cleaner. 

Also note we are working on a replacement weather module. It might be beter to implement the change in that module:

https://forum.magicmirror.builders/topic/9335/default-weather-module-refactoring

weirdkid:
That's what I was thinking too, but the boolean show methods are now at least lined up with the internal functions -- i.e. you have one big function called "addExtraInfoWeather". I didn't feel like refactoring the whole thing for this. Just wanted to separately show "Feels Like".  At least now the functions don't clobber each other.  It sounds like you're moving on from this version of the module anyway.  Cheers.

kitten4224:
Stay weird my friend

MichMich:
Since we are moving to a new module anyway I'm closing this PR for now. Feel free to reopen.

PR #1564, opened 2019-02-09
Added autoLocation and autoTimezone option for weather modules and clock respectively.

jacob-ebey:
Added autoLocation option for the current weather and weather forecast modules.


jacob-ebey:
Going to get this reviewed at any point?

jacob-ebey:
Thanks, I hope this is of use :)

PR #1577, opened 2019-02-18
fix relative date fulldate events to use start of day to start of day diff… issue 1572

sdetweil:
default calendar in relative mode uses a time based 'fromNow()' to compare against start of DAY for fulldate events..  this causes elapsed day rounding down when the current hours get past noon. 

change to use start of day from(date only, as it is midnight, start of day)


sdetweil:
> This change should also be included in "urgency mode". See line 300, where "timeformat:absolute" falls into the urgency timeframe and gets relative.

done

PR #1586, opened 2019-02-24
Feature/alert fontawesome

Tom-Hirschberger:
A small change to getStyles function in alert.js makes it possible to use Font Awesome icons in alerts. I want to display alerts which contain the bluetooth icon for example. The Font Awesome icons are already include in the vendor stuff so they only need to be included.


Tom-Hirschberger:
Sorry, needed to change the source to develop instead of master

PR #1589, opened 2019-02-25
Regularly trigger ADD_CALENDAR to ensure calendar fetcher is running

hudashot:
Currently an `ADD_CALENDAR` notification is only sent once when the calendar module is started in the UI. This notification creates a calendar fetcher on the server side that is used to regularly update the calendar.

When MagicMirror is running in `serveronly` mode, the server process can be restarted even while the UI keeps running. When the server process restarts, it launches with no calendar fetchers running, and there is currently no way to start them again other than to refresh the UI.

This change just fires an `ADD_CALENDAR` event every fetch interval. When such event is received by the server that already has a fetcher running, it's [effectively ignored](https://github.com/MichMich/MagicMirror/blob/de57daa3cd881ce1a14b88307bf61e8109879c81/modules/default/calendar/node_helper.js#L72) (i.e. no duplicate fetcher is started), however if there is no fetcher running (because the server has been restarted) this makes sure it gets started.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Calendar: Fix relative date handling for fulldate events, calculate difference always from start of day  [#1572](https://github.com/MichMich/MagicMirror/issues/1572)
 - Fix null dereference in moduleNeedsUpdate when the module isn't visible
 - Calendar: Fixed event end times by setting default calendarEndTime to "LT" (Local time format). [#1479]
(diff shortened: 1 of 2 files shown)


hudashot:
Updated `CHANGELOG.md`.

MichMich:
It might be a cleaner idea to broadcast a message to all modules whenever the socket connection is (re-) established. In that case you can broadcast ADD_CALENDAR whenever this happens. For now your solution seems fine and we can always change the impementation... Thanks!

PR #1591, opened 2019-02-27
Updated auto location logic

jacob-ebey:
Had some issues with the auto-location implementation. This is a bit hacky, but if someone wants to clean up the "use the same result" code and you are ok with adding a new express route, this works a lot better. 

Not having global notifications from a node_helper to every frontend module really limits how this can be implemented, and this is currently the best I could come up with.

### DIFF
--- a/js/server.js
+++ b/js/server.js
@@ -14,6 +14,7 @@ var ipfilter = require("express-ipfilter").IpFilter;
 var fs = require("fs");
 var helmet = require("helmet");
 var Utils = require(__dirname + "/utils.js");
+var request = require("request");
 
 var Server = function(config, callback) {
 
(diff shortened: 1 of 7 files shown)


jacob-ebey:
The frontend modules that use the auto-location don't even really need to do it in the node_helpers at this point, they could just refetch if they recieve a 500 from the /location url. I don't have the drive right now to clean this up more though.

MichMich:
Hi Jacob, 

The amount of changes necessary for this function feels way to impactful for the number of users that will use it. Most of the code would fit perfectly fine in a separate 3rd-party module which would broadcast the location to any interested module. In that case the default modules only need to listen to a `LOCATION` notification, which would set the location of the module if a location for that specific module is not provided by the config.

I suggest you take this route, remove all the current auto location code from the MagicMirror² core (effectively reverting this PR: https://github.com/MichMich/MagicMirror/pull/1564) and only add the `LOCATION` notification listeners for the modules that need it. Then you can create your own module that does the heavy lifting. 

This would allow other modules to leverage the 'LOCATION' notification as well, both by broadcasting one, as well as listing to the notification.

Since the current solution is a bit hacky, and doesn't add a lot of functionality to the core, I'll remove your current implementation before I merge develop into master if no alternative solution is provided via a PR.

I hope you understand. 

Looking forward to your PR using the above suggested route. Thanks!

MichMich:
FYI: I already created a PR to revert PR 1564: https://github.com/MichMich/MagicMirror/pull/1592

PR #1604, opened 2019-03-15
Change spelling of parameter to match documentation

dgburr:
According to the documentation, this setting is called "useKMPHWind" (not "useKMPHwind").  Changed the implementation rather than the documentation since this seems to be more consistent with the naming scheme used for the other parameters in the currentweather module.


MichMich:
JavaScript is a case-sensitive. This means that this would be a breaking change. For anyone currently using `useKMPHwind` this would change te behaviour. Therefor I can't merge this PR.

dgburr:
In that case, would it be possible to update the documentation so that it matches the implementation?

MichMich:
Yeah that sounds logical. Feel free to send a PR. Thanks!

PR #1628, opened 2019-04-01
Add includePastEvents global and calendar-specific settings

kolbyjack:
For my module MMM-MonthlyCalendar, I need access to events that have already past.  This change will allow users to broadcast those events, while still only displaying future events in the default calendar module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Russian translation for “Feels”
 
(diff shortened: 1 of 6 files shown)


MichMich:
Thank!

Im afraid most users don't understand what "event broadcasts" are which will result in users expecting to show events from the past when setting `includePastEvents` to true. Maybe we should use an other config name?

dxfan227:
is there any way to "broadcast only todays events"? I would really love this cause my other module ( google route) is trying to show me a route for an event in 20 days

MichMich:
This PR seems to have merge conflicts. Please rebase the PR. Thanks for your contribution.

MichMich:
Merged! Sorry for taking so long.

PR #1647, opened 2019-04-08
Added option to show location of calendar events

retroflex:
Added option to show location of calendar events. Default is false. If showLocation=true and a location has been entered for the event, it is shown below:

image


MichMich:
Awesome! Thanks

retroflex:
Thanks for the quick merge @MichMich!

It can be useful at times. I have a common calendar for my kid's soccer practices, where the locations vary a lot. This change makes my life a lot easier. Hope it will be useful for others as well.

The location also has its own class name so it can be styled in custom.css.

dxfan227:
cool! what's the cleanest way to update to this option?

PR #1657, opened 2019-04-14
Develop

vladakru:
Serbian latin and Serbian cilylic translation.

[translations.zip](https://github.com/MichMich/MagicMirror/files/3077338/translations.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Finnish translation for "Feels" and "Weeks"
+
 ### Updated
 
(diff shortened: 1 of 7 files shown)


vladakru:
how?

sdetweil:
so, the model here is that you fork the base repo.. now u have your copy... 

you clone the copy to your system and make updates.. (commits), and then push those commits back to  your copy... then you select the original (Mich's) and select the create Pull Request button, and pint the new to your fork..  in your case this should only include the new translation files.. (not a zip)

note that git allows multiple parallel instance of the source, these are called branches.
Mich wants all PR's to be submitted against the dev branch..    so you should switch to dev  before u start making changes on your cloned fork.

PR #1666, opened 2019-04-24
Update config.js.sample

kevbodavidson:
Changes the URL to show the default calendar since the old URL was bad


kevbodavidson:
Changes the URL to show the default calendar since the old URL was bad

MichMich:
Please add an updated Changelog. Thanks.

Issue #1228, opened 2018-03-25
Calendar Bug? Reccuring fullDayEvent (Birthday) - today is wrong (one day before real date)


eckonator:
**Platform**: Raspberry Pi 3, OwnCloud based Calendar
**Node Version**: v6.13.1

**MagicMirror Version**: Latest JavaScript Version (v2).

**Description**: I have a Birthday Calendar stored on my OwnCloud Instance. There are hosted multible Calendards. Dates that are based on a time works well. But full day events (birthdays) that are recurring are shown on my magic mirror one day before as today instead of the real today-date.

**Steps to Reproduce**: Just include this ICS File:

--------------------------------------------------

BEGIN:VCALENDAR
VERSION:2.0
CALSCALE:GREGORIAN
PRODID:-//SabreDAV//SabreDAV//EN
X-WR-CALNAME:Test-Kalender
X-APPLE-CALENDAR-COLOR:#c274e7
BEGIN:VEVENT
CREATED:20180325T120003
DTSTAMP:20180325T120003
LAST-MODIFIED:20180325T120003
UID:N25F8LCTXDSGTY7VDVEY
SUMMARY:Birthday People 1
CLASS:PUBLIC
STATUS:CONFIRMED
DTSTART;VALUE=DATE:20180326
DTEND;VALUE=DATE:20180327
RRULE:FREQ=YEARLY
END:VEVENT
BEGIN:VEVENT
CREATED:20180325T120033
DTSTAMP:20180325T120033
LAST-MODIFIED:20180325T120033
UID:V7QP0NKOH8MBCAP90JPAK6
SUMMARY:Birthday People 2
CLASS:PUBLIC
STATUS:CONFIRMED
DTSTART;VALUE=DATE:20180327
DTEND;VALUE=DATE:20180328
RRULE:FREQ=YEARLY
END:VEVENT
END:VCALENDAR

--------------------------------------------------

**Expected Results**:
Today output should be on the right day (but it's one day before)

**Configuration**: 

`
{
            module: "calendar",
            header: "Birthdays",
            position: "top_left",
            config: {
                calendars: [
                    {
                        //symbol: "calendar-check-o ",
                        symbol: "birthday-cake ",
                        maximumEntries: '2',
                        maximumNumberOfDays: '14',
                        url: "webcal://link-to-my-calendar.ics"
                    }
                ]
            }
 }
`

Hope someone can help or have a fix for me? This project is so cool :)

eckonator:
[...]

Issue #1105, opened 2017-12-26
iCAL problem - All upcoming appointments dates wrong - Month value


AWSW2017:
Hello there and merry Christmas to you,

I think I have a reproducible problem with the calendar module. 

**Platform**: (Raspberry Pi 3, Mac, Recent Linux NOOBS).

**Node Version**: Recent version installed with your automatic install.

**MagicMirror Version**: Recent version installed with your automatic install.

**Description**: 

The problem is, that the months are not fetched correctly. Today is the 26th of Dezember 2017 and it shows appointments for the next upcoming days that are in several weeks or months. Please see the attached picture. The fetched day seems to fit, but the months seem to be set to this month or mostly to January 2018. 
![cal](https://user-images.githubusercontent.com/34800740/34356666-158f14e2-ea40-11e7-8bfa-fbc3ec704a7e.png)

My system is set to German language and time zone Europe/Berlin.

I already did a clean install of the recent MagicMorror version and just changed the URL in the config.js to my public ical calendar. The problem is then reproducible again.

Thanks for your help in advance. :)

**Configuration**: What does the used config.js file look like? Don't forget to remove any sensitive information!

Just changed the sample URL to my public iCal URL.

AWSW2017:
Hello,

I think I found the reason for this. All not correctly shown appointments are recurring ones and they have 2 notifications: 1 for the day of the appointment itself and one 1 day before. 

Removing the one of the day before of the appointment helped to remove most of the not correctly shown entries in MagicMirror. For the remaining ones I just needed to open and store the appointment again.

Thanks for your help and for this nice project. :)

Kind regards

MichMich:
Ah. For now it would be great to add that to the documentation. It would be great if you could send a PR for this. I'll close this for now. Feel free to reopen!

E3V3A:
@MichMich 
I don't see why you would want to close the issue if there has not been any PR fixing it?
I just found a similar issue reported.

MichMich:
@E3V3A Reopened.

E3V3A:
FWIW:  
I recall seeing an ical file that had some additional timezones and multiple entries. Perhaps have a close look at that to rule out bad or confusing ical files.

pinsdorf:
Another aspect is that the calendar module cannot cope with iCal URLs from Microsoft Live/Hotmail/Outlook calendars. I tried this and got a parsing exception because of start/end times containing timezone information in the date/time string. 

[...]

Issue #1070, opened 2017-10-24
Events from Google Calendar showing at incorrect dates, or not showing at all


Txukie:
Edited:
I have been using a Google Calendar of birthdays to test this module out. Sometimes events show up but incorrectly, like several months in advance although they seem to show up on the same day, as in an event created for the 25th of March will actually show up as occurring on the 25th of October for example.

```
{
        module: "calendar",
        header: "Calendario",
        position: "top_left",
        classes: 'default everyone',
        config: {
            maximumEntries: 5,
            maximumNumberOfDays: 100,
            calendars: [
                  {
                           symbol: "exclamation-circle",
                           url: "webcal://calendar.google.com/calendar/ical/cf9euqob2a0tqe2409jffnpk2o%40group.calendar.google.com/private-a8870caf6c3ca46a785e7fd1fec0dcb5/basic.ics"
                   }
                  ]
              }
}
```

Just FYI, my locales are set as follows:

```
LANG=es_ES@euro
LANGUAGE=
LC_CTYPE="es_ES@euro"
LC_NUMERIC="es_ES@euro"
LC_TIME="es_ES@euro"
LC_COLLATE="es_ES@euro"
LC_MONETARY="es_ES@euro"
LC_MESSAGES="es_ES@euro"
LC_PAPER="es_ES@euro"
LC_NAME="es_ES@euro"
LC_ADDRESS="es_ES@euro"
LC_TELEPHONE="es_ES@euro"
LC_MEASUREMENT="es_ES@euro"
LC_IDENTIFICATION="es_ES@euro"
LC_ALL=
```

morozgrafix:
@Txukie it may not be a good idea to post link to your private calendar in a public forum.

Txukie:
You're totally right, let me make another account and copy this calendar accross and I will update my report accordingly.

javilandoando:
I have same issue.
I don't know why although on google calendar any birthday remind is displayed properly, using this module I get a reminder related to a birthday on a different month although day, 21st for example, is correct.

Does any body find how to correct this?

kleinmantara:
The same problem with me. I also use Google Calendar and all-day entries are displayed at the wrong time.

mattmurf:
I'm getting garbled times..  The date is showing correctly.  

This was working and broke at some point in the past couple months.

Example:
Quiz blah, Blah, Blah
Today at AMb0oo12/4/2017ut1

matthewpmangus:
[...]

Issue #576, opened 2016-12-20
deleted weekly reoccurring calendar items still showing in calendar


dinkybluebug:
**Platform**: RPI 3

**Node Version**: 6

**MagicMirror Version**: 2

**Description**: Weekly recurring calendar items that have been deleted, are still showing in the calendar - which is odd as when I view my icloud calendar in my phone, my partners phone, online, and in outlook the items are not there.  But they are definitely showing still in my magicmirror calendar.

Ive restarted - but this doesnt fix issue.  Ive added in new items into the calendar, and these show - so its definitely working - not sure why deleted items  are still showing.

Any ideas as to why this might be?

roramirez:
When you download the calendar by url the events are still there?

FatBasta:
I have the same issue.

Platform: RPI 3
Node Version: 6
MagicMirror Version: 2
Ical from google calendar

It seems that if you have a recurring event and remove one from the middle, it still shows in the mirror.
For example a recurring event for weeks 18 - 25, and you delete week 23, google calendar no longer shows it, but the mirror still does.

dinkybluebug:
Yes that's what's happened to me. One event in the middle of a reoccurring appointment was deleted but it still shows on mirror - even though Apple calendar doesn't show it

roramirez:
Hi!, some way to debug the problem is what happend with data into url you called. Maybe the events still there and flagged "deleted" or something like data.

I think a good way is downloaded the file .ical the moment this happend.

morozgrafix:
I believe skipped recurring events (ones that have EXDATE timestamp) aren't skipped in ical.js https://github.com/peterbraden/ical.js/issues/75 which leads to problem in rrule dependency (which may be unmaintaned).

Here is a part of my ics file that has 2 skipped events from the recurring series. They still show up in MM (note the EXDATE values for 12/24 and 12/31 which were "deleted" events):
```
BEGIN:VEVENT
DTSTART;TZID=America/Los_Angeles:20160924T083000
DTEND;TZID=America/Los_Angeles:20160924T093000
EXDATE;TZID=America/Los_Angeles:20161224T083000
EXDATE;TZID=America/Los_Angeles:20161231T083000
RRULE:FREQ=WEEKLY
DTSTAMP:20161231T105941Z
UID:FCE045E3-5271-442E-9E96-D5A1DE0958D3
CREATED:20160917T153742Z
DESCRIPTION:
LAST-MODIFIED:20161231T022115Z
[...]

PR #1701, opened 2019-06-13
Calendar: repeating event rrule & exdate fixes

mattdb:
Previously, if a repeating event had specific recurrences deleted (exdate) or modified (rrule), the calendar would not display those changes, instead still showing canceled events, or events with their originally scheduled times or titles. This PR makes handling these types of exceptions to calendars much more robust.

In order to do this, I:
* updated calendar's vendored ical.js
* made extensive changes to the calendar fetcher within the logic branch where an event has an rrule attached
* extracted title-parsing logic into a function (`getTitleFromEvent`) since modified events may need to re-parse their title.

I believe this will solve and/or make progress on a slew of calendar-related issues, including:
* #1105 (Which seems to have become the omnibus issue for the ical parser)
* #1070 
* #1228 
* #565 
* #576

The updated ical.js should also help with comma-separated EXDATEs, as noted in peterbraden/ical.js#84 and peterbraden/ical.js#76


mattdb:
Doh! That conflict just got pulled in as I was writing up the PR. I'll work on resolving.

mattdb:
Investigated the conflict-- #1699 made manual changes to the vendored ical.js/node-ical.js file, but similar improvements were already included in the upstream file that I am updating here. In my opinion, best to stay with the "official" upstream version as long as it resolves the issue the "unofficial" modification was addressing.

mattdb:
Updating to the up-to-date ical.js does offer one complication: it uses rrule, rather than rrule-alt, which MagicMirror manually switched to a while back. Since that switch, rrule-alt has not been updated, but rrule has seen quite a bit of updating. 

I propose we hop back on the rrule train to handle bug fixes like all of the myriad ical formatting quirks, etc.

However, we can't remove rrule-alt from the package's dependencies because some 3rd party modules have come to expect it (MMM-CalendarExt is the one I ran into, at least).

MichMich:
> I propose we hop back on the rrule train to handle bug fixes like all of the myriad ical formatting quirks, etc.
Agree. 

>However, we can't remove rrule-alt from the package's dependencies because some 3rd party modules have come to expect it (MMM-CalendarExt is the one I ran into, at least).
What's you suggested solution for this?

MichMich:
I've merged your PR for now. We can always add extra PR's to solve open or created issues.

PR #1715, opened 2019-06-30
update installer to handle non-pi and other issues

sdetweil:
Updated installer
give non-pi2+ users (pi0, odroid, jetson nano, mac, windows, ...) option to continue install
use current username vs hardcoded 'pi' to support non-pi install
check for npm installed. node install doesn't do npm anymore
check for mac as part of PM2 install, add install option string
update pm2 config with current username instead of hard coded 'pi'
check for screen saver config, "/etc/xdg/lxsession", bypass if not setup


MichMich:
Great PR! Thank you very much!

PR #1744, opened 2019-07-28
Skip from worktree the css/custom.css:

roramirez:
On the next release the css/custom.css will rename to
css/custom.css.sample

This change run git instructions to detach the file from own local
repository. This instructions are called in untrack-css.sh file from
run-start.sh and npm postinstall step

Reference #1540


sdetweil:
see my upgrade script 

https://forum.magicmirror.builders/topic/10859/new-upgrade-script-ready-for-testing

which saves and restores custom.css across pull requests (actually fetch/merge)

PR #1753, opened 2019-08-11
Tests 404 vendors

roramirez:

PR #1755, opened 2019-08-18
Fixing weatherforecast module not displaying rain amount if using fal…

buxxi:
Added support for showing the amount of rain when using fallback endpoint.

The rain-property contains values like on fallback endpoint and not a numeric value directly:
`
"rain":{"3h":1.813}
`

So I made it sum it from all of that days forecasts since that rain forecast is for 3 hours and it returns a forecast for every 3 hours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Updatenotification module: Properly handle race conditions, prevent crash.
 - Send `NEWS_FEED` notification also for the first news messages which are shown
 - Fixed issue where weather module would not refresh data after a network or API outage [#1722](https://github.com/MichMich/MagicMirror/issues/1722)
+- Fixed weatherforecast module not displaying rain amount on fallback endpoint
 
 ## [2.8.0] - 2019-07-01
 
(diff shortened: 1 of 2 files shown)


MichMich:
@roramirez Do you have time to check this travis issue?

lavolp3:
TheTravis CI build fails in tests/e2e/vendor_spec.js


`		Object.keys(vendors).forEach(vendor => {
			it(`should return 404 HTTP code for vendor https://localhost/"${vendor}"`, function() {
				urlVendor = "http://localhost:8080/" + vendors[vendor];
				request.get(urlVendor, function (err, res, body) {
					expect(res.statusCode).to.equal(404);
				});
			});
		});
	});
`

MAYBE the only error is in the urlVendor variable? Trying to make a suggestion.

lavolp3:
@roramirez can you please check if this needs to be updated from your commit #1753? I guess there is a `/vendor` missing in the path of the urlVendor variable which stops the Travis CI build here.

roramirez:
@lavolp3 I pushed the https://github.com/MichMich/MagicMirror/pull/1767, you should sync (merge) on your own branch the last version of `develop`. Thanks to notice the failure.

MichMich:
Merged!

PR #1773, opened 2019-09-28
Timestamp log

qistoph:
This PR adds a timestamp to the logging/stdout. This is helpful for troubleshooting. It works for output on the console (like with `node serveronly`) and for the log file `.pms/logs/mm-out.log` when using pm2.

The timestamp is in UTC, which is how the module I've found for this, uses. If that's an issue and you'd prefer local time, please let me know and I'll see what I can do.


MichMich:
Travis still failing. Can't merge this. Sorry.

qistoph:
Finally managed to get it fixed

PR #1790, opened 2019-10-25
left electron left in dependencies section vs pr 1788

sdetweil:


MichMich:
Isn't electron required anymore?

sdetweil:
its in the optionalDependencies section..   
```
  "optionalDependencies": {
    "electron": "^3.0.13"
  },
````

that way if there is no version for armv6 (or whatever) , install won't fail, and serveronly mode will still work.

right now, install fails. leaving a mess

don't want it in the mandatory dependencies too

sdetweil:
and run-start.sh has been changed to support non-electron systems by using chromium-browser
and intentional serveronly mode (config option)  and also on non-gui systems

MichMich:
Ah, missed that. Thanks.

PR #1793, opened 2019-10-27
added latitude, longitude and timezone options to global config options

prydonian:
There are a lot of modules that require location data, so you have to add it multiple times in the config file. It makes sense to have that data as global variables that any module can access.


MichMich:
Please fix the conflicts. Thanks! :)

MichMich:
Please fix linting errors. Thanks.

MichMich:
Closed due to inactivity.

PR #1801, opened 2019-11-03
Develop

richardwochenende:
I'm very new to all of this, but I made something I guess could be useful for others. If this is the wrong place for it, please be kind and just move it to the right one.

I just installed my MM and thought it could be nice to see by color if the "feels like" temperature on currentweather-addon is higher or lower than the actual color outside, so I added following lines to the code (from start of line 257):
`
if(this.temperature = this.feelsLike){
feelsLike.className = "same";     // feels as warm/cold as is
} else if(this.temperature > this.feelsLike){
feelsLike.className = "colder";     // feels colder than is
} else if(this.temperature < this.feelsLike){
feelsLike.className = "warmer";     // feels warmer than is
} else {
feelsLike.className = "error";     // something's wrong with that
}
`
for this, i added the classes "same", "colder", "warmer" and "error" to custom.css with only color-values. so e.g. if the "feelsLike"-temperature is colder, it's shown in a light blue. if it's warmer, it's shown light red and so on.

just thought it may help someone wanting the same (Y)


MichMich:
Please send a PR to the Develop branch. Thanks!

Issue #1800, opened 2019-10-31
electron downlevel on Raspian Buster


sdetweil:
**Platform**: Place your platform here... give us your web browser/Electron version *and* your hardware (Raspberry Pi 2/3, Windows, Mac, Linux, System V UNIX).
Raspian Buster 

**Node Version**: Make sure it's version 0.12.13 or later.

**MagicMirror Version**: Now that the versions have split, tell us if you are using the PHP version (v1) or the newer JavaScript version (v2).
2.9.0
**Description**: Provide a detailed description about the issue and include specific details to help us understand the problem. Adding screenshots will help describing the problem.
electron cannot start 
```
 magicmirror@2.9.0 start /home/pi/MagicMirror
> sh run-start.sh

/home/pi/MagicMirror/node_modules/electron/dist/electron: /lib/arm-linux-gnueabihf/libdbus-1.so.3: no version information available (required by /home/pi/MagicMirror/node_modules/electron/dist/electron)
/home/pi/MagicMirror/node_modules/electron/dist/electron: /usr/lib/arm-linux-gnueabihf/libnss3.so: version `NSS_3.22' not found (required by /home/pi/MagicMirror/node_modules/electron/dist/electron)
```
**Steps to Reproduce**: List the step by step process to reproduce the issue.

**Expected Results**: Describe what you expected to see.

**Actual Results**: Describe what you actually saw.

**Configuration**: What does the used config.js file look like? Don't forget to remove any sensitive information!

**Additional Notes**: Provide any other relevant notes not previously mentioned. This is optional.

Electron project says upgrade electron is only fix.

MichMich:
So, what version do we need to bump to? Feel free to send a PR.

sdetweil:
working on that. 6.0.12 seems good..

MichMich:
As discussed, the version change of Electron was reverted because of Travis issues. Electron definitely needs to be updated, so we need to find a way to make it work with Travis.

@roramirez Any idea?

MichMich:
For anyone running Buster, if you want to switch to a working Electron version, use the following command:
```
npm install electron@6.0.12
```

Please report if this solves your issue. Thanks!

khassel:
concerning the electron issue, you are talking about the failed tests [here](https://travis-ci.com/MichMich/MagicMirror/builds/142809334)?

[...]

PR #1806, opened 2019-11-08
New electron version

sdetweil:
#1800 

change the electron version required


sdetweil:
@MichMich how do I fix this build error? where package-lock.json has the old electron version number in it?
we should never save the package-lock file anyhow

johankvint:
> @MichMich how do I fix this build error? where package-lock.json has the old electron version number in it?
> we should never save the package-lock file anyhow

Remove your local package-lock.json and run `npm install`. It should generate a new lock file with your dependencies

sdetweil:
And then I check that new package-ock file. Duh. I hadn't don't nom install, just wanted to change the text file

sdetweil:
so, what is the problem here ? all I did was npm install, and the commit the package.json file

package.json
```diff
@@ -56,7 +56,7 @@
     "time-grunt": "latest"
   },
   "optionalDependencies": {
-    "electron": "^3.0.13"
+    "electron": "^6.0.12"
```
chadasapp: 6.0.x isn't supported anymore? I think it should be 6.1.x, if not 7.1.x

sdetweil:
where is the comment about 6.1/7.1 vs 6.0.12? says it should be here, but I can't find it

we  WERE at 3.0.13... so its a big jump.. there are side effects.. as armv6l is no longer supported in 6.0.12.. we haven't tested beyond 6.0.12...  but we need to upgrade to fix other problems

rvoitenko:
I have tested setup with electron 6.x and 7.x. The latest 6.1.5 works just fine, but not the 7.x
The platform was running debian buster and node 12.x

sdetweil:
@rvoitenko   thanks.. as I suspected..

sdetweil:
what api.js is this error from?

```
  1) Electron app environment
       "before each" hook for "should open a browserwindow":
     unknown error: Cannot read property 'name' of undefined
  Error: An unknown server-side error occurred while processing the command.
      at execute(<Function>, "require") - api.js:73:26
```

this sounds like the 
		webPreferences: {
			**nodeIntegration**: false,  (which should be true to enable require to work)

ezar:
Can you apply?

sdetweil:
@ezar what does that mean?

MichMich:
@sdetweil Unfortunately I had to revert this PR. This was the one causing all the issues. I'll make an issue that we need to update Electron, but Travis needs to pass before we can.

If anyone want to run buster, they can always manually install the updated electron version, using:

```
npm install electron@6.0.12
```

sdetweil:
It all rubs ok locally. So something to do with Travis setup. Not electron problem

MichMich:
Absolutely. Electron probably gives issues with the used bistro on Travis. But we need Travis to release a new version and for me to merge PR's. So for now, this is the best option.

sdetweil:
Yeh. We need instructions on how to update the Travis process as we enhance the system

sdetweil:
Please reopen my electron  version issue

Issue #1804, opened 2019-11-07
Weather Forecast, decimalSymbol parameter does not affect precipitation format


GattoJohnny:
**Platform**: RPi 3 B+

**Node Version**: v10.17.0

**MagicMirror Version**: v2.9

**Description**: In Weather Forecast module, predicted amount of rain value uses DOT as decimal separator, regardless of the decimalSymbol parameter. Max and min temperature formats follow the value set in decimalSymbol, precipitation format does not.
It is a cosmetic error, not a big deal. However, it would be nice to have it displayed correctly in those countries where comma is used as the decimal separator.

**Steps to Reproduce**: Add decimalSymbol parameter to your config file, and give it any values other than the default **'.'**.

**Expected Results**: See the attached screen shots.

**Actual Results**: See the attached screen shots.

**Configuration**: 

**Global parameters in config.js**
language: "hu",
timeFormat: 24,
units: "metric",

**Module specific parameters in config.js**
config: {
...
	updateInterval: 10 * 60 * 1000, // every 10 minutes
	initialLoadDelay: 2500,
	maxNumberOfDays: 5,
	decimalSymbol: ",",
	units: "metric",
	fade: false,
	showRainAmount: true,
	colored: false,
	appendLocationNameToHeader: false,
	scale: false
}

![weatherforecast_decimalsymbol-01](https://user-images.githubusercontent.com/47010649/68397302-4218a680-0173-11ea-9b3e-66d0148e9d29.jpg)

![weatherforecast_decimalsymbol-02](https://user-images.githubusercontent.com/47010649/68397322-493fb480-0173-11ea-9cbc-92bda8b50258.jpg)

MichMich:
Fix merged into develop.

PR #1808, opened 2019-11-13
Use config.decimalSymbol when displaying the predicted amount of rain in weatherforecast

FritzJay:
Fix for issue #1804 

I copied logic from the min/max temperature section and used it to display the correct decimal symbol in the rain amount section.

**Module specific parameters in config.js**
> config: {
>   ...
>   decimalSymbol: ",",
>   showRainAmount: true
> }

image


MichMich:
Thanks!

PR #1815, opened 2019-11-22
fix regression and add support for commented lines in config.js

sdetweil:
update scripts before release


MichMich:
Hi Sam, could you please add an updated changelog? Thanks!

sdetweil:
we need this

sdetweil:
This does not fail when run locally

PR #1835, opened 2019-12-20
Added a padding between days in dateheader mode

akraus53:
Very tiny addition to the calendar module: in dateheader-mode there's now a padding between the days, which makes it look nicer IMO


akraus53:
Sorry - didn't use delevop branch

PR #1847, opened 2019-12-31
Fixes some events not being recognized as full day events

oemel09:
Some full day events were shown as normal events starting at 00:00 o'clock.

- Fixes some events not being recognized as full day events


oemel09:
I've one more thing I'd suggest to implement / fix.
Full day events spanning over multiple days such as school holidays are displayed as: 8 days ago.
I'd prefer them to be shown as: 'today', if they are still ongoing.

Can I just change it or should I implement a config variable for this?

MichMich:
Sounds like a nice fix. Feel free to send a PR.

PR #1827, opened 2019-12-09
Fix helper being erased accidentally

sdetweil:
* Does the pull request solve a **related** issue?
#1758 
* If so, can you reference the issue?
* What does the pull request accomplish? Use a list if needed.
move the node_help npm to a listed official dependency
* If it includes major visual changes please add screenshots.

PR #1849, opened 2019-12-31
Fix CI

roramirez:
Revert commit introduced in #1827. The change introduce a failure for all e2e tests. 

This change revert the commit 2fbedca746655f38fa89a6436afa888f88d518db

--
Happy new year ("20".repeat(2))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Timestamps in log output.
 - Padding in dateheader mode of the calendar module .
-- Move node_helper module to dedicated github repo, to prevent being erased accidentally.
 - New upgrade script to help users consume regular updates installers/upgrade-script.sh.
 - New script to help setup pm2, without install installers/fixuppm2.sh.
 
--- a/modules/node_modules/node_helper/index.js
+++ b/modules/node_modules/node_helper/index.js
@@ -1,0 +1,125 @@
+/* Magic Mirror
+ * Node Helper Superclass
+ *
+ * By Michael Teeuw http://michaelteeuw.nl
+ * MIT Licensed.
+ */
+
+var Class = require("../../../js/class.js");
+var express = require("express");
+var path = require("path");
+
+NodeHelper = Class.extend({
+	init: function() {
+		console.log("Initializing new module helper ...");
+	},
+
+	loaded: function(callback) {
+		console.log("Module helper loaded: " + this.name);
+		callback();
+	},
+
+	start: function() {
+		console.log("Starting module helper: " + this.name);
+	},
+
+	/* stop()
+	 * Called when the MagicMirror server receives a `SIGINT`
+	 * Close any open connections, stop any sub-processes and
+	 * gracefully exit the module.
+	 *
+	 */
+	stop: function() {
+		console.log("Stopping module helper: " + this.name);
+	},
+
+	/* socketNotificationReceived(notification, payload)
+	 * This method is called when a socket notification arrives.
+	 *
+	 * argument notification string - The identifier of the notification.
+	 * argument payload mixed - The payload of the notification.
+	 */
+	socketNotificationReceived: function(notification, payload) {
+		console.log(this.name + " received a socket notification: " + notification + " - Payload: " + payload);
+	},
+
+	/* setName(name)
+	 * Set the module name.
+	 *
+	 * argument name string - Module name.
+	 */
+	setName: function(name) {
+		this.name = name;
+	},
+
+	/* setPath(path)
+	 * Set the module path.
+	 *
+	 * argument path string - Module path.
+	 */
+	setPath: function(path) {
+		this.path = path;
+	},
+
+	/* sendSocketNotification(notification, payload)
+	 * Send a socket notification to the node helper.
+	 *
+	 * argument notification string - The identifier of the notification.
+	 * argument payload mixed - The payload of the notification.
+	 */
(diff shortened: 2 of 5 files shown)


sdetweil:
NO.. we need the helper fix.. need to fix CI some other way..

sdetweil:
the vendor tests still fail.

MichMich:
@sdetweil I Will test this PR tonight. If this fixes the issue I’ll merge it since we need Travis to complete successfully in order to release 2.10 today (according to release schedule).

sdetweil:
How do we fix the  testing cache to have the helper in the main node_modules instead the modules/node_modules?

MichMich:
Will try tonight. Don’t want to have the helper in an external repo anyway. Might need to push that change onto the next release.

sdetweil:
I think the helper being erased is a big problem. I think it is a security fix that is erasing. If modules not listed but installed, they are removed

MichMich:
Will look into it. Hope I can fix it before release. 

Vriendelijke groeten,
Michael Teeuw

---
Sent from my iPhone.

Michael Teeuw
Xonay Media
www.xonaymedia.nl
06 51 71 36 15

sdetweil:
Another user reported that install failed to install the helper file. Which
means it was cloned and then erased on npm install

https://forum.magicmirror.builders/topic/11883/black-screen-on-startup

MichMich:
Reading that, I’m not confident it was caused by an NPM install. At at least not an non install ran in the correct folder. But non the less I will try to look for a solution. First priority is to get Travis to work again.

sdetweil:
Ok. I am for stable system. Test system later. The vendor stuff has been
failing for quite a while now.
Only the weather tests are affected so far by the node_helper.

sdetweil:
We have a whole quarter to solve the Travis  problems

MichMich:
I can't release without Travis. Releases are blocked. And for a good reason. Wel work on a solution now.

MichMich:
@sdetweil the node helper is no longer part op de node_modules folder. Now looking for the weather tests fails.

roramirez:
> I think the helper being erased is a big problem. I think it is a security fix that is erasing. If modules not listed but installed, they are removed


I can understand why will be the better if we has a node helper in external repository and what kind of issue could fix include in this way. We should keep this code in main code of project, its to more easy maintent this code in this repo.

sdetweil:
I asked multiple times for Mich to create the repo on his GitHub, but got no action. So I created it on mine. 

Mich is welcome to move it to his.
I  trying to stop user acceptance problems

MichMich:
This is already fixed. The node helper is already outside the modules folder (in /js to be exactly). It is aliased so you can just use the same require statement. (Non breaking). 

I want to keep everything in one repo to keep it simple. This accomplishes that.

sdetweil:
As for the Travis problem there are electron version dependencies on some of the libraries used by the test cases. Spectron, for example.
I updated that one to match the electron version.
Don't know if there are other lib problems too.

Tests run locally after install work great.
So how do we get those libs updated in Travis environment?  

There is no change mgmt guidance for the Travis tests

sdetweil:
Helper in js. good! I didn't see that as a possibility.

roramirez:
> As for the Travis problem there are electron version dependencies on some of the libraries used by the test cases. Spectron, for example.
> I updated that one to match the electron version.
> Don't know if there are other lib problems too.
> 
> Tests run locally after install work great.(weather works, vendor still fails, but still reports ok)
> So how do we get those libs updated in Travis environment?
> 
> There is no change mgmt guidance for the Travis tests

What O.S, npm and node versions are you using? 

Could you paste here the package.json in your local environment?

Also, `npm list`

sdetweil:
Will do. Node 10.1x.0, npm 6.1x.0
Tried on raspian, Ubuntu, 

All tests fail on macOs Catalina. 10.15.2

I am away for most of the rest of the day
I posted a new package-lock with this pr. which solved the problem originally (Travis went from failing to passing)

I did a clone, checkout for develop, npm install, then ran two testcase sets

sdetweil:
looks like the screen setup for travis has changed

https://electronjs.org/docs/tutorial/testing-on-headless-ci

when I do export DISPLAY=:0
and then run the npm run test:e2e (locally) it works 
but not the vendor tests..

the vendor tests are because the port didn't get set when using the js/electron.js  launch
I added debug to the vendor_spec.js and here is the output of the before, launching js/electron.js
the port is not set 

  Vendors
starting js/electron
electron path=/home/sam/MagicMirror/node_modules/.bin/electron
new electron app =[object Object]
parms={"host":"127.0.0.1","**port":9515**,"quitTimeout":1000,"startTimeout":5000,"waitTimeout":5000,"connectionRetryCount":10,"connectionRetryTimeout":30000,"nodePath":"/usr/bin/node","path":"/home/sam/MagicMirror/node_modules/.bin/electron","args":["/home/sam/MagicMirror/tests/e2e/../../js/electron.js"],"chromeDriverArgs":[],"env":{},"workingDirectory":"/home/sam/MagicMirror","webdriverOptions":{},"requireName":"require"}

if I change the get urls to the 9515 port, the it works.. 

this didn't work  (tests/e2e/from vendor_spec.js)
	describe("Get list vendors", function () {

		before(function () {
			process.env.MM_CONFIG_FILE = "tests/configs/env.js";  // < ----
		});

that before() is never called. 
and if u move the process.env. to before the app.start() it doesn't work there either..

sdetweil:
the untouched e2e test also works on macOS catalina after doing export DISPLAY=:0
i had to change the address for the vedor tests to port 9515, but it worked then

Issue #1856, opened 2020-01-02
new update script fails trying to overwrite the dumpactivemodules.js script


sdetweil:
V2.10 upgrade-script.sh checks to see if the little javascript code file is present,
and if not copies it... 

but 2.10 supplies it now, and the git merge fails if it is there already 
need to move the test to the point of needing it.

PR #1858, opened 2020-01-02
fix upgrade-script

sdetweil:
#1856 
move download of dumpactivemodules.js to after merge, download may not be needed anymore


sdetweil:
this is a really a hotfix, need before 2.11, so that the upgrader works

MichMich:
Before I release a hot fix (which I really try to avoid in this project due to all the update notifications), it might be worth to discuss this issue: https://github.com/MichMich/MagicMirror/issues/1859

sdetweil:
the script wants to make sure all  installed and configured modules which have dependencies are also updated..  npm install for the base does not do this, and results in a dead mirror.  This is the number 1 error on each update cycle. 

now, how to determine the configured modules...   i wrote quick js to load the config and loop thru the module array to check the disabled flag. most of us developers have many modules installed but only a few enabled. 

except this new file is not present on the users system ..  so the script downloads it if needed. 

except, it is downloaded BEFORE the repo updates are done, and because the file is now included in the repo, we have a merge conflict.   this change moves the download to just before running if not already present. 

without this fix, the script will always fail. 
the PR that include the script also updated the readme on the update process to use the script.
so the github site says use the script. 

there is a second fix, about to push, when there are changed files, the user has the opportunity to save them (via git stash). except stash requires the users email and name. most users have never set that, and so stash will fail, and update will fail.  (pull would have failed too due to the conflicts, and the user has no idea what to do)..  i use a temporary name and email, just for these commands, but in two cases the $ variable identifier was left off.  so the stash command still fails.

sdetweil:
this build failed again for travis problems.. only a script file was changed... and weather tests failed again.. unrelated to my change

MichMich:
Restarted Travis. Succeeded without issues.

MichMich:
If we need to release this before 2.11, we might want to make a separate branch, so all other changes that are part of the 2.11 release won't be in the hot fix release. Do you want me to create a separate branch so we can release this one time hotfix?

pgporada:
There's an unfixed bug that I think this PR can address. The variable `$ logfile` should be `${logfile}` at https://github.com/MichMich/MagicMirror/pull/1858/files#diff-481c0639db88cf5d7ddd89304e0bf726R247

sdetweil:
separate branch, I am ok with that

pgporada:
Sure, I'll get it in the morning.

sdetweil:
@pgporada thanks for the report../  I've added this fix to the separate script repo now too

https://github.com/sdetweil/MagicMirror_scripts

sdetweil:
@MichMich   as the original PR for these changes also changed the upgrade instructions in the readme, do you want to revert those back to the pull approach as well?

MichMich:
Sam, just to be sure (before I go trough the process of making a new branch and preparing a 2.10.1 release): Is this change also necessary if we continue the work discussed in #1860? Does this save an issue for people currently upgrading? Or does this solve an issue for future upgrades? 

Thanks for your patience!

sdetweil:
The installer is also broken, so I think u need to do this.

MichMich:
I created the branch based on the current master branch. This PR probably needs to be rebased so it doesn't include the previous commits made on the develop branch.

Thanks!

sdetweil:
i think I want to change the focus of this hotfix.  

if I change the readme, so that it points to my repo for automated install and update, etc, then 
I can fix whatever, whenever.  no code changes in the repo itself.  and U can work towards #1860

sdetweil:
I propose close this pr for #1856 instead. all future PR's against  install/update rejected, redirected to external repo

Issue #1860, opened 2020-01-03
Discussion: Simplify the core.


MichMich:
Since I open sourced MagicMirror², I have received many awesome pull requests adding a lot of nice features to the Magic Mirror core. I am very grateful for that! Unfortunately, this comes with a downside: the application gets a lot more complicated.

The reason I open sourced MagicMirror² was because it is a perfect starter project for newcomers. Getting their feet wet in javascript and software development in general (see [manifesto](https://github.com/michmich/magicmirror#manifesto)). With the current complexity, this project is slowly moving away from this.

I also noted that the amount of bash code is increasing, while the main intention of version 2 was to try to do everything in 1 program language: javascript. (Opposed to v1, which was a combination of languages like PHP, Javascript en in some cases Python).

Looking at other OS projects, there are some beautiful examples on how the project can be focused at what it does best. One example I like to mention is OctoPrint/OctoPi. Software to control your 3D printer. While @foosel does a great job to maintain the core application (OctoPrint), the packaged Rasberry Pi image (OctoPi) is maintained in a different repo by a different user (@GuySoft).

If we use that same approach there are a lot of things we can take out the core repo, and let other users start their own repo's to accomplish the same features while separating concerns.

In my opinion, the following parts could initially be taken out of the core repo, and taken over by other repo's:

- Installer script
- Update script
- Start Script
- PM2 integration
- Docker Package

Of course, we won't take out these features as long as no other solutions are in place. As an install and update replacement, we need to write a good manual. 

The added benefit of this approach is that the release schedule of the above features is no longer tied to the quarterly release schedule of MagicMirror².

With the new approach I think it's good to set a few guidelines to which the future versions should adhere:
- Javascript, HTML, CSS only.
- No compile scripts (ie. webpack, sass, less)
- Focus (only) on Raspberry support.
- Install/Run/Update using default commands: `npm install`, `git pull`, etc. Which will help people to learn and understand these commands.

Since this project would have never been such a success without the help of the community, I want to discuss this approach before taking any action. So I'm looking forward to your input.

Please let me know what you think with an open mind.  

Thanks! Michael.

# EDIT: Provisional TO DO List

Based on the discussion below, I made a provisional to do list which could be the base for the simplification of this repository. The to do list is not set in stone, and could change based on the feedback in this issue.

- [X] Remove installer scripts (make sure an external repository is available as a replacement). 
- [X] Remove update scripts.
- [X] Remove start script.
[...]

PR #1861, opened 2020-01-03
fix run-start.sh: if running in docker-container, just start electron…

### LINKED ISSUE #1859: new run-start.sh fails detecting xwindows in docker setup
**Platform**: Raspberry Pi 3, running magicmirror as docker image with `npm start` ([full setup](https://gitlab.com/khassel/magicmirror))
**Node Version**: 13.5.0
**MagicMirror Version**:  v2.10.0

**Description**: 
With the new release v2.10.0 the script run-start.sh was changed. The new version tries to handle some special cases falling back to `node serveronly` instead of starting electron.

- I think such a fallback does not lead to the desired results and make things more complicated. So if someone calls `npm start` he should know what he is doing and there should be no fallback to `node serveronly`. May you can create warnings but in the end the electron call should be executed.

- In a docker setup the following line (from run-start.sh) is not retrieving the expected result:

````
# check for xwindows running
xorg=$(pgrep Xorg)
````

`xorg` is empty because the Xorg-process is not running in the container but on the host and the container cant see this process. 

So calling `npm start` results here in starting `node serveronly` instead of electron and the mirror remains black.

**Possible Solutions**:
- my favorite: Remove the fallback to `node serveronly`, may display warnings.

- use `DISPLAY` variable setting your `xorg` variable (untestet):

remove the line `xorg=$(pgrep Xorg)` and use this to set `xorg`:

````
if [ -z "$DISPLAY" ]; then #If not set DISPLAY is SSH remote or tty
	export DISPLAY=:0 # Set by default display
else
	xorg=xwindowsrunning
fi
````

So Is there any chance to get a patched release fixing this bug?

Thanks,

Karsten.

khassel:
Fixes #1859

With this fix the `run-start.sh` script differentiate if running in a docker-container or not:

-  running inside a docker-container: Just start electron
-  running not in a docker-container: Do things as before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Force declaration of public ip adress in config file (ISSUE #1852)
+- Fixes `run-start.sh`: If running in docker-container, don't check the environment, just start electron (ISSUE #1859)
 
 ### Updated
 
--- a/run-start.sh
+++ b/run-start.sh
@@ -2,66 +2,74 @@
   # use bash instead of sh
 ./untrack-css.sh
 
-if [ -z "$DISPLAY" ]; then #If not set DISPLAY is SSH remote or tty
-	export DISPLAY=:0 # Set by default display
-fi
-# get the processor architecture
-arch=$(uname -m)
-false='false'
+if grep docker /proc/1/cgroup -qa; then
+  # if running in docker, only start electron
 
-# get the config option, if any
-# only check non comment lines
-serveronly=$(grep -v '^\s//'  config/config.js | grep -i serveronly: | awk '{print tolower($2)}' | tr -d ,\"\')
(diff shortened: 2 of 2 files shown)


MichMich:
@sdetweil What are your thoughts on this PR?

sdetweil:
I would have captured the result of the grep and added it to the other decision tree choices, and launched electron at the end like it does, rather than add another launch

my old training is don't duplicate

khassel:
> I would have captured the result of the grep and added it to the other decision tree choices, and launched electron at the end like it does, rather than add another launch
> 
> my old training is don't duplicate

thought about this solution first, but decided against. So the first  `if` is a decision between `docker` or `nodocker` and your part is untouched in the `nodocker`-section. The duplicate is one row.

And this fix is only needed until #1860 is implemented.

But if it is consensus here i can change it.

MichMich:
I'll merge it for now, but chances are that the script is removed in the next version. Thanks for your contribution.

PR #1863, opened 2020-01-04
Updated raspberry.sh installer script to save apt-get log output to a file 

cherifya:
- Updated raspberry.sh installer script to save apt-get log output to a file that can be tailed
- Fixed an issue where the raspberry.sh installer script can get stuck when there are packages that need to be upgraded


MichMich:
Please add the changelog entries to the 2.11 release. The 2.10 release is already published. Thanks!

sdetweil:
why the separate log entry?  the  update=$(sudo apt get ... 2>&1)  captures all the stdout and error content,
and then the echo $update >>$logfile should save it to the existing log

sdetweil:
i added the --assume-yes to the upgrade.. thanks

sdetweil:
this fix (--update=yes) is in the scripts repo

MichMich:
Closed since we are moving this script.

PR #1875, opened 2020-01-12
Add different GitHub repos to issue template

bastilimbach:
This commit adds the official docker image repo as well as the installation scripts repo to the issue template to avoid the creation of issue which don't belong to the MagicMirror core.

I send this PR to the master branch as this has nothing to do with the actual product. If you want I can also change it back to point to the development branch.


bastilimbach:
Should I add something to the changelog? This isn't really something I would consider to be part of the changelog.

MichMich:
I merged it Into the dev branch, so it will be updated on the next release. Thanks

PR #1884, opened 2020-01-18
fix: Issue #1798 - fixing recurrent calendar events crosstime DST

kblankenship1989:
**Issue #1798**  - converting to Local times for JS Date Objects before submitting to 'rrule' library.  Issue appears to be in `Rrule.between()` function where date and time are broken apart and then put back together, it looses reference to timezone with the UTC time passed.

**Fix** - trick the function to thinking local time is UTC time and then convert back to UTC time after array of recurring events is returned.

**Files impacted**
- modules/default/calendar/calendarfetcher.js
- modules/default/calendar/vendor/ical.js/node_ical.js

**Example:**

- Event every day @ 7:30 CST created on 10/16/2019
- Daylight Savings Time was applicable at this time, so 7:30 CST -> 12:30 UTC
- Convert start time to 7:30 UTC (2:30 CST) by adding the timezone offset
- Generate array of start dates using 7:30 UTC
- Convert all start dates back to regular 7:30 CST by subtracting timezone offset (-5 /-6 respectively based on Daylight Savings Time)

**Additional changes** - Ran lint and inadvertently changed 8 other files.


modules/default/calendar/vendor/ical.js/node-ical.js
```diff
@@ -26,6 +26,17 @@ exports.parseFile = function(filename){
 
 var rrule = require('rrule').RRule
 
+function getLocaleISOString(date) {
+	var year = date.getFullYear().toString(10).padStart(4,'0');
+	var month = date.getMonth().toString(10).padStart(2,'0');
```
kurtzettel: getMonth() returns a month between 0 and 11, not between 1 and 12 so this creates an invalid date.

kblankenship1989: thanks for the catch.  I just put up PR #1914 to address this.

PR #1885, opened 2020-01-18
add sun/moon rise/set times

dtreskunov:
Icons become bright when the object is above the horizon. Also shows
time until next sunset/sunrize and percent of moon illumination.

image


hangorazvan:
On old iPad Safari _toLocaleString_ not working
`183   const illuminatedFractionString = moonIllumination.fraction.toLocaleString(undefined, {style: 'percent'});`
I think better way is:
`183   const illuminatedFractionString = moonIllumination.fraction.toFixed(2)*100 + '%';`

dtreskunov:
@hangorazvan thanks, I'll try to get it fixed soon. Additionally, it appears that moonrise/moonset times are not quite correct...

hangorazvan:
Here another error, on 15 February morning:

WhatsApp Image 2020-02-15 at 20 21 13

PR #1898, opened 2020-01-23
Improves backwards compatibility for Safari (fixes #1897)

### LINKED ISSUE #1897: 'Undefined' module header on safari < 11
Hey guys,
So i use MM as server only set up in a docker container and an old ipad as a screen. After last couple updates i noticed that my setup is all messed up with `UNDEFINED` headers above every module. So i check it on my laptop (chrome, w10) and everything is fine. Then i checked it on an iPhone (current ios) and also everything is fine. Thought this might be browser support issue so i tried browserstack.com for legacy safari versions and sure enough:
Safari 11.1:
image

Safari 10.1:
image

Safari 9.1:
image

I don't know if thats really a large issue for the whole community or how much we want to support older browsers but sure enough it worked fine circa december 2019. Thanks!

sejka:
fixes #1897

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [2.11.0] - Unreleased (Develop Branch)
 
 *This release is scheduled to be released on 2020-04-01.*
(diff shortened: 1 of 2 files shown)


MichMich:
Somehow this PR keeps failing in Travis. Could you please resubmit this PR?

PR #1900, opened 2020-01-24
Removed whitespaces and typo fix

mmphego:
This PR removes any whitespaces from the config sample file as well as fix a typo.


MichMich:
Let's not go into the tabs vs spaces discussion. ;) Also the changes contain language errors.

config/config.js.sample
```diff
@@ -3,97 +3,97 @@
  * By Michael Teeuw http://michaelteeuw.nl
  * MIT Licensed.
  *
- * For more information how you can configurate this file
+ * For more information how you can configuration this file
```
MichMich: This doesn't seem correct.

config/config.js.sample
```diff
@@ -3,97 +3,97 @@
...
  * See https://github.com/MichMich/MagicMirror#configuration
  *
  */
 
 var config = {
-	address: "localhost", // Address to listen on, can be:
-	                      // - "localhost", "127.0.0.1", "::1" to listen on loopback interface
-	                      // - another specific IPv4/6 to listen on a specific interface
-	                      // - "0.0.0.0", "::" to listen on any interface
-	                      // Default, when address config is left out or empty, is "localhost"
```
MichMich: It can also be empty. Why remove this?

mmphego: Figured left out and empty more or less meant the same thing. I'll revert the change. Thanks

config/config.js.sample
mmphego: I completely butchered the Queen's language. Will fix it on next commit.

MichMich:
Thanks for the updates! I think every other file is using tabs. So we might want to use tabs in the configuration as well.

mmphego:
Instead of messing up the commit history, I will close this PR and create a new one with only the grammar fixes :+1:

Issue #1842, opened 2019-12-29
Issues with Vendor_spec tests (e2e)


sdetweil:
when running the automated tests, in the E2E section, there is a test to insure download of the vendor provided javascript files. 

each access is returning  ECONNREFUSED 127.0.0.1:8080
but that should be correct and does work when manually executed (via curl or wget)

there are two sections of the test... get from correct path
and get from bad path... 

both sections fail the same
but report passed

v 2.9.0 or above

reproduce
install selected version, npm install
then run 
npm run test:e2e

[screen shot](https://www.dropbox.com/s/08eq118ez6b29cf/2019-12-29%20%281%29.png?dl=0)

MichMich:
Just to be sure: the test passes, but the test isn't correct?

sdetweil:
the test does not fail, but the actual items to be tested are not actually tested

sdetweil:
code changes, vendor.js
```
	before(function () {
		console.log("starting js/electron");
		return helpers.startApplication({
			args: [__dirname + "/../../js/electron.js"]
		}).then(function (startedApp) { app = startedApp; console.log("parms="+JSON.stringify(app.getSettings()))});  // dump settings after app created
	});

				urlVendor = "http://localhost:9515/vendor/" + vendors[vendor];
```
output
```
starting js/electron
parms={"host":"127.0.0.1",**"port":9515**,"quitTimeout":1000,"startTimeout":5000,"waitTimeout":5000,"connectionRetryCount":10,"connectionRetryTimeout":30000,"nodePath":"/usr/bin/node","path":"/home/sam/MagicMirror/node_modules/.bin/electron","args":["/home/sam/MagicMirror/tests/e2e/../../js/electron.js"],"chromeDriverArgs":[],"env":{},"workingDirectory":"/home/sam/MagicMirror","webdriverOptions":{},"requireName":"require"}
    Get list vendors
      ✓ should return 200 HTTP code for vendor "moment.js"
      ✓ should return 200 HTTP code for vendor "moment-timezone.js"
      ✓ should return 200 HTTP code for vendor "weather-icons.css"
      ✓ should return 200 HTTP code for vendor "weather-icons-wind.css"
      ✓ should return 200 HTTP code for vendor "font-awesome.css"
      ✓ should return 200 HTTP code for vendor "nunjucks.js"
      ✓ should return 200 HTTP code for vendor "suncalc.js"
[...]

PR #1903, opened 2020-01-25
Revert error logging introduced in 1e97b5c

roramirez:
Ref https://github.com/MichMich/MagicMirror/issues/1842#issuecomment-578114552


khassel:
my idea was changing the 2 (now deleted) lines from `mlog.pending` to `mlog.error`

PR #1912, opened 2020-01-31
Support HTTPS Magicmirror

normankong:
To support HTTPS with issue ticket #1911  . Thanks a lot !

Issue #1911, opened 2020-01-31
Contribute Request : Support HTTPS for chromecast


normankong:
@MichMich 

I would like to use the MM on my Pi which cast the display into my Nest Hub. However, chromecast require an HTTPS connection instead of HTTP so I would like to contribute MM to support HTTPS. Please allow me to submit the pull request accordingly. Many thanks !

PR #1915, opened 2020-02-01
Support HTTPS Magicmirror (resubmit)

normankong:
To support HTTPS with issue ticket #1911 . 
The original PR mark closed #1912 
Thanks a lot !


config/config.js.sample
```diff
@@ -21,6 +21,10 @@ var config = {
 	                                                       // or IPv4 range of 192.168.3.0 --> 192.168.3.15 use CIDR format :
 	                                                       // ["127.0.0.1", "::ffff:127.0.0.1", "::1", "::ffff:192.168.3.0/28"],
 
+	useHttps: false,			// Support HTTPS or not
+	httpsPrivateKey: "",		// HTTPS Private Key path
```
MichMich: I think it’s a good idea to explain that these variables are optional and only necessary if you want to use https. Agree?

normankong: Yeah.... I will update it with below remark. Is that fine ? 

useHttps: false,		// Support HTTPS or not, default "false" will use HTTP
httpsPrivateKey: "",		// HTTPS private key path, only require when useHttps is true
httpsCertificate: "",		// HTTPS Certificate path, only require when useHttps is true

MichMich:
Yes, perfect.

Issue #1840, opened 2019-12-29
Issues with Weather Tests. - Tests currently disabled.


MichMich:
For some unknown reason most of the weather module tests seem to fail (both on Travis as well as local) due to timeout issues. To be able to continue the development I have disabled the weather tests by adding a skip statement: https://github.com/MichMich/MagicMirror/blob/develop/tests/e2e/modules/weather_spec.js#L18

This means the weather tests are currently not used. This could impact the quality of the code. I can't seem to find any reason why these tests time out. 

If anyone is able to solve this issue and get travis to successfully complete the tests, please send a PR. Your contribution would be highly appreciated.

Thanks!

roramirez:
After merge #1849 I could check it

MichMich:
Fixed by reverting electron. See https://github.com/MichMich/MagicMirror/issues/1800

sdetweil:
still failing today

MichMich:
They run ok now. They timeout after 10sec. We might need to stretch that 10 sec.

sdetweil:
The code says if under CI then timeout is 30 seconds

MichMich:
This line (for example) uses 10000ms:
https://github.com/MichMich/MagicMirror/blob/69a887fb05af79071eb20875c1c4a856627f7811/tests/e2e/modules/weather_spec.js#L48

sdetweil:
Ok. Other tests call a function in global_setup to set timeouts. Figured they all would

roramirez:
I've try catch the failed of this and  apparently is because is too slow to run these tests. There a other line
https://github.com/MichMich/MagicMirror/blob/69a887fb05af79071eb20875c1c4a856627f7811/tests/e2e/modules/weather_spec.js#L11
for a wait. Maybe @fewieden can check a look of this.

fewieden:
@roramirez @MichMich I tried this morning the newest version of develop and run the tests locally, they all passed after ~7s.

What is the setup I need to reproduce the issue?

Since this seems to be still relevant, we should reopen the issue.

roramirez:
Hi @fewieden 

You should try into a similar environment where is running like Travis. Maybe you can using the follow guide (I'm use this one) https://jonlabelle.com/snippets/view/markdown/run-travis-ci-builds-locally-with-docker.

Or also requests the debug mode for your Fork in Travis 
https://docs.travis-ci.com/user/running-build-in-debug-mode/#enabling-debug-mode

[...]

PR #1927, opened 2020-02-10
1840 fix weather tests

fewieden:
https://github.com/MichMich/MagicMirror/issues/1840

Fix timing issue in weather tests => returning response by url instead of index

By debugging the tests locally I noticed, that the tests were failing randomly. By taking a closer look I spotted that the failing tests rendered the weather data as JSON in the spectron window. Increasing the timeouts helped me to detect in the network console that the weather data was retunred for the request to the nunjuck template. It seemed that in some random cases there was a timing issue, which requested first the template and then the weather.

The following change is resolving the timing issue, as it is returning the data based on the url, rather than relying on a specific order of requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,24 @@
 dist: trusty
 language: node_js
 node_js:
- - "10"
+  - 10
+  - lts/*
+  - node
 before_install:
   - npm i -g npm
 before_script:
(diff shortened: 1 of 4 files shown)


fewieden:
Also I enabled the tests for always the latest stable and long term support version of nodejs

PR #1937, opened 2020-02-18
use relative path for socket.io

### LINKED ISSUE #1934: Socket.io can't be used with Reverse Proxy in serveronly mode
**Platform**: Raspberry Pi 3 using docker as server (f9d9ff1a354d), Chrome on Windows 10 as client

**Node Version**: 12, the version in the latest(f9d9ff1a354d) docker image

**MagicMirror Version**: v2.10.1

**Description**:  I want to use Magic Mirror behind a Reverse proxy so I can access it via a sub path in the url as `http://localhost/magic-mirror/`. I setup a Preserve Proxy so I can access `http://localhost/magic-mirror/`, but it only shows a blackscreen. The console log of the browser shows that the `socket.io.js` could not be loaded. That's because it's loaded from an [absolute path in index.html](https://github.com/MichMich/MagicMirror/blob/master/index.html#L40) (`src="/socket.io/socket.io.js"`). All other scripts are loaded without problem, because they are using relative paths.

I removed the `/` from the path in the `index.html` but socket.io also uses absolute path somewhere else. So this simple change did not solve the whole problem.

**Steps to Reproduce**: 
1. Start the MagicMirror docker container
2. setup Reverse Proxy using e.g. apache2:
```
ProxyPass "/magic-mirror/" "http://localhost:8080/"
ProxyPassReverse "/magic-mirror/" "http://localhost:8080/"
ProxyPreserveHost On
```
3.  open `http://localhost/magic-mirror/` 

**Expected Results**: open `http://localhost/magic-mirror/` should show magic mirror

**Actual Results**: Blackscreen, chrome console log:
```
GET http://localhost/socket.io/socket.io.js net::ERR_ABORTED 404 (Not Found)
main.js:455 Initializing MagicMirror.
...
```
the url should be `http://localhost/magic-mirror/socket.io/socket.io.js`

**Configuration**: default docker config

**Additional Notes**: I want to use Magic Mirror behind a Reverse Proxy, because then I can better manage all exposed services and protect them with password and TLS. The `localhost` urls are only used for simplification.

Legion2:
This Pull Request fix #1934

* It changes the javascript include path of the `/socket.io/socket.io.js` to a relative one, like all the other scripts
* It appends the path of the current window to the path of the socket.io endpoint

### DIFF
--- a/js/socketclient.js
+++ b/js/socketclient.js
@@ -8,7 +8,9 @@ var MMSocket = function(moduleName) {
 	self.moduleName = moduleName;
 
 	// Private Methods
-	self.socket = io("/" + self.moduleName);
+	self.socket = io("/" + self.moduleName, {
+		path: window.location.pathname + "/socket.io"
+	});
 	var notificationCallback = function() {};
 
 	var onevent = self.socket.onevent;
--- a/js/socketclient.js
+++ b/js/socketclient.js
@@ -9,7 +9,7 @@ var MMSocket = function(moduleName) {
 
 	// Private Methods
 	self.socket = io("/" + self.moduleName, {
-		path: window.location.pathname + "/socket.io"
+		path: window.location.pathname + "socket.io"
 	});
 	var notificationCallback = function() {};
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix regression in currentweather module causing 'undefined' to show up when config.hideTemp is false
 - Fix FEELS translation for Croatian
 - Fixed weather tests [#1840](https://github.com/MichMich/MagicMirror/issues/1840)
+- Fixed Socket.io can't be used with Reverse Proxy in serveronly mode [#1934](https://github.com/MichMich/MagicMirror/issues/1934)
 
 ### Updated
 - Remove documentation from core repository and link to new dedicated docs site: [docs.magicmirror.builders](https://docs.magicmirror.builders).
--- a/index.html
+++ b/index.html
@@ -37,7 +37,7 @@
 		<div class="region bottom right"><div class="container"></div></div>
 	</div>
 	<div class="region fullscreen above"><div class="container"></div></div>
-	<script type="text/javascript" src="/socket.io/socket.io.js"></script>
+	<script type="text/javascript" src="socket.io/socket.io.js"></script>
 	<script type="text/javascript" src="vendor/node_modules/nunjucks/browser/nunjucks.min.js"></script>
 	<script type="text/javascript" src="js/defaults.js"></script>
 	<script type="text/javascript" src="#CONFIG_FILE#"></script>


js/socketclient.js
```diff
@@ -8,7 +8,9 @@ var MMSocket = function(moduleName) {
 	self.moduleName = moduleName;
 
 	// Private Methods
-	self.socket = io("/" + self.moduleName);
+	self.socket = io("/" + self.moduleName, {
+		path: window.location.pathname + "/socket.io"
```
khassel: the slash before `socket.io` causes socket-errors when the path is empty (which means = `/`). Without the slash it works for me with subpath and without.

Legion2: but what if the `pathname` does not contain a `/` at the end

mrfifiak: I don't code in js, but what about[ path.join](https://nodejs.org/api/path.html#path_path_join_paths)?

Legion2: here we are not in a node.js environment, this is executed in the web browser

khassel: I did the following tests:

http://mm.localhost/ with "/socket.io"
````
Request URL:http://mm.localhost//socket.io/?EIO=3&transport=polling&t=N1UpKax
Request Method:GET
Remote Address:[::1]:80
Status Code:
404
Version:HTTP/1.1
````

http://mm.localhost/ with "socket.io"
````
Request URL:http://mm.localhost/socket.io/?EIO=3&transport=polling&t=N1Upjzt
Request Method:GET
Remote Address:[::1]:80
Status Code:
200
Version:HTTP/1.1
````

http://localhost/mm/ with "/socket.io"
````
Request URL:http://localhost/mm//socket.io/?EIO=3&transport=polling&t=N1UrITs
Request Method:GET
Remote Address:127.0.0.1:80
Status Code:
200
Version:HTTP/1.1
````

http://localhost/mm/ with "socket.io"
````
Request URL:http://localhost/mm/socket.io/?EIO=3&transport=polling&t=N1Ur1Cb
Request Method:GET
Remote Address:127.0.0.1:80
Status Code:
200
Version:HTTP/1.1
````

So the first test fails und I think this is the reason for the failed travis-tests.
The mirror fails to load module data, they remain in the state 'loading':

grafik

khassel: > but what if the `pathname` does not contain a `/` at the end

may depends on the reverse-proxy setup. In my case http://mm.localhost/ is redirected to http://mm.localhost, http://localhost/mm/ works and http://localhost/mm is not found. But this could be fixed with a rewrite rule in the proxy.

What about testing the end of the path and adding the `/` before `socket.io` only when necessary?

Legion2: I wanted to keep the changes as simple as possible. Adding additional logic to create the path is one way to solve this.

But it is also possible to add a configuration option, where you can configure the proxy URL, so other modules can use it too. This is required if some modules want to create self links, they must use the proxy URL as base. Because all path that are using localhost with absolute URLs are not accessible when using the Proxy.

Legion2: OK i did a quic test with `http://somedomain.com/magic-mirror` without `/` as proxy url, and this Proxy setup will break the complete application. Because the browser then uses `http://somedomain.com/` as base and all secondary requests fail.

Legion2: So as of this I would consider `pathname` not containing a `/` at the end is the result of a misconfigured proxy and can be ignored here.

Legion2:
@MichMich should I rebase the PR?

khassel:
@MichMich  can we get this merged before the new release next week? Or why are you hesitating?

MichMich:
Merged.

RoadRunnr1469:
I don't understand what you mean by "Remove additional slash from path" below I'm tasing my socket client.js

var MMSocket = function(moduleName) {
        var self = this;

        if (typeof moduleName !== "string") {
                throw new Error("Please set the module name for the MMSocket.");
        }

        self.moduleName = moduleName;

        // Private Methods
        self.socket = io("/" + self.moduleName, {
                path: window.location.pathname + "socket.io"
        });
        var notificationCallback = function() {};

        var onevent = self.socket.onevent;
        self.socket.onevent = function(packet) {
                var args = packet.data || [];
                onevent.call(this, packet); // original call
                packet.data = ["*"].concat(args);
                onevent.call(this, packet); // additional call to catch-all
        };

        // register catch all.
        self.socket.on("*", function(notification, payload) {
                if (notification !== "*") {
                        notificationCallback(notification, payload);
                }
        });

        // Public Methods
        this.setNotificationCallback = function(callback) {
                notificationCallback = callback;
        };

        this.sendNotification = function(notification, payload) {
                if (typeof payload === "undefined") {
                        payload = {};
                }
                self.socket.emit(notification, payload);
        };
};

Legion2:
@MichMich @RoadRunnr1469 I created a PR #1976 which reverts the path to its old default, but it can be changed in the config.

Issue #1756, opened 2019-08-23
Feature Request: Easier Ability to use Environment Variables in Configuration


tillig:
I've been using Magic Mirror 2.8.0 and want to store secrets (e.g., OpenWeather API keys) outside of my configuration file so I can commit the file to source control. This would also allow for easier Docker image creation where you could poke in some environment variables rather than full config files.

Doing this has been a challenge.

- The `config.js` is used in both a server and a client context. That is, instead of the Electron client app retrieving configuration from the `/config` endpoint of the server, the `config.js` is used directly. This means any references to `process.env` don't work because...
- Electron doesn't provide the process environment on the server to the running client. [This issue shows an example of that.] `process` doesn't exist by default, and even if you enable `nodeIntegration`, the process that Electron is running under isn't the same as the process the server started under. Instead, you need to `require('electron')` and access the `electron.remote.process.env` to get those variables marshaled over. But...
- Electron `nodeIntegration` is off by default. It's also turned off in [the main electron.js](https://github.com/MichMich/MagicMirror/blob/a31546b1ff8c7c10e12d7acbab063297d59c6b93/js/electron.js#L28). This can be overridden by config.

[There is a discussion where I raised this over in the forum.](https://forum.magicmirror.builders/topic/11023/is-there-a-way-to-store-secrets-in-the-environment)

A working config.js with environment variables looks like this:

```js
const owApiKeyName = "OPENWEATHER_API_KEY";
var remote = null;
if (typeof window !== "undefined") {
  remote = window.require("electron").remote;
}

var readEnvironment = function (name) {
  if (typeof process !== "undefined" && process.env[name]) {
    // process is undefined in the Electron app.
    return process.env[name];
  }
  if (remote && remote.process.env[name]) {
    // remote is null if the Electron nodeIntegration value isn't set to true.
    return remote.process.env[name];
  }
};


var config = {
  address: "localhost",
  port: 8080,
  ipWhitelist: ["127.0.0.1", "::ffff:127.0.0.1", "::1"],
  language: "en",
  timeFormat: 24,
  units: "imperial",

  electronOptions: {
    webPreferences: {
      nodeIntegration: true
    }
  },

  modules: [
    {
      module: "alert"
    },
    {
      module: "updatenotification",
      position: "top_bar"
    },
    {
      module: "clock",
      position: "top_left"
    },
    {
      module: "calendar",
      header: "US Holidays",
      position: "top_left",
      config: {
        calendars: [{
[...]

PR #1940, opened 2020-02-20
Loading the configuration file in the client from the server instead

buxxi:
I took a first stab at making it possible to use environment variables in the configuration file ( https://github.com/MichMich/MagicMirror/issues/1756 ).

The biggest problem with this seemed to be that both the client and the server both loads the config file each, where the server has access to the environment variables and the client doesn't. But since the server already exposes the possiblity to get the config file like this:  https://github.com/MichMich/MagicMirror/blob/5bf90ae31d600e3f595ffd242b99515fa7905b1a/js/server.js#L56-L58
I figured fetching that as config instead of just including it would cause it to actually have the servers evaluated config instead (instead of this line):
https://github.com/MichMich/MagicMirror/blob/5bf90ae31d600e3f595ffd242b99515fa7905b1a/index.html#L43

I've tested it by setting configuration values to 
```
{
	module: "helloworld",
	position: "lower_third",
	config: {
		"text": process.env.HELLO_WORLD
	}
}
```
And it seems to work and even apply to the client. Maybe this shouldn't be the recommended way to use environment variables in the configuration file, but this pull request would probably be needed whatever is decided.

Hope its ok to use fetch() since it works with ```npm start``` and all modern browsers seems to have supported it for a few years ( https://caniuse.com/#feat=fetch )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   - To start in server only mode, use `npm run server`.
 - Timestamp in log output now also contains the date
 - Turkish translation.
+- Loading the evaluated config file from the server instead of including it in the client (to make it possible to use environment variables in the config file)
 
 ## [2.10.2] - 2020-02-21
 
--- a/index.html
+++ b/index.html
@@ -39,8 +39,6 @@
 	<div class="region fullscreen above"><div class="container"></div></div>
 	<script type="text/javascript" src="/socket.io/socket.io.js"></script>
 	<script type="text/javascript" src="vendor/node_modules/nunjucks/browser/nunjucks.min.js"></script>
-	<script type="text/javascript" src="js/defaults.js"></script>
-	<script type="text/javascript" src="#CONFIG_FILE#"></script>
 	<script type="text/javascript" src="vendor/vendor.js"></script>
 	<script type="text/javascript" src="modules/default/defaultmodules.js"></script>
 	<script type="text/javascript" src="js/logger.js"></script>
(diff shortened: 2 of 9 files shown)


buxxi:
Forgot about there even being tests :)

But I do not know how to make the appropriate changes to fix this.

buxxi:
Having a hard time debugging this, can't get breakpoints or console.log work for anything that isn't in the testfile. Guessing there's something that disables console.log on purpose to not clutter the test output, but I can't find it.

And the tests seems to work if they are run isolated.
```
NODE_ENV=test ./node_modules/mocha/bin/mocha -g "12hr clock" tests/e2e/modules/clock_es_spec.js


  Clock set to spanish language module
    with default 12hr clock config
      ✓ shows date with correct format (41ms)
      ✓ shows time in 12hr format (61ms)


  2 passing (6s)
```
vs this
```
NODE_ENV=test ./node_modules/mocha/bin/mocha tests/e2e/modules/clock_es_spec.js


  Clock set to spanish language module
    with default 24hr clock config
      ✓ shows date with correct format (40ms)
      ✓ shows time in 24hr format (57ms)
    with default 12hr clock config
      1) shows date with correct format
      2) shows time in 12hr format
    with showPeriodUpper config enabled
      ✓ shows 12hr time with upper case AM/PM (58ms)
    with showWeek config enabled
      ✓ shows week with correct format (53ms)


  4 passing (16s)
  2 failing
```
So it seems to be something that isn't reset while stopping/starting the application between tests.

buxxi:
The tests only waited for the all scripts and stylesheets to load before doing the verification.
Changed that to behave like the weather-tests where it waits for a DOM-element to appear and now it works for my changes too. (Ran it a couple of times to be sure).

Had some git issues so I will recreate the PR.

PR #1941, opened 2020-02-21
Add new default config to 'current weather' module

bibaldo:
The idea is to configure the size of the current weather module by passing 'tableClass' config.


MichMich:
Could create a PR for https://github.com/MichMich/MagicMirror-Documentation as well? Thanks!

PR #1948, opened 2020-03-08
Run linter tests on travis (was: Run linter before commit)

rejas:
MM has some linting tasks to keep the codestyle clean. Unfortunately they are not enforced upon git commits so some PR have a wring codestyle.

~This PR adds automatic linting on the pre-commit git hook~
This PR runs the linting tasks on travis builds to report issues found.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,7 @@
 		"curly": "error",
 		"camelcase": ["error", {"properties": "never"}],
         "no-multiple-empty-lines": ["error", { "max": 1, "maxEOF": 1 }],
+		"no-multi-spaces": "error",
 		"no-trailing-spaces": ["error", {"ignoreComments": false }],
 		"no-irregular-whitespace": ["error"]
 	},
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,9 @@ before_script:
   - "sh -e /etc/init.d/xvfb start"
   - sleep 5
 script:
+  - npm run test:lint
   - npm run test:e2e
   - npm run test:unit
-  - grunt
 after_script:
   - npm list
 cache:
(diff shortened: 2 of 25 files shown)


Legion2:
Why not check linting in the travis ci for pull requests. And fail the pipeline if the codestyle is not conform the rules.

MichMich:
I agree with @Legion2 - This also prevents the need for adding a dependency.

rejas:
Ok, will rework my PR accordingly. Although I personally like to have my code linted before people send it over, reduces the round trip time ;-)
 
Question: Since grunt is only used for linting, would it be an option to remove that as a dependency too and call the linting tools directly?

MichMich:
The less dependencies, the better ... :)

Legion2:
@rejas The grunt task config must be changed, to report problems instead of fixing them. Can you give me push access to this branch then I can update the Gruntfile.

rejas:
LIke this @Legion2  ?

Legion2:
yes but now the automatic fixing of linting errors don't work anymore. So we have to create two configs for eslint, one that fixes all errors and the other only reports them without fixing them.
[Gruntfile.js.diff.txt](https://github.com/MichMich/MagicMirror/files/4317425/Gruntfile.js.diff.txt)

rejas:
Done @MichMich 
Replacing grunt would be done on another day, couldnt find a good replacement for the grunt-jsonlint so far.

.eslintrc.json
```diff
@@ -7,6 +7,7 @@
 		"curly": "error",
 		"camelcase": ["error", {"properties": "never"}],
         "no-multiple-empty-lines": ["error", { "max": 1, "maxEOF": 1 }],
+		"no-multi-spaces": "error",
```
Legion2: looks like `jsonlint` does not enforce one indent style

rejas: Something one should look at when getting rid of the grunt dependencies...

PR #1974, opened 2020-04-03
Feature/add profile support to newsfeed

Tom-Hirschberger:
This is a simple implementation of profile support (MMM-ProfileSwitcher) for the official newsfeed module.
Default behavior is like before but now an profiles string can be added to each feed.
Only if the profiles string of the current active news item contains the name of the current profile it will be displayed otherwise the next one is chosen. If the item does not have any profile string it will be displayed, too.

The profile strings will be added to the items during the fetch process.

It is my first real commit to the project. I do not know if there is interest to add such features and if my code quality is good enough. But i give it a try...


khassel:
I don't see any changes made in this PR. @Tom-Hirschberger looking into your MM-repo the branch `local` may contains the things you want to get merged, but not `feature/AddProfileSupportToNewsfeed` ...

Tom-Hirschberger:
Sorry for that. Think i cleaned up the mess and now the right branch should contain my changes.

MichMich:
Thanks for your contribution!

Can you explain what this does? also, if this needs extra configuration work. Please send a PR to the configuration repo as well: https://github.com/MichMich/MagicMirror-Documentation

Tom-Hirschberger:
Hi,

the idea is to make it possible to display different newsfeed items depending to the current profile.
This is my example configuration:
```json5
		{
			module: "newsfeed",
			position: "bottom_bar",	// This can be any of the regions. Best results in center regions.
			config: {
				feeds: [
					{
						title: "Tagesschau",
						url: "http://www.tagesschau.de/xml/rss2",
						profiles: "Mom Dad",
					},
					{
						title: "Heise",
						url: "https://www.heise.de/rss/heise-atom.xml",
						profiles: "Dad"
					},
					{
						title: "Welt",
						url: "https://www.welt.de/feeds/latest.rss",
					}
				],
			},
		},
```
The result will be:
Active Profile "Mom": Items of "Tagesschau" and "Welt" will be displayed
Active Profile "Dad": Items of "Tagesschau", "Welt" and "Heise" will be displayed
Any other Profile is active: Items of "Welt" will be displayed

To do this i did the following changes to the code:
During the creation process of the items of a newsfeed every item that is part of a newsfeed with configured profile string gets added this string.
Items of newsfeed without an profile string configured do not get an string attached.
While the next item to display is chosen only items that match the current profile or do not have attached an profile string will be displayed.
Because items without profile string are accepted in all profiles the default behavior stays the same for users that do not want to use this feature.

I forked the doc repo a few minutes ago and will try to add my changes in a view hours.

MichMich:
Ok, so I now understand what this is doing thanks to https://github.com/MichMich/MagicMirror-Documentation/pull/10#issuecomment-621748598. Somehow I missed the reference to the MMM-ProfileSwitcher in the openingspost. My apologies!

I'm not sure if we should make specific changes to the core modules to support 3rd party modules. There is no such concept as "profile" in MagicMirror. While it's perfectly possible to create the same functionality without the need of adding this change.

Maybe it's better to modify the 3rd party module to work with the current setup?

Tom-Hirschberger:
Hi,
i can understand your argumentation that including options of third party modules to the core modules is tricky. (Especially in case of MMM-ProfileSwitcher because the module works but is unmaintained at the moment)
The profiles work well without modifying the core because normally whole modules will be hidden or shown.
My problem is that i do not want the newsfeed module to get hidden completely but only choose news of different sources.
I do not see a chance to include such a feature by modified any third party module.
Maybe changing the newsfeed module to disable data sources during runtime by sending a specific notification will be an better solution?

MichMich:
Just a thought: why don’t you add multiple instances of the news module in your config (with different sources) and use the one you need based on a profile?

Tom-Hirschberger:
Now i feel stupid.
Did not know that newsfeed supports multi instance.
Tried in my setup and yes this works fine.

MichMich:
🎉😬👍🏻

Tom-Hirschberger:
I close the request now. Thank you for your easy solution.

PR #1976, opened 2020-04-05
Added basename config

### LINKED ISSUE #1973: changes in socketclient.js stops MMM-RemoteControl to work
The changes in `socketclient.js` introduced with this PR https://github.com/MichMich/MagicMirror/pull/1937 in v2.11.0 breaks MMM-RemoteControl, maybe other modules too.

See https://github.com/Jopyth/MMM-Remote-Control/issues/185

After reverting the changes in `socketclient.js` the module works again.

Legion2:
use basename in socket.io path fix #1973

The use of `window.location.pathname` in the socket.io path (introduced in #1937) broke some modules, because the url was not always the basename of MagicMirror. To resolve this the basename is now a configuration option which can be set in the configuration. The default value is `/` which results in the behavior as before #1937 was added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 *This release is scheduled to be released on 2020-07-01.*
 
-# Added
+### Added
 
-# Updated
+### Updated
 
-# Deleted
+### Deleted
 
 ### Fixed
 - The broken modules due to Socket.io change from last release [#1973](https://github.com/MichMich/MagicMirror/issues/1973)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 # Deleted
 
+### Fixed
+- The broken modules due to Socket.io change from last release [#1973](https://github.com/MichMich/MagicMirror/issues/1973)
 
 ## [2.11.0] - 2020-04-01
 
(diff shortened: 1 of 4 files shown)


MichMich:
Thanks for this PR. I like this approach, but I'm not sure `basename` is the best config name. It doesn't explain well what it does. Any other options?

Legion2:
Other name could be `basepath` or `rootpath`.

MichMich:
Let’s go for `basePath`. Could you update the PR?

Legion2:
Yes I can update the PR

khassel:
sorry, but couldn't test earlier ...

This works for the basic MM installation but still not for modules using own html sites. This should be fixed on module site, but I think we should be backward compatible for the standard use case with `basePath="/"`.

Now the used `config` object is `undefined` in the case of MMM-Remote-Control so loading the html page end in a reference error.

So I would prefer a fix like the following in `socketclient.js`:

````
  // Private Methods
  var base = "/";
  if (typeof config !== "undefined") {
    base = config.basePath;
  }

  self.socket = io("/" + self.moduleName, {
    path: base + "socket.io"
  });
````

Legion2:
I think this is a problem of the MMM-Remote-Control it should define the configuration object, or should not load the `socketclient.js`.

khassel:
yes, I know this, that was meant by 

> This should be fixed on module site

but this module (and may others) must be updated and you know not if or when this happens. And meanwhile all users will getting errors using these modules and have again to patch something.

This could be avoided by using `/` if config is undefined.

Legion2:
@khassel Ok, can you create the PR?

khassel:
o.k., next evening ...

PR #1981, opened 2020-04-10
Remove grunt, use non-grunt linters instead

rejas:
Grunt is a dependency that is not really needed since its only used for linting the code. Therefor replace it and use the vanilla linters instead


MichMich:
🎉❤️

PR #1992, opened 2020-04-19
Update font package

rejas:
Updated the roboto font package to the latest version. That provides no longer a ttf version of the file but with woff being supported by 97.5% this seems reasonable (https://caniuse.com/#feat=woff)

Also removed the yarn.lock files in the subdirectories since there isnt one in the main directory neither.


rejas:
any opinion on this @MichMich  ?

PR #1995, opened 2020-04-20
Cleanup default alert module

rejas:
The alert module was using some really old libraries and code. These days the modernizr and classie libraries are not needed anymore, the webkit css prefixes neither.

This PR cleans up the code and uses modern js syntax.

Also fxied some linting issues in general ;-)


hangorazvan:
the alert notification do not hide even with timer :(

PR #1996, opened 2020-04-21
Update eslint config to use eslint:recommended

rejas:
Eslint offers an `eslint:recommended` package that provides sensible defaults. 

Switching to this found some warnings and more issues like duplicate properties. Fixed most of them while disabling three rules for now from the recommended package that triggered more warnings (454 to be exact) than I wanted to fix without getting your opinion on it :-)

So if this gets merged we could look at the other rules and their warnings:

		"no-prototype-builtins"
		"no-undef"
		"no-unused-vars"

and how and when to fix them.


MichMich:
Yeah ... that sounds like a sensible idea. :)

rejas:
Got to check first why the tests fail :-)

sdetweil:
can we move the eslint  dependency out of the devdependencies?  as we need it now in prod. 
i do 
```
npm install --only-prod
```
 to eliminate problems on the armv6l devices.  but then eslint doesn't get installed..

rejas:
why do we need it in prod?

sdetweil:
because it is used by check:config now

rejas:
ah, yes, you are correct. should i do the change here or open a new PR for this @MichMich ?

MichMich:
You can use this PR. Thanks for your hard work!

rejas:
Moved eslint to dependencies and added another useful eslint rule. If the tests are fine and you are ok with it, the PR would be ready for merging...

rejas:
So @MichMich what is your opinion on this PR? Ready to merge? Fix other stuff too or rather in another PR?

MichMich:
Merged! Thanks again!

PR #2007, opened 2020-05-03
Cleanup eslints no-undef warnings

rejas:
I enabled the no-undef rule for eslint and cleaned up teh resulting warnings.

One thing though: 
the file `js/socket.js`seems to be referenced nowhere and I cant seem to find its purpose? Can you elaborate @MichMich what its good for? 

That file still fails the test, I would remove it for good if its not used anymore. Otherwise I will fix the test there too.


rejas:
Thanks for clarifying by deleting the file @MichMich Updating my PR now....

MichMich:
Already did that :)
Waiting for travis to finish.

PR #2017, opened 2020-05-11
Make logger a little more configurable

rejas:
I dont want to have all infos logged on the console once I deploy my MagicMirror. Therefor I added a config entry "logLevel" where one can specify an array of loglevels you wants to see. 

Any comments on the idea so far before I implement more?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ _This release is scheduled to be released on 2020-07-01._
 ### Added
 
 - Compliments Module - Add Advice API (https://api.adviceslip.com/) Option
-
+- Added option to config the level of logging
 - Added prettier for an even cleaner codebase
 
 - Hide Sunrise/Sunset in Weather module
--- a/config/config.js.sample
+++ b/config/config.js.sample
@@ -28,6 +28,7 @@ var config = {
 	httpsCertificate: "", 	// HTTPS Certificate path, only require when useHttps is true
 
 	language: "en",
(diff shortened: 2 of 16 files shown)


MichMich:
Maybe it's an idea to follow the log levels defined in the RFC 5424 specification: `emergency`, `alert`, `critical`, `error`, `warning`, `notice`, `info`, and `debug`.

rejas:
Will take a look at those...

rejas:
Mhhh... not sure if we can those log levels from the RFC are all needed here:

              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages

Could be a little overengineered if we use all of them. Plus we would have to deprecate those that are not in there (like console.log)...

rejas:
So in my mind its ready for review. As I wrote, adding the RFC log levels seems to be a little too complex for this scenario. Plus one would have to add some warnings for deprecated methods and such. 
Any more comments @MichMich ?

MichMich:
Hm, I agree. In that case, let's use the defaults for `console`:

- `log`
- `info`
- `warn`
- `error`

sdetweil:
just saw this pr, how does this compare to the MMM-Logging module?  which merges all reporting into a single log,  server and client side?

PR #2020, opened 2020-05-13
Add Random Advice API Option

ndom91:
Added https://api.adviceslip.com/ Option to the Compliments Module.

To use: pass `advice: true` as a config parameter and leave the remoteFile parameter empty. 

Will return a random piece of advice from their API.

### DIFF
--- a/modules/default/compliments/compliments.js
+++ b/modules/default/compliments/compliments.js
@@ -1,7 +1,9 @@
+/* global Log, Module, moment */
+
 /* Magic Mirror
  * Module: Compliments
  *
- * By Michael Teeuw https://michaelteeuw.nl
+ * By Michael Teeuw http://michaelteeuw.nl
  * MIT Licensed.
  */
 Module.register("compliments", {
@@ -39,7 +41,8 @@ Module.register("compliments", {
 		afternoonStartTime: 12,
 		afternoonEndTime: 17,
 		random: true,
-		mockDate: null
+		mockDate: null,
+    advice: false
 	},
 	lastIndexUsed:-1,
 	// Set currentweather from module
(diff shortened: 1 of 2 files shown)


modules/default/compliments/compliments.js
```diff
@@ -1,7 +1,9 @@
+/* global Log, Module, moment */
+
 /* Magic Mirror
  * Module: Compliments
  *
- * By Michael Teeuw https://michaelteeuw.nl
+ * By Michael Teeuw http://michaelteeuw.nl
```
rejas, line 6: why did you undo this?

modules/default/compliments/compliments.js
```diff
@@ -62,7 +65,19 @@ Module.register("compliments", {
 				self.config.compliments = JSON.parse(response);
 				self.updateDom();
 			});
-		}
+		} else if (this.config.advice) {
+      var xobj = new XMLHttpRequest(),
```
rejas, line 69: this needs to be a ; at the end not a ,
thats probably why the tests fail

modules/default/compliments/compliments.js
```diff
@@ -125,7 +140,7 @@ Module.register("compliments", {
 
 		compliments.push.apply(compliments, this.config.compliments.anytime);
 
-		for (var entry in this.config.compliments) {
+		for (entry in this.config.compliments) {
```
rejas, line 143: please do not undo this

modules/default/compliments/compliments.js
```diff
@@ -186,7 +201,7 @@ Module.register("compliments", {
 		// create a span to hold it all
 		var compliment = document.createElement("span");
 		// process all the parts of the compliment text
-		for (var part of parts){
```
rejas, line 189: please do not undo this

modules/default/compliments/compliments.js
```diff
@@ -1,7 +1,9 @@
+/* global Log, Module, moment */
```
rejas, line 1: This isnt neccessary and was already removed

modules/default/compliments/compliments.js
```diff
@@ -39,7 +41,8 @@ Module.register("compliments", {
 		afternoonStartTime: 12,
 		afternoonEndTime: 17,
 		random: true,
-		mockDate: null
+		mockDate: null,
+    advice: false
```
rejas, line 45: please use the same intendation here as in the lines above. makes the code more readable.

modules/default/compliments/compliments.js
rejas: also: please use more intendation here

ndom91:
I'm going to close this one and open another one from a correct branch ;)

PR #2021, opened 2020-05-14
Update Compliments Module - Add Advice API

ndom91:
Add Advice API (https://api.adviceslip.com/) Option instead of hard coded compliments.

PR #2026, opened 2020-05-24
WIP: Compliments - Advice API Update

ndom91:
Did some more testing locally and this works now!

Just add: 

```
config: {
  advice: true
}
```

to the compliments section of the config and the following will happen:

It makes a request to the API, previously it was just returning 1 random piece of advice, which is not ideal if you're trying to loop through them over time.

So this update changes things so I make a simple query, enough to return many results (in this case simply for 'a'). And then filter for pieces of advice less than 30 characters. This is then put in the `anytime` object key and written over the compliments object resulting in the advice array being displayed.

I will make a PR to the docs repo shortly @MichMich


rejas:
Why overwrite the config? Shouldnt i bei able to add more than one source of compliments to the Array?

ndom91:
What do you mean overwrite the config? No config gets overwritten. Just the object which contains the original compliments. If yuo set `advice` back to false and restart you'll be back to normal.

modules/default/compliments/compliments.js
```diff
@@ -63,20 +63,28 @@ Module.register("compliments", {
...
-				if (xobj.readyState === 4 && xobj.status === 200) {
-					const adviceResp = JSON.parse(xobj.responseText);
-					self.config.compliments = adviceResp.slip.advice
-					self.updateDom();
-				}
-			};
-			xobj.send(null);
 		}
 
+    if (this.config.advice) {
```
rejas, line 68: your intendation is wrong again

modules/default/compliments/compliments.js
```diff
@@ -63,20 +63,28 @@ Module.register("compliments", {
...
+        xobj.onreadystatechange = function() {
+        if (xobj.readyState === 4 && xobj.status === 200) {
+            const adviceResp = JSON.parse(xobj.responseText);
+            const anytime = []
+            adviceResp.slips.forEach(item => {
+                if (item.advice.length < 30) {
+                  anytime.push(item.advice)
+                }
+            })
+            self.config.compliments = { anytime: anytime }
```
rejas, line 81: here you overwrite ALL other compliments the user sets in his config.  you should only append the received advices to the array of anytime compliments

rejas: also missing a ; at the end

modules/default/compliments/compliments.js
```diff
@@ -63,20 +63,28 @@ Module.register("compliments", {
...
+        var xobj = new XMLHttpRequest();
+        xobj.overrideMimeType("application/json");
+        xobj.open("GET", "https://api.adviceslip.com/advice/search/a", true);
+        xobj.onreadystatechange = function() {
+        if (xobj.readyState === 4 && xobj.status === 200) {
+            const adviceResp = JSON.parse(xobj.responseText);
+            const anytime = []
+            adviceResp.slips.forEach(item => {
+                if (item.advice.length < 30) {
+                  anytime.push(item.advice)
```
rejas, line 78: missing a ; at the end

modules/default/compliments/compliments.js
```diff
@@ -63,20 +63,28 @@ Module.register("compliments", {
...
+    if (this.config.advice) {
+        var xobj = new XMLHttpRequest();
+        xobj.overrideMimeType("application/json");
+        xobj.open("GET", "https://api.adviceslip.com/advice/search/a", true);
+        xobj.onreadystatechange = function() {
+        if (xobj.readyState === 4 && xobj.status === 200) {
+            const adviceResp = JSON.parse(xobj.responseText);
+            const anytime = []
+            adviceResp.slips.forEach(item => {
+                if (item.advice.length < 30) {
```
rejas, line 77: maybe make the number of advices configurable?

modules/default/compliments/compliments.js
```diff
@@ -63,20 +63,28 @@ Module.register("compliments", {
...
-					self.updateDom();
-				}
-			};
-			xobj.send(null);
 		}
 
+    if (this.config.advice) {
+        var xobj = new XMLHttpRequest();
+        xobj.overrideMimeType("application/json");
+        xobj.open("GET", "https://api.adviceslip.com/advice/search/a", true);
```
rejas, line 71: maybe make the searched letter a randon one?

rejas:
> What do you mean overwrite the config? No config gets overwritten. Just the object which contains the original compliments themselves. If yuo set `advice` back to false and restart you'll be back to normal.

yeah, thats not good that the advices overwrite all other compliments. They should all be displayed and not depend on advice being set to false. For example I might want to show an advice anytime but in the evening also some other compliments. With your code that is not possible at the moment anymore

ndom91:
Well advice defaults to false, I pictured it as being like an optional different mode for the compliment module, either you can run it normally in "compliment mode" - no action required. Or if you want to get these advice lines, you set `advice: true` and you're in "advice mode". 

I can just push the pieces of advice onto the anytime array though if you guys would rather have it that way. No problem. 

And yeah I have some experience writing tests, let me take a look when I get back home to a computer.

hangorazvan:
> yeah, thats not good that the advices overwrite all other compliments. They should all be displayed and not depend on advice being set to false. For example I might want to show an advice anytime but in the evening also some other compliments. With your code that is not possible at the moment anymore

I agree with this observation, it was more useful for the advice to appear randomly with compliments

modules/default/compliments/compliments.js
ndom91: This is not the number of advices, rather the length of the string so that we dont get pieces of advice running over 5 lines in huge font in the middle of the screen, for example.

modules/default/compliments/compliments.js
ndom91: Yeah we could, searching for 'a' seems to return the most possible results after my admittedly short testing. With this 'a' search we get 192 results, which i then filter for results less than 35 chars.

ndom91:
Okay so I initially wrote it in vim and it was using spaces instead of tabs. VS Code picked up on this projects tab settings and fixed all that. Also I adjusted the core code to append the pieces of advice to the anytime array. That all seems to work well.

Then I began on the tests. I got it mostly setup afaik, including the mocking results. However, I couldn't get the actual spec test to work. If someone could take a look, I'm not super familiar with mocha so I'm probably just missing something simple. Happening at the bottom of `/tests/e2e/modules/compliments_spec.js`.

Thanks!

modules/default/compliments/compliments.js
rejas: Ah, I see. Adding a little comment to that line with that info would be useful for later generations I guess :-)

modules/default/compliments/compliments.js
```diff
@@ -42,43 +42,57 @@ Module.register("compliments", {
...
-			this.complimentFile(function(response) {
+			this.complimentFile(function (response) {
 				self.config.compliments = JSON.parse(response);
 				self.updateDom();
 			});
-		} else if (this.config.advice) {
+		}
+
+		if (this.config.advice) {
 			var xobj = new XMLHttpRequest();
```
rejas, line 69: that could be a const too, right?

modules/default/compliments/compliments.js
rejas: Looking at the advice api I see they also offer a random adivce call. How about not loading a set of advices here but rather call https://api.adviceslip.com/#endpoint-random in the randomCompliment function?

modules/default/compliments/compliments.js
ndom91: Yes it absolutely could. Just updated ;)

hangorazvan:
Same error as in #2021 

Then:
`Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'currentweather' in Some of life's best lessons are learnt at the worst times.`

Now:
`Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'currentweather' in Remember that spiders are more afraid of you, than you are of them.`

ndom91:
> Same error as in #2021
> 
> Then:
> `Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'currentweather' in Some of life's best lessons are learnt at the worst times.`
> 
> Now:
> `Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'currentweather' in Remember that spiders are more afraid of you, than you are of them.`

For context, the trace included:
```
if (this.currentWeatherType in this.config.compliments) {
	compliments.push.apply(compliments, this.config.compliments[this.currentWeatherType]);
}
```

Im not familiar with this `if (... in ...) ` syntax, but doesnt sound like you have the newest version either because the pieces of advice themselves are no longer directly in the compliments object, but in the `anytime` key of the compliments object. This should simply return false and continue on as far as I can tell..

rejas:
> but doesnt sound like you have the newest version either

The thing is: your PR here @ndom91 isnt yet merged in to develop. But develop technically is the "newest" version for developers. So in order to get it merged (and those errors fixed) this PR needs to get in good shape. Next step would be to fix those merge-conflicts, then tackle the test suite.

ndom91:
> > but doesnt sound like you have the newest version either
> 
> The thing is: your PR here @ndom91 isnt yet merged in to develop. But develop technically is the "newest" version for developers. So in order to get it merged (and those errors fixed) this PR needs to get in good shape. Next step would be to fix those merge-conflicts, then tackle the test suite.

Yeah I just meant it doesnt sound like he had the newest version of this branch, because he was specifically mentioning behavior regarding this advice stuff.

Anyway, I cleared up the conflicts now.

tests/e2e/modules/compliments_spec.js
```diff
@@ -123,3 +126,39 @@ describe("Compliments module", function () {
...
+
+	describe("Config - advice:true", function () {
+		before(function () {
+			process.env.MM_CONFIG_FILE = "tests/configs/modules/compliments/compliments_advice.js";
+		});
+
+		it("one of less then 35 char options", async function () {
+			// No template available for compliments to pass in here ?
+			await setup(generateAdvice())
+			return app.client.waitUntilTextExists(".compliments .module-content div", oneOf([
```
rejas, line 158: travis doesnt recognize this oneOf... https://travis-ci.com/github/MichMich/MagicMirror/jobs/340236682#L969

MichMich:
Hi @ndom91, what's the status of this PR? Travis is failing, so I can't merge it.

ndom91:
@MichMich Okay so I fixed that specific issue, but I'm still having trouble getting it to accept my mock data. I am basically doing the same thing that is being done with the weather module, to mock data normally gathered from a REST call, and for some reason it doesn't seem to get in there..

If anyone has some time, can they please take a look at: https://github.com/ndom91/MagicMirror/blob/964b92116e170905ba0f4e655b01ea1225fc4d31/tests/e2e/modules/compliments_spec.js#L155-L156

MichMich:
Not sure what's going on. It also gives some type lint errors. Does this PR fix any current issues with the new advice API? Otherwise I have to revert the Advice implementation altogether since the next master release is planned for 1st of July.

ndom91:
@MichMich no unfortunately I couldn't get the testing working. This was just a little fun addition anyway. No worries if we just revert this all together. Ya win some, ya lose some

rejas:
I'll try to take a look at the tests this afternoon. I like the idea and would be sad to see it removed :-)

MichMich:
I've reverted the initial implementation of the Advice implementation. Feel free to send a new PR with the full implementation including tests. Just to be sure I won't be merging it before the upcoming release.

hangorazvan:
bad option anyway

PR #2031, opened 2020-05-28
[calendar] Use wrapEvents to also truncate the location

radokristof:
Use the provided function to also truncate the event location (if enabled).

Without this, currently really long location (which can happen really often) titles are displayed in full length and messes up the whole module.

I think it makes sense to also truncate the location with the calendarTitle as well, if you don't like it this way, maybe I could introduce a new config variable to control this also individually.


modules/default/calendar/calendar.js
```diff
@@ -396,7 +396,7 @@ Module.register("calendar", {
 					var descCell = document.createElement("td");
 					descCell.className = "location";
 					descCell.colSpan = "2";
-					descCell.innerHTML = event.location;
+					descCell.innerHTML = this.titleTransform(event.location);
```
rejas, line 399: maybe rename the function to show that it not only transform the title but now it transforms other stuff too?

radokristof: Yes I also thought about that. I wanted to create this PR and ask the editors and maintainers here, what they might prefer better (just rename the function to better represent what it does or do this completely independent from the title and have a seperate variable to control the truncating of the location title).

radokristof: I couldn't come up with a better name for this function (since the event location title is also a title), but I think truncate better represent what this function does right now.

modules/default/calendar/calendar.js
```diff
@@ -710,16 +710,16 @@ Module.register("calendar", {
...
 	 * Shortens title based on config.maxTitleLength and config.wrapEvents
 	 *
 	 * argument title string - The title to transform.
 	 *
 	 * return string - The transformed title.
 	 */
-	titleTransform: function (title) {
+	truncateTitle: function (title) {
 		for (var needle in this.config.titleReplace) {
 			var replacement = this.config.titleReplace[needle];
```
rejas, line 725: Oh, this does more than just shorten the title, it also replaces stuff...  Maybe you only need line 735 ?

radokristof: Yes, this really transforms the title. The real question is the same here: should we do this transformation to the event title as well?
Anyway, I will revert these changes and introduce a new function which only really just strips the string.

radokristof:
@rejas How do you like it this way?
This might be a neat solution for this.
However maybe all function could be added to the event location title also (like the event title), replacing based on regex, but I don't know if anyone would need this...

rejas:
a) it might be helpful for event titles too indeed
b) maybe you dont need locationTitleShorten at all and jsut call the shorten function directly? 
c) would it make sense to have a special config value for how maxLocationTitleLength? Or does it only make sense to be that always the same value as maxTitleLength ?

radokristof:
a) Yes indeed. I just created that one because I thought that I could add other transformations there (like the transformTitle has). However if everything should be supported on event location title, maybe the transformTitle should be generalized... (maybe by passing the other config values as well as a parameter rather than having hardcoded in the function).
b) See above.
c) Well for me, with the same value works really great, but I think it would be great to have a different value for this as well.

radokristof:
@rejas How do you like it this way?
With this, there is no additional function introduced, everything can be customized (individually for event title and event location title).

Still need to update the docs there, and also update the module docs to refer this new optional variables.

PR #2032, opened 2020-05-28
[weather] Ability to hide sun details

radokristof:
The new clock module allows to show sun and moon details more detailed than the weather module does.

Right now, you can't hide just hide the sun details. With this additional config variable, you can hide just these.
So if you use the clock and weather module together, you don't have to display these information duplicated.


rejas:
Looks good. Would you be able to write a test for this feature too?

radokristof:
If you are happy with this, I will sure write a test for this.
Also I will include a changelog and should I also create a PR in the docs repo updating the module readme with this setting?

radokristof:
I had a look at the tests, but I don't know what should I test.
Usually as I see, most of the features are turned off by default and tested in the "Options" section.

So I would like to follow this pattern I would need to test if it is showing up (I don't know if it would make sense to test if it is not showing up?!) but I think this should be "true" by default so it the apperance of the module will not be different for users after updating to the next version which contains this feature.
And in the default section, the check for sunrise and sunset is already there...

rejas:
So shouldnt your test set the feature to false and see that nothing like that shows up?

radokristof:
Yes, that's what I wanted to do, but all of the tests seems to do it other way around and only check for existing elements. But I will write the test like this.

radokristof:
Well I did not find anything on how could I do this the simplest way. Also I could not find a test in any of the modules tests which checks for an element which is not present (all tests are checking that an element is there), so I would be happy if you could help me out on this.

sdetweil:
@radokristof 

well, if u FIND it then the test failed..  its just what you report from the test result..

radokristof:
@sdetweil Yes I understand that.
However @rejas asked me to make a test for the situation when this is set to false (showSun: false), so the test should check for the sunrise/sunset element that it is really not visible.

sdetweil:
is it not present or display:none?

radokristof:
It is not present, if I'm right, because if this config option is not set to true, that span element will not even get added to the DOM.
I have tried using getText() and adding the right selector for that element, but that would also fail, because it is completely not there in the current DOM.

sdetweil:
i think u want to use the webdriver.client apis  it has 
https://webdriver.io/docs/api/element/waitForExist.html
```
$(selector).waitForExist({ timeout, reverse, timeoutMsg, interval })
```
note that it says 
```
 If the reverse flag is true, the command will instead return true if the selector does not match any elements.
```

MichMich:
Merged. Feel free to send a PR with tests as well. That would be awesome.
Also a PR for the documentation repo would be welcome as well:

https://github.com/MichMich/MagicMirror-Documentation

radokristof:
Thanks! Sure, I will try to do both of these PR soon!

PR #2058, opened 2020-06-28
UK Met Office DataHub weather provider

XBCreepinJesus:
I have created a new weather provider that uses the Met Office's new DataHub API (their existing DataPoint service will eventually be shut down, to be replaced by this one).

I have written this provider to return precipitation as a probability, in line with the existing Met Office provider (although it can also supply absolute values in mm). As such, there needs to be a small addition to `weather.js` where the `if` statement checks for the existing Met Office module when returning precipitation units.

The documentation for the weather module will also need to be updated (if approved, of course).

It solves a *created* issue - editing weather.js to check for this provider when returning precipitation units
N/A
Adds a provider for the Met Office's new API.
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _This release is scheduled to be released on 2020-07-01._
 - Added option to config the level of logging
 - Added prettier for an even cleaner codebase
 - Hide Sunrise/Sunset in Weather module
+- Added Met Office DataHub (UK) provider
 
 ### Updated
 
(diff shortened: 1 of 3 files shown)


rejas:
Looks nice on first sight. 
a) Could you run `npm run lint:prettier` and commit the changes? That would fix the tests
b) Do you know when the old API is shut down?

XBCreepinJesus:
Ok, I've run the prettier script - it just removed a couple of trailing commas by the looks of things. I also replaced the use of `config.long` with `config.lon` to be more consistent with existing providers - something I just noticed while looking at the other providers documentation!

As for when the old API is shut down - there's nothing to say at the moment, but it's probably not for a while yet. Some info here: https://www.metoffice.gov.uk/services/data/datapoint/notifications/weather-datahub

PR #2060, opened 2020-06-30
fixed minor typo (apiKey vs. appid)

bryanzzhu:
Not much of note. I was initially getting nothing from the module and got confused by the typo since I thought the config file needed an `appid` key instead of an `apiKey` key, which is used everywhere else in the module.


MichMich:
This is a weird issue indeed. Not sure why some use `appid` and some use `apiKey`. I will look into this after releasing today's new master.

PR #2066, opened 2020-07-02
 Adjusts maxNumberOfDays depending on the endpoint 

### LINKED ISSUE #2018: Weather forecast ignores "maxNumberOfDays"
Hi there,

currently the config parameter "maxNumberOfDays" of the module "weatherforecast" seems to be ignored - at least it does not work increasing it to more than 7 days. That is because the API needs an additional parameter "cnt", which indicates how many days are to be retrieved. Adding it to the weatherforecast.js in line 305 solves the issue for me:

`params += "&cnt=" + this.config.maxNumberOfDays;`

Best regards,
Mirko

oemel09:
Fixes #2018 

Now the `cnt` is set depending on the endpoint
- for 'forecast/daily' the `maxNumberOfDays` is used directly
- for 'forcast' `maxNumberOfDay` is adjusted. The maximum here is 5 days

The documentation for this module need to be adjusted as well:
https://docs.magicmirror.builders/modules/weatherforecast.html#configuration-options

If a free API token is used this description fits the maxNumberOfDays config:
Possible values: 1 - 5
Default value: 5 (5 days) 

Else the config description remains the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ _This release is scheduled to be released on 2020-10-01._
 ### Deleted
 
 ### Fixed
+- Fix the use of "maxNumberOfDays" in the module "weatherforecast depending on the endpoint (forecast/daily or forecast)". [#2018](https://github.com/MichMich/MagicMirror/issues/2018)
 
 ## [2.12.0] - 2020-07-01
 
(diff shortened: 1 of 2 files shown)


oemel09:
Which to use, floor or ceil, may even depend on how the API works exactly...
Not sure if that check is needed anyway, but what I'm trying to do is to only get a full day from the API.
Because once you get the forecast till noon, it could still be cooler in the evening on that day, so the forecast would display some wrong minimal temperature for that day, because it doesn't have the data for the evening yet.

PR #2078, opened 2020-07-08
weatherforecast OpenWeather onecall api support

cjbrunner:
Relating to [Issue 2076](https://github.com/MichMich/MagicMirror/issues/2076), adds support to weatherforecast for OpenWeather onecall API.

PR #2081, opened 2020-07-09
adds current, hourly, and daily forecasts to the Weather module (OpenWeatherMap One Call API)

bryanzzhu:
**Adds current, hourly (max 48) and daily (max 7) weather forecasts to the default Weather module via the OpenWeatherMap One Call API.**

[Sample Module Configuration Code](https://gist.github.com/bryanzzhu/d77817b1dfbfc7fb9a5d29feac73e875)

[Image of Sample MagicMirror Output Using Above Configuration](https://i.imgur.com/K2F04pr.png)

I've been using these modules for about a week now with no issues, so I think the features they implement should be safe to integrate into the main branch, albeit with some changes.

Unlike other OWM APIs, the One Call API takes in a latitude and longitude to determine weather. In the code these are called `lat` and `lon` with default values of `0` for both. Since they don't return any sort of city-country name, `fetchedLocation` is currently set to the latitude longitude coordinates returned by the API instead. On the upside, the new API returns a `timezone_offset` based on the given latitude and longitude, meaning that we don't need to deal with any sort of timezone wizardry on our end, OWM takes care of all of that for us and we just need to factor in the offset ourselves.

Currently with the way it's implemented, `weatherprovider.js` has a new set of default methods under the `weatherData` umbrella. The `openweathermap.js` provider implements the `fetchWeatherData()` method in particular by calling another new function `fetchOnecall()`, which is used to fetch weather information for all three of the new current, hourly, and daily One Call requests. Since the One Call API lets you exclude information, we specify which information we want as part of the call to save some data (these were the additions to `getParams()`). All this also means that existing customization options for things like fade, showing precipitation, degree labels, etc. should all still work with the Weather module when using the One Call API.

Most of the above was done because it was the "easy way out" where I wouldn't have to combine/modify code and potentially break existing functionality. As it stands, before this gets merged decisions should be made by someone more qualified about the following:

- ideally the stuff the new methods implement for the OpenWeatherMap provider would replace the existing methods, at least to some degree
    - the existing implementation of `forecast` using the `forecast/daily` endpoint shouldn't get replaced by the One Call version since paying users are able to get more days out of their API call then One Call users can
   - the `forecast` type I feel should probably be changed into the `daily` type to better reflect the info it provides, with `forecast` probably left in for backward compatibility until some point in time
- it may be better to set `fetchedLocation` to the `location` config option (if provided by the user), so that the user can display whatever they want for the location name, instead of the current coordinate default
- if these are being left as separate calls, the names of the types being used right now for the One Call API (`wDataCurrent`, `wDataHourly`, and `wDataDaily`) should definitely be changed to something less stupid


khassel:
have you seen https://github.com/MichMich/MagicMirror/pull/2078?

bryanzzhu:
> 
> 
> have you seen #2078?

Just did, though it looks like that request adds the onecall endpoint to the weatherforecast module. This one is for the weather module and all three possible forecast types.

MichMich:
Please run prettier ... it contains code style issues. Thanks!

bryanzzhu:
small update, been getting rid of the functions I created and moving their code inside existing ones, will probably add another type `daily` that does the same thing as type `forecast`

will run prettier after that

modules/default/weather/providers/openweathermap.js
```diff
@@ -95,6 +116,18 @@ WeatherProvider.register("openweathermap", {
...
 	},
 
+	/*
+	 * Generate WeatherObjects based on One Call forecast information
+	 */
+	generateWeatherObjectsFromOnecall(data) {
+		if (this.config.weatherEndpoint === "/onecall") {
+			return this.fetchOnecall(data);
+		}
+		// if weatherEndpoint does not match onecall, what should be returned?
```
rejas, line 126: is that a real question or a rhetorical? is there a TODO still here?

modules/default/weather/weather.js
```diff
@@ -11,8 +11,10 @@ Module.register("weather", {
 	defaults: {
 		weatherProvider: "openweathermap",
 		roundTemp: false,
-		type: "current", //current, forecast
+		type: "current",
```
rejas, line 14: could you add all new types to the comments you deleted? //so people can see what the possible values are?

modules/default/weather/weather.js
```diff
@@ -124,6 +127,9 @@ Module.register("weather", {
 
 	// Select the template depending on the display type.
 	getTemplate: function () {
+		if (this.config.type === "daily") {
```
rejas, line 130: maybe refactor it into a switch statement? and also add toLowerCase just in case.

modules/default/weather/providers/openweathermap.js
bryanzzhu: Rhetorical, the `generateWeatherObjectsFromForecast()` method I based the new one off of had a similar comment, I guess because no one settled on what type of "empty" object the method should return.

modules/default/weather/weather.js
bryanzzhu: Whoops, I thought that was me, yeah I'll put that back. I've been adding them to the documentation as well.

Issue #2045, opened 2020-06-13
Contribution available to revise default/weather/providers/weather.gov per API guidelines


easyas314:
The weather.gov api needs to call several URLs to collect weather information. I have revised the code for this and I'd like to contribute it. Interested? And I'll ask for proper submission guidance in advance. Should I sync my fork/branch specifically with the develop branch? I have similar and more questions about a documentation update. Thanks.

MichMich:
Sorry for the slow response. Feel free to send a PR to discuss the implications.

MichMich:
PR Merged.

PR #2084, opened 2020-07-11
Wx gov

easyas314:
Ref #2045 

No visual changes; just tried to stay consistent with the new weather objects' model.

Quick weather.gov API description:
1. weather.gov API starts with using the lat,lon for a /points/ URL. 
2. response of step 1 provides additional URLs based on a "gridpoint" determined by the API.
3. these subsequent URLs that have correct "gridpoint" paths for the /forecast and /forecast/hourly data.
4. the most complete 'current' weather data is obtained by following a 'stations' URL from step 2, and then following another URL from that result to the final reporting station information.

I'm a complete amateur for this type of coding, so set your expectations low. Needing all these dependent calls to get to the data makes me think there's a lot of room for errors.

On a general weather provider configuration level -- I think the definition of a weather provider completely drives both the apiBase as well as the weatherEndpoint, so these options seem superfluous for a user to further configure.

Thanks for considering this change.
 -brad


package.json
```diff
@@ -4,7 +4,7 @@
 	"description": "The open source modular smart mirror platform.",
 	"main": "js/electron.js",
 	"scripts": {
-		"start": "DISPLAY=\"${DISPLAY:=:0}\" ./node_modules/.bin/electron js/electron.js",
+		"start": "./node_modules/.bin/electron js/electron.js",
```
khassel, line 7: the `DISPLAY` var is needed (except on windows) so removing this is no good idea.

easyas314: Thank you. My poor git skills. I need to remove both the package.json and the package-lock.json files from this PR.

Issue #2118, opened 2020-08-26
Tomorrow's calendar events are being displayed as 'Today'. 


DivingIvan:
Hi,

I have found an issue with the MagicMirror calendar module, whereby it is displaying single full day events with tomorrow's date as being 'Today'.

So, for example, if the event is a single full day event (eg a birthday) with date '20200827', and the current date is 26th August 2020, it will show up in on the MagicMirror calendar as being 'Today' whereas it should show as 'Tomorrow'.

**Note - after further investigation, it appears that this problem only occurs with repeating events; so, for example a birthday or anniversary that repeats each year. The problem does not appear to happen with single (one off) events.**

Events for later dates do show up correctly; for example 'In 2 days' for an event with a date '20200828'.

I am fairly certain that this is due to the way that events are added to the event list; and the way that the 'event.today' boolean property is calculated; ie., the lines of code:-

`var today = moment().startOf("day");`

and

`event.today = event.startDate >= today && event.startDate < today + 24 * 60 * 60 * 1000;`

I suspect that the 'moment().startOf("day")' is not taking account of the fact that UK time is currently set to British Summer Time, as opposed to GMT, so is returning an absolute time value that is offset by +1 hour. As a result, the code:-

`event.startDate < today + 24 * 60 * 60 * 1000`

returns a true value instead of false. Indeed, I have observed that if I manually set the system time to be 'GMT', the calendar event will show up correctly on the calendar as 'Tomorrow'.

I am running on a Raspberry Pi 3 B with the latest version of Pi OS (Rasbian Buster). Kernel version 5.4.51.

Node version is 6.14.6.

MagicMirror version is 2.12.0

magic21nrw:
Facing the same issue here. Events for tomorrow are displayed as today.

MM Version 2.12.
Node Version v10.21.0

Tried to replace ical ^0.8.0 with 0.5.0 without success. Seems that the other bug reports are referring to other issues and that yours is the same I am facing here.

rejas:
This issue I think is already fixed in the develop branch. Could you guys test if it goes away when using that branch?

DivingIvan:
Hi rejas.

Many thanks for the information. I have just checked out the develop branch, and can confirm that this does appear to have at least partially fixed the issue. I've not done a comprehensive test, but the one recurring event that I have in my calendar for tomorrow does indeed show as being 'Tomorrow'.

[...]

Issue #2111, opened 2020-08-07
On day of event, Calendar doesn't show recurring events starting within TZ difference from midnight (static calendar)


castletonroad:
**Platform**: Raspberry Pi 4, 2 GB, Raspbian 10 (buster); Electron 6.1.12

**Node Version**: 10.22.0

**MagicMirror Version**: 2.13.0-develop

**Description**: On the day of an event, Calendar module does not show any recurring events where the start time is BEFORE 10am - these events start within the TZ difference (10hrs) from midnight. (My TZ is AEST, currently GMT+10.00). If the event is NOT recurring, it will show on the MM. If the event IS recurring, but starts after 10.00, it will also show on the MM

**Steps to Reproduce**:
1. Create a new recurring (two-weekly, 4 occurrences, 09.20-10.15) event (via either of iPhone Calendar app, or Nextcloud web-based Calendar app) for next day - recurring event shows correctly on the MM ('Tomorrow at 09.20').
2. On the day of the event, the recurring event starting before 10.00 does not show; any recurring event starting after 10.00 shows correctly.
3. Create a NON-recurring event for same day (09.20-10.15) and restart the MM - the event shows correctly on the MM.
4. Make this event RECURRING and restart the MM - the event DOES NOT show.
5. Change this event's start time to 10.30 (vs. 09.20) - this recurring event shows correctly on the MM!
6. Change start time to 10.05 - recurring event shows correctly on the MM.
7. Change start time to 10.00 - recurring event shows correctly on the MM.
8. Change start time to 09.55 - recurring event DOES NOT show on the MM.
9. Remove the recurrence from the event - recurring event shows correctly on the MM.

**Expected Results**:
Single instance events starting before 10.00 show correctly on the MM
Recurring events starting before 10.00 show correctly on the MM
Recurring events starting on/after 10.00 show correctly on the MM

**Actual Results**:
Single instance events starting before 10.00 show correctly on the MM
**Recurring events starting before 10.00 DO NOT show on the MM**
Recurring events starting on/after 10.00 show correctly on the MM

**Configuration**: What does the used config.js file look like? Don't forget to remove any sensitive information!
NextCloud: 18.0.06, Calendar App 2.0.3

Export of .ICS file to MM/modules/default/calendar folder (so using a static calendar)

~/MagicMirror/config/config.js:

                {
                        module: "calendar",
                        header: "School Timetable",
                        position: "top_right",
                        config: {
                                calendars: [
                                        {
[...]

Issue #2110, opened 2020-08-06
Calendar not showing reoccurring full-day events on the day of the event (static calendar file)


Feedy88:
_Hey there, first of all, thanks for this amazing project. I started setting up a few weeks ago and enjoy every time working on it._

**Platform**: Raspberry Pi 4, 2 GB, Raspbian 10 (buster); Electron 6.1.7

**Node Version**: 10.21.0

**MagicMirror Version**: 2.13

**Description**: For birthdays I am using a static calendar file saved in a folder using `https://localhost:8080/PATHNAME/calendar.ics` to feed it to the mirror. Before the actual date of the event the Birthdays are displayed correctly. On the actual day of the event, the events don't show up in the mirror. This goes only for **reoccurring full day events.** Full day events without reoccurance and reoccurring events which are not full day work fine.

**Steps to Reproduce**:

- Create and ics file with the following format:
```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Static Birthdays//Rafal Rüßmann//EN
CALSCALE:GREGORIAN
X-WR-CALNAME;VALUE=TEXT:Birthdays_static_short
X-WR-TIMEZONE:Europe/Berlin
BEGIN:VTIMEZONE
TZID:Europe/Berlin
LAST-MODIFIED:20200806T101500Z
BEGIN:STANDARD
DTSTART:20191027T030000
RRULE:FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:20200329T020000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
TZNAME:CEST
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
SUMMARY:Max Mustermann
DESCRIPTION:
DTSTAMP:20050806T080000Z
DTSTART;VALUE=DATE:20050806
DTEND;VALUE=DATE:20050806
RRULE:FREQ=YEARLY
END:VEVENT
END:VCALENDAR
```
- Save calendar locally on MagicMirror
- Feed calendar to via `localhost`

**Expected Results**: Seeing the full day even throughout the day of the event in the calendar.

**Actual Results**: Event only showing a day before but not on the actual day.

**Configuration**: 

```javascript
[...]

PR #2121, opened 2020-09-01
fix full day recurring events not showing

sdetweil:
#2110 #2111 #2118 

- What does the pull request accomplish? Use a list if needed.
check for full day event and NOT adjust timezone time (as it should be midnight local), before calling rrule to get the events from the rule. adjusting the time causes today to be skipped as its past midnight.


sdetweil:
@MichMich what is wrong here, travis builds failed no env set..  that is not a code problem

sdetweil:
@MichMich   this build failed cause prettier failed in travis

The command "npm run test:prettier" exited with 1.

MichMich:
I fixed the issue. For future reference:
`npm run test:prettier` shows the issues, `npm run lint:prettier` fixes the issues. :)

sdetweil:
I thought the whole point of the husky hook was to not have to run anything?

I don't see where the problem was

rejas:
Yeah that should happen automatically when you git commit (see https://git-scm.com/docs/githooks#_pre_commit) Do you invoke git with --no-verify ?

sdetweil:
@rejas  no.

it did get invoked on my desktop, not on my Chromebook. it looked like a normal precommit hook for prettier. ( on my last commit/push)

trying to figure out what didn't work now. this is 1st commit / push since the prettier stuff was added in 2.12

I added it via precommit hook to all my repos

sdetweil:
ok, this is my fault.. when i added the changelog on my chromebook, the hook did not work. and i pushed the bad file.
when I went to my desktop, i forgot the changelog, so when the hook fired it only processed the js file..

sorry.. still didn't see the error in the job log..

Issue #2130, opened 2020-09-19
Module config not deep merging.


MichMich:
See issue #2113.

rejas:
A good place for the code would be in the Utils class, I think.

bugsounet:
and why not [this](https://github.com/MichMich/MagicMirror/pull/2114) too ? pfff

MichMich:
@bugsounet For the last time: I kindly ask you to stop. Thanks.

rejas:
Becuase like @MichMich said: You are not helping anyone with the way you are acting right now. So please take some time to cool down.

bugsounet:
ok i give my link of the PR ... ok ... so delete too ...

MichMich:
(I removed some off-topic comments to keep this issue clean. Feel free to open a separate issue to discuss further.)

MichMich:
> A good place for the code would be in the Utils class, I think.

Agree!

dvbit:
a good thing would be start talking to all devs

sdetweil:
how do I do this with utils?  only place utils is used is on the server side..

```
	setConfig: function (config) {
		//this.config = Object.assign({}, this.defaults, config);
		this.config = Utils.configMerge({}, this.defaults, config);
	},
```
gets Utils  not defined in browser

MichMich:
We can just import utils in the browser just like any other js file.

rejas:
We could if there wasn't the colours dependency in there. Either rework that or use a browser-require solution...

MichMich:
Or take a simple route: BrowserUtils.js :)

sdetweil:
vs just add it to the module.js? it's only appropriate in this one context

rejas:
@sdetweil true

MichMich:
Agee. If is is used in other spaces as wall later, we can always refactor it. 👍🏻

sdetweil:
so, can we reopen pr 2114? which does that?

MichMich:
Reopened it. Will take a look later.

sdetweil:
thanks

bugsounet:
> Or take a simple route: BrowserUtils.js :)

it will be difficult ... you can't require any file

better way, include it in modules.js like `cmpVersions()`

sdetweil:
@bugsounet  we got it back to your submission. all good

bugsounet:
I know ... I try many solutions but not works ... the require function of electron is disabled by default
[...]

PR #2132, opened 2020-09-20
configMerge() and Correct test of Compliments module

bugsounet:
PR Number: 16/16... configMerge()

related to [#2130](https://github.com/MichMich/MagicMirror/issues/2130) and others...

1) If you don't want to respond in public ... find my mail ...
2) because i'm nice ... i will recreate it... but: don't count on me for other PR... (you know why... see your mail)


bugsounet:
what's this kind of error with Compliments module ?

rejas:
Maybe your merge code doesnt handle the config style in that module correctly?

sdetweil:
you change the function name to MMmergeConfig

but the code still trying to call old merge config

see the build log ( hate that the build log is one more click away than before )

bugsounet:
It's done Sam, result is the same

bugsounet:
If ... that i think ... I have to put my v1.1.0 ... :'(

I will do some test in local

bugsounet:
no it's good...

anytime: Array(1)
0: "Hey there sexy!"
length: 1
__proto__: Array(0)

test config:
anytime: Array(4)
0: "test!"
1: "test2"
2: "test3"
3: "test4"

I don't merge the Array but it's normal

bugsounet:
@rejas : it's a check test erreur ...
`AssertionError: expected 'You look nice!' to be one of [ 'Anytime here' ]`

it's logical !
i can be one result of: `evening: ["Wow, you look hot!", "You look nice!", "Hi, sexy!"],` OR result of `anytime`

in the script you check ONLY `anytime` result...

I test the check config ... i have `evening` or `anytime` result

bugsounet:
so 2 PR of My PR 16 ... fixed ... but I have just say ONLY one...

rejas:
Ah, now I see what happened here: Your PR merges the user-config (and therefore the test config) deep into the default config where the non-anytime fields are set.
Before your PR those fields where overwritten and therefor empty. Now they are filled with the default values and the tests can randomyl fail.

Unfortunately any user who has not overwritten those fields in his own config will now start to see those default compliments. Is that a behaviour you are willing to have @MichMich and @bugsounet ? IMO any change that that leads to other behaviour after an update is non-acceptable...

MichMich:
Hmm... valid point, let discuss here: https://github.com/MichMich/MagicMirror/issues/2130#issuecomment-696024633_

MichMich:
Closed in favor of: https://github.com/MichMich/MagicMirror/pull/2134

PR #2161, opened 2020-10-11
Fixes the un-hide problem with currentwether and weatherforcast modules.

Snille:
- Does the pull request solve a **related** issue?
I don't really know if there is an "issue" created. I just read three different threads in the forum regarding the "automatic un-hide". And I have myself had the same problem.

- What does the pull request accomplish? Use a list if needed.
Gets rid of the problem that you never can hide the "currentweather" and "weatherforcast" for more then the time it updates.

- If it includes major visual changes please add screenshots.
No visual changes. Just piece of mind. :)

Issue #2164, opened 2020-10-13
Weather modules dont show up anymore


rejas:
After https://github.com/MichMich/MagicMirror/pull/2161 got merged into develop my weather modules (current & forecast) dont show up anymore.

Not sure what happens but IMHO that PR seems to be incomplete and breaks existing configs and should therefor should be reworked.

Snille:
Hmm... Sorry, don't know why its not working for you, all is good here, my weather modules keeps showing them self even if I have hidden them with another module. Do you not experience this?

rejas:
I dont hide my weather modules from other modules. How do you do that?

sdetweil:
@rejas 
MMM-Voice will allow you to hide individual modules on command
my MMM-SleepWake provides an option to hide all modules instead of turning off the screen (where not possible)

there are others

Snille:
@rejas I'm using my own MMM-Modulebar to hide and show modules. And also I'm using "profiles" MMM-ProfileSwitcher to show different modules for different users (@ different times of the day). But the problem is that both the weather modules do not "respect" that they have been hidden. They keep showing up when they update (default every 10 minutes).

sdetweil:
their getDom() routines should become aware of the state and not draw when hidden
they should have suspend() and resume() methods that track the state

MichMich:
This might be an issue in the core itself. getDom should not be called when they are hidden.

sdetweil:
@MichMich yep. updateDom() does not check the visibility state before calling getDom()

MichMich:
@Snille @rejas Does https://github.com/MichMich/MagicMirror/pull/2166 solve the issue?

rejas:
No, I only see a big "undefined" in the place of the module. See my comment in the PR for possible reason.

MichMich:
I added a new commit returning an empty promise. Could you give it another try, @rejas?

rejas:
The undefined was an error on my side. Still, returning an empty promise is the better way I think. So: thumbs up from me for this.

sdetweil:
looks good..  i  think the real bug was in updateDomWithContent, 
but this approach is cleaner

```
var updateDomWithContent = function (module, speed, newHeader, newContent) {
		return new Promise(function (resolve) {
			if (module.hidden || !speed) {                       // <--- maybe ig(!module.hidden && !speed)// speed handled later
```

Snille:
[...]

PR #2166, opened 2020-10-14
Do not run updateDom() on hidden modules.

MichMich:
Some modules update their dom even when they are hidden. This PR prevents modules from running updateDom whenever the module is hidden. This should fix #2161 #2164.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -111,7 +111,7 @@ var MM = (function () {
 	 */
 	var updateDom = function (module, speed) {
 		if (module.hidden) {
-			return new Promise();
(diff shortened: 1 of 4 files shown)


js/main.js
```diff
@@ -110,6 +110,10 @@ var MM = (function () {
 	 * @returns {Promise} Resolved when the dom is fully updated.
 	 */
 	var updateDom = function (module, speed) {
+		if (module.hidden) {
+			return;
```
rejas, line 114: The function should return a Promise.

MichMich: Do you suggest an empty promise?

PR #2178, opened 2020-10-26
Fix calendar when no DTEND record in event 

sdetweil:
fixes  #2177 #2171 #2180

PR #2184, opened 2020-11-06
Add new log level "debug"

rejas:
Added a new log level for usage during development.

Can be used in https://github.com/MichMich/MagicMirror/pull/2178/files# so that no debug flags need to be when logging during development.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2021-01-01._
 
 ### Added
 
+- Added new log level "debug" to the logger.
+
 ### Updated
 
 - Weather module - forecast now show TODAY and TOMORROW instead of weekday, to make it easier to understand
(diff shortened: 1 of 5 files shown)


khassel:
With the change of 

````
logLevel: ["DEBUG", "INFO", "LOG", "WARN", "ERROR"],
````

in the `config.sample.js` now everybody using this sample as starting config has `DEBUG` enabled. This is also the default, if logLevel is not set. 

@rejas  Was this intended?

rejas:
@khassel not really... PR incoming :-)

rejas:
the question would be @MichMich @sdetweil @khassel what log level would be sufficient for the sample? Dont want to spam too much noise but dont want to miss too much. info? log? warn?

khassel:
for me the "old" config with `logLevel: ["INFO", "LOG", "WARN", "ERROR"],` would be o.k.

sdetweil:
correct  don't need debug in the sample as default setting

Issue #474, opened 2016-10-19
Weather config enhancements


zydecat:
I would like to suggest a couple of minor enhancements that are configurable.

Firstly, with the Current Weather module, there is the option to display Beaufort Scale wind speeds, or the default data from openweather, which is in metres per second. Would it be possible to add a config option showing wind speed as kilometres per hour? I have forced the change on my mirror on occasion by changing the following variable:

> this.windspeed = this.roundValue((data.wind.speed) \* 60 \* 60 / 1000) + 'km/h'

Windspeed in km/h makes more sense to me, and wind speed information in Australia is generally given in km/h.

Secondly, with the Weather Forecast module, I'd like to be able to configure the roundValue for the forecast temperatures to zero, rather than leave it at 1. Again, forecast temperatures are generally delivered with no decimal points.

I can go and make these changes in the code of each of the above modules myself, but it is annoying to have to redo the changes if I update the mirror to the latest version.

Anyway, that is my two cents worth.

MichMich:
Hi @zydecat, if you make these changes yourself and make them configurable, you can send a Pull Request so I can incorporate it in the next version.

zydecat:
Ok I will give that a go. Thanks!

PR #2185, opened 2020-11-07
Weather config enhancement

Sub028:
Add new parameter 'useKmh' to display wind speed in km/h instead of m/s.

The issue #474 (https://github.com/MichMich/MagicMirror/issues/474) was previously created but closed without any modifications about this point. So I add this feature to default module "weather".

### DIFF
--- a/modules/default/weather/current.njk
+++ b/modules/default/weather/current.njk
@@ -9,7 +9,11 @@
                 {% if config.useBeaufort %}
                     {{ current.beaufortWindSpeed() | round }}
                 {% else %}
-                    {{ current.windSpeed | round }}
+                    {% if config.useKmh %}
+                        {{ current.kmhWindSpeed() | round }}
+                    {% else %}
+                        {{ current.windSpeed | round }}
+                    {% endif %}
                 {% endif %}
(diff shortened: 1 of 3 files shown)


modules/default/weather/current.njk
```diff
@@ -9,7 +9,11 @@
                 {% if config.useBeaufort %}
                     {{ current.beaufortWindSpeed() | round }}
                 {% else %}
-                    {{ current.windSpeed | round }}
+                    {% if config.useKmh %}
```
rejas, line 12: I would find it nicer if you made that switch in the js code and just pass the windspeed here like before. That way the calculation (logic) stays in the js and display-related stuff stays in the njk.

Sub028: Ok i will try to revert changes in njk and implement modification in js.

Sub028: According to your suggestion, I make some changes, see commit 1460f002ab79c0aa209e56a44a41b91291d26d8e

(I don't know how to change the commit reference in this pull request. Previous commit is no longer necessary.)

PR #2186, opened 2020-11-07
Use Log.debug where applicable

rejas:


sdetweil:
we don't need to generate all that info all the time
I would not approve this change

rejas:
So if we dont need the info then we can remove it from the code?

sdetweil:
we don't need it generated ALL the time. but on demand if a user has a
problem. we don't need to waste the cpu cycles.

rejas:
If on demand is what you want then why not use the Log.debug? It is easily turned on via the config.js and it doesnt take much more cpu then an if clause ...

sdetweil:
ok, I have never seen any instruction on using config.js for setting log levels

rejas:
Submitted that a while ago: https://github.com/MichMich/MagicMirror/pull/2017
Is there a place to document that better?

sdetweil:
well I think it would have to go into the pages for the config.js doc

MichMich:
I agree with sam that we should add information about this in the docs.

PR #2385, opened 2020-12-29
Add missing function call in module.js

rejas:
I noticed the show method is missing a `()` in the show method when calling the callback. The hide method has it for example. 

Also cleaned up the jsdoc and adjusted some log levels

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ _This release is scheduled to be released on 2021-01-01._
 - update Node-ical 0.12.4 , fix invalid RRULE format in cal entries
 - fix package.json for optional electron dependency (2378)
 - update node-ical version again, 0.12.5, change RRULE fix (#2371, #2379)
+- Added missing function call in module.show()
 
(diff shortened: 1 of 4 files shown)


sdetweil:
I don't think the log level needs to be warning on the open whitelist.
its not bad

rejas:
>  its not bad

It isnt bad per se but the log is in the warning color already. And why not show the user if he opens his ip list a lot? A more experienced user will know how to change his log level IF it really annoys him ;-)

fewieden:
@rejas nice. Stumbled over the issue today as well.

rejas:
@fewieden so someone is using that call actually ;-) ?

fewieden:
@rejas yes, in order to only blur other modules when I actually can show the modal here https://github.com/fewieden/MMM-Modal/pull/7/files#diff-62b05431c54d3d6bceb46bb4e84772f44070a474852a5272e9e468bc0692299bR329-R333

Ideally, we also trigger the callback with an error here if the lock strings prevent from showing and not only on success.
I would like to send another module a notification in that case.

PR #2387, opened 2020-12-30
No Text Select for Touch Screen use

bugsounet:
I code some module with TouchScreen Feature
Some time the text is selected

maybe, the better way is disable the select text in main css?


bugsounet:
it's for the 0.01 € ? :D

sorry ...

PR #2390, opened 2020-12-31
update node-ical to latest 0.12.7, for try/catch around bad rrules

sdetweil:
another update for #2381


khassel:
the check for node12 is still running, may this job should be killed ...

MichMich:
Will merge with PR #2392.

PR #2393, opened 2020-12-31
Permit substituting a falsy value in translator variables

### LINKED ISSUE #2389: Translator can't substitute empty strings
**Platform**: Raspberry Pi 4

**Node Version**: 10.23.0

**MagicMirror Version**: 2.13.0

**Description**: In translator.js, the code checks whether a variable is present with the following code:

    return variables[varName] || "{" + varName + "}";

However, when the content of the variable is an empty string, this erroneously substitutes `{varName}` back in.    

**Steps to Reproduce**: Call `this.translate("something with {VAR}", {VAR: ""})`

**Expected Results**: If the variable is present with the empty string, the empty string should be substituted.  This could be achieved by using `return varName in variables ? variables[varName] : "{" + varName + "}";` instead.

**Actual Results**: Variable is not subsituted

MikeBishop:
Fixes #2389; this no longer evaluates the provided value, merely the presence of the key.  That means values which are falsy to JS (empty string, undefined, number 0, etc.) now get substituted where before they didn't.

For some possible values, it's possible that's not what you wanted -- but leaving {VARNAME} in place probably wasn't either, so 🤷‍♂️.  It fails differently now.  For empty strings, zeros, etc. this fixes a bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ _This release is scheduled to be released on 2021-01-01._
 - update node-ical version again, 0.12.5, change RRULE fix (#2371, #2379)
 - Added missing function call in module.show()
 - remove undefined objects from modules array (#2382)
+- Translator variables can have falsy values (e.g. empty string)
 
 ## [2.13.0] - 2020-10-01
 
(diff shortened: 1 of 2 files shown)


rejas (approved):
LGTM!

PR #2395, opened 2020-12-31
Added support for optional DEGREE position in "FEELS" label

ashishtank:
In some languages the value (number) should be first and then sentence. Added support for Degree label so it shows correct translation for Hindi and Gujarati with degree label at first and not at last.


rejas:
Please revert @MichMich or please @ashishtank  fix this issue that got introduced: Using the weather (and not the currentweather-) module the translation looks like this:

grafik

Porblem is both modules use the same translation but not the same logic (weaather is doing it in NJK file)

ashishtank:
@rejas let me check

ashishtank:
@MichMich @rejas I could not find a way to pass value to translate. If you know how to do that please let me know else revert the pull request (or may be keep it as it only affect the languages with {DEGREE} placeholder ). I will try to fix it other way.

ashishtank:
Looks like in our custom translate filter we do not have provision to pass key and values ? 
image

rejas:
Indeed. So maybe add some params to it like the filters in weather.js have?

ashishtank:
Yes, I have solved it.. doing final testing.

MichMich:
Ah. I’ll wait revering it.

MichMich:
*reverting

ashishtank:
Finally solved it #2397, sending updated PR. Also added support for placeholder values for translate in nunjucks translate filter.

English
image


Gujarati
image

PR #2397, opened 2020-12-31
Added support for variables in nunjucks templates for translate filter

ashishtank:
Added support for variables in nunjucks templates for translate filter
Fix issue with weather module with DEGREE label in FEELS like

### DIFF
--- a/modules/default/weather/current.njk
+++ b/modules/default/weather/current.njk
@@ -69,7 +69,10 @@
         <div class="normal medium">
             {% if config.showFeelsLike %}
                 <span class="dimmed">
-                    {{ "FEELS" | translate }} {{ current.feelsLike() | roundValue | unit("temperature") | decimalSymbol }}
+                    {{ "FEELS" | translate({DEGREE: current.feelsLike() | roundValue | unit("temperature") | decimalSymbol }) }} 
+					{% if not config.feelsLikeWithDegree %}
+					{{ current.feelsLike() | roundValue | unit("temperature") | decimalSymbol }}
+					{% endif %}
                 </span>
             {% endif %}
             {% if config.showPrecipitationAmount %}
--- a/modules/default/weather/weather.js
+++ b/modules/default/weather/weather.js
@@ -52,7 +52,8 @@ Module.register("weather", {
 		onlyTemp: false,
 		showPrecipitationAmount: false,
 		colored: false,
-		showFeelsLike: true
+		showFeelsLike: true,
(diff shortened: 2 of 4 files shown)


modules/default/weather/current.njk
```diff
@@ -69,7 +69,10 @@
         <div class="normal medium">
             {% if config.showFeelsLike %}
                 <span class="dimmed">
-                    {{ "FEELS" | translate }} {{ current.feelsLike() | roundValue | unit("temperature") | decimalSymbol }}
+                    {{ "FEELS" | translate({DEGREE: current.feelsLike() | roundValue | unit("temperature") | decimalSymbol }) }} 
+					{% if not config.feelsLikeWithDegree %}
```
rejas, line 73: I'd rather add the DEGREE placeholder in all other translations too, so we can get rid of these if clauses.

modules/default/weather/weather.js
```diff
@@ -88,6 +89,8 @@ Module.register("weather", {
 		// Let the weather provider know we are starting.
 		this.weatherProvider.start();
 
+		this.config.feelsLikeWithDegree = this.translate("FEELS").indexOf("{DEGREE}") > -1;
```
rejas, line 92: with DEGREE in all translations this would not be necessary either

rejas (changes requested):

modules/default/weather/weather.js
ashishtank: I would suggest to keep it as it is. As it is optional and very few languages will need DEGREE placeholder. If we add DEGREE placeholder it might break other modules who are using FEELS like resource string in their modules. New languages (Hindi and Gujarati) only have it now and if any third party is using it they need to correct it for this lang. only.

MichMich:
@Rejas, does this solve the issue for you? Should I merge this or revert the other?

rejas:
@MichMich that solves the issue. you can merge it. still the question is: do other modules rely on this translation? becuase I really think this should be solved in the translation files and not in the code. or what is your stance on this. 
We can always "clean it up" in 2021 :-)

MichMich:
Yeah. I agree we should clean it up. It's a bit too specific. But for now I'll merge it.

ashishtank:
@MichMich @rejas I will clean it up in 2021 ! Happy new year 🎊

MichMich:
U2! 🎉

rejas:
Happy new year to everybody! Stay healthy!

PR #2401, opened 2021-01-02
Add github action for uploading codecov results

rejas:
I added codecov testing coverage a while ago. This PR would now add a github action that uploads the results to codecov io

It is only a draft since:

a) I dont know if @MichMich wants it :-) PLus he would need to register the MM repo with codecov.io

b) I dont know on what occasion we should upload the results? Every push? Or only on merges?

c) Other comments?

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "/home/runner/work/rejas/MagicMirror/::"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,0 @@
-fixes:
-  - "/home/runner/work/rejas/MagicMirror/::"
--- a/.github/workflows/codecov-test-suites.yml
+++ b/.github/workflows/codecov-test-suites.yml
@@ -1,4 +1,6 @@
-name: Run Test Suites
+# This workflow runs the automated test and uploads the coverage results to codecov.io
+
+name: "Run Codecov Tests"
 
 on: [push]
 
@@ -12,4 +14,4 @@ jobs:
           export DISPLAY=:99
           npm install
           npm run test:coverage
-          bash <(curl -s https://codecov.io/bash) # Upload to Codecov
+          bash <(curl -s https://codecov.io/bash)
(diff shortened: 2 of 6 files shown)


codecov.yml
```diff
@@ -0,0 +1,2 @@
+fixes:
+  - "/home/runner/work/rejas/MagicMirror/::"
```
rejas, line 2: This woul dhave to be adjusted to the original repo

MichMich:
I like the idea. But I don't think this will help making the tests more complete. This will only show how awful the current test coverage is. 😂

Speaking of: the test suite could be so much improved. Currently for every test we start electron, while in most cases we could just "virtually" start the module and test what getDom returns.

codecov.yml
fewieden: @rejas I liked the idea of the automatic coverage check and added it to one of my modules. I first tried what you did and I could find online, but couldn't make it work properly.

Then I switched to using the action provided by codecov itself, where I didn't need to specify the path override.

https://github.com/fewieden/MMM-soccer/blob/72f1ef0ba84ece103e639faaadcdfaf1676c0ae6/.github/workflows/coverage.yml#L30-L48

rejas: Thanks for the hint @fewieden I switched the PR to use the codecov action. Any more helpful hints?

MichMich:
Good idea!

MichMich:
Do you have some info on what I have to do to make it work on the mm repo?

rejas:
> Do you have some info on what I have to do to make it work on the mm repo?

Only register the repo (and yourself via github) on https://codecov.io/

MichMich:
Done. Hope I did it correctly. :)

rejas:
Maybe we can take add the badge from codecov to the README. You should find under https://app.codecov.io/gh/MichMich/MagicMirror/settings/badge so we can see if it works

fewieden:
Badge this is the badge for develop

Badge this is the badge for master

MichMich:
Added it to the readme. Used the master because this will end up in the readme on the master branch. Or is there a markdown template thing to use the branch you are currently viewing?

Issue #2221, opened 2020-12-02
MMM-weatherforecast. Icon problem - still displays the same.


majdzik84:
Today I noticed that the icons in the module are not updating. It’s still wi-night-cloudy.
I don’t know at what point it broke. I know it used to work properly.
Can I fix it somehow?

![1606906890664-7f930a2f-3358-474e-9150-6dbe97140b5f-image](https://user-images.githubusercontent.com/31352476/100932673-839d9180-34ec-11eb-9237-cc686f47681d.png)
![image](https://user-images.githubusercontent.com/31352476/100933606-e2afd600-34ed-11eb-969d-205976653c95.png)

![1606907148924-922908ca-bfba-437a-a2a7-94df65fa4a32-image](https://user-images.githubusercontent.com/31352476/100932679-85675500-34ec-11eb-9327-4588fc811437.png)

MichMich:
Please refer to http://forum.magicmirror.builders for support and troubleshooting.

I'm closing this issue for now. If you feel this issue should remain open, please let me know.

majdzik84:
I believe the problem should remain open.
It also appears in the main module of magic mirror (weatherforcast). I found that this is not a problem with non-changing icons as they change but only in NIGHT mode. The icons do not switch to DAY. So I always have a moon icon - even at 12.00 when the sun should be.
Could you please check with yourself?

MichMich:
The 401 error is caused by the limitation of your (free account) which can be solved by limiting the number of days via the config. Please refer to the forum for more info. The icons seems to work fine here. Are  there more reports by users with the same issue? Is it a confirmed bug?

majdzik84:
I suggest not to look at the error because I have already figured it out.

The icons that do not change (day / night) are a mistake that I confirmed when shopping for 2 people because I asked so many

12:30
![130157057_3826110194090375_8947662787641997918_o](https://user-images.githubusercontent.com/31352476/101503978-fb4b4080-3972-11eb-989a-a8be18142d56.jpg)

14:40
![130157057_3826110194090375_8947662787641997918_o (1)](https://user-images.githubusercontent.com/31352476/101504092-1ddd5980-3973-11eb-92fd-31314ec6f877.jpg)

MichMich:
Reopened.

ashishtank:
Hi @majdzik84 

You are referring to below icons ? If yes, then I could also reproduce the issue and I know the cause so might also able fix it.
[...]

PR #2413, opened 2021-01-10
Issue 2221 Weather module - Always displays night icons because of fixed day start and end time

ashishtank:
Added support to get day start and end time from city sunrise and sunset time if available.
Also added unit test cases for data with and without sunset time.

Refer : #2221


khassel:
@ashishtank  I think we have problems with the tests after merging this.
The tests are now depending from the timezone.

In my `TZ=Europe/Berlin` its now ~ 23:00 and starting `npm run test:unit` succeeds.

Set `TZ=Australia/Sydney` test results in:
````
  104 passing (3s)
  1 failing

  1) Functions module weatherforecast
       forecastIcons
         forecastIcons sunset specified
           returns correct icons with sunset time:

      AssertionError: expected 'wi-night-cloudy' to equal 'wi-rain'
      + expected - actual

      -wi-night-cloudy
      +wi-rain

      at Context.<anonymous> (tests/unit/functions/weatherforecast_spec.js:84:37)
      at processImmediate (internal/timers.js:461:21)
````

Set `TZ=America/New_York` test results in:
````
  104 passing (3s)
  1 failing

  1) Functions module weatherforecast
       forecastIcons
         forecastIcons sunset specified
           returns correct icons with sunset time:

      AssertionError: expected 'wi-night-rain' to equal 'wi-rain'
      + expected - actual

      -wi-night-rain
      +wi-rain

      at Context.<anonymous> (tests/unit/functions/weatherforecast_spec.js:84:37)
      at processImmediate (internal/timers.js:461:21)
````

ashishtank:
@khassel : Thanks for the feedback, I looked at it but could not figure out how I can set default time zone for this tests so it always takes Europe Time zone while executing, as the test data is for Poland.

ashishtank:
@khassel I have solved the issue by setting time zone in test case and changing the way we create dates in weather forecast module. Tomorrow after some more testing will create pull request.

fewieden:
@ashishtank the keyword here is mocking. You need to mock the date so it will always be the same for the test case. This way you can simulate executing the logic at night and day at the same time to cover both. Even if you force the timezone to Europe you will have different results if you run the tests at day or night.

rejas:
@ashishtank see https://github.com/MichMich/MagicMirror/search?q=mockDate for an example on how in the compliments module the date is mocked

ashishtank:
Thanks for the tips Guys. 

Original issue was with locale that hour sign was being set to Unicode text ( ७ in Hindi) which was not being compared to offset (8). After solving that we found that for many countries day starts/ends different then our hard coded offset so in this PR we are using sunset and sunrise time if returned by weather provider. Now as this time is in EPOCH it was creating some issue with different time zones for testcase.
So I have corrected the way we are creating date objects in weather forecast module now for this and set the default time zone in test case.

I have tested this with multiple time zones with different times. Hope this will work as expected. Refer PR #2417

PR #2414, opened 2021-01-10
Feels like translation code cleanup

ashishtank:
Code cleanup for feels like translation for current weather and weather modules.


rejas:
Thanks for the PR. 
But what is the policy with translations @MichMich : Can other modules rely on the translations to not change? Or should they watch out for a change like adding DEGREE to this one here? 
In the former case we should use a new name for the translation (like FEELS_WITH_DEGREE or something) 
In the later case this PR is fine :-)

MichMich:
Honestly, for these kind of things we shouldn't worry too much. We're not building a space ship or hospital hardware. 😅

PR #2420, opened 2021-01-15
Add new Notification CURRENTWEATHER_TYPE

### LINKED ISSUE #2416: Compliments dont show based on current data from weather module
The compliments  modulerelies on the CURRENTWEATHER_DATA notification to display the compliments for a specific weather. 

Only the currentweather module is sending it, not the new one (simply called weather).

rejas:
- send it from the weather and currentweather module
- use it in the compliments module directly instead of the data

Fixes #2416


rejas:
PR is ready @MichMich, added a test for it too, but it seems coverage doesnt take that into account. But that is a problem for future-rejas :-)

PR #2422, opened 2021-01-16
Refactoring newsfeed-module to use templates instead of dom-generation

buxxi:
Made the newsfeed-module use templates and css-files instead:
- Moved code that didn't really have anything to do with rendering to other places (hiding module while loading, stripping of tags)
- Moved style-attributes for "full article"-notification to use classes instead in a css-file
- Separated the rendering to 3 different templates (one for migrating old configs, one for full article and one for the normal viewing)

The only thing that differs slightly is that the description isn't rendered behind when full article is displayed, but since that is placed above everything else the user wouldn't see that anyway (and the title was hidden so I guess that the description being shown was a bug).


rejas (approved):
LGTM!

PR #2423, opened 2021-01-16
Added start:dev command to npm scripts

rejas:
I noticed there was the option to start electron with the dev command line argument. So I figured it would make sense to add an entry for that in the package.json scripts section for this.


MichMich:
Thanks! Would you mind sending a PR to the docs repo as well?

rejas:
> Thanks! Would you mind sending a PR to the docs repo as well?

Will do. Which section do you think would it fit in @MichMich ?

MichMich:
No idea. 😂

rejas:
I just found out it's already there:
https://docs.magicmirror.builders/getting-started/installation.html#usage

PR #2427, opened 2021-01-20
Issue translate encoded html

dannoh:
I didn't create an issue for this, I jumped right to PR, but happy to go back and create the issue if needed.

Problem:  The translated string of `"LOADING": "Loading &hellip;",` shows up exactly like that instead of `Loading ...` when you use `{{ "LOADING" | translate }}` in a nunjucks template.

I've marked all translated strings as 'safe' so they will be properly decoded now.  Since it's the input from a known file this seems safe to me.

Alternatives would be to manually mark all `| translate` as `| translate | safe` to ensure there are no encoded characters in it, and that didn't seem great.


fewieden:
@dannoh you probably also want to remove the safe filter from the following locations.

https://github.com/MichMich/MagicMirror/blob/3dbe8bfbbf538931e0a9c1e0c147e7d48ba42bc0/modules/default/weather/hourly.njk#L27
https://github.com/MichMich/MagicMirror/blob/10dc315f3b96ebb5e980e47a21011d44bb9da979/modules/default/weather/forecast.njk#L38
https://github.com/MichMich/MagicMirror/blob/10dc315f3b96ebb5e980e47a21011d44bb9da979/modules/default/weather/current.njk#L87

PR #2433, opened 2021-01-23
deprecate module currentweather and weatherforecast

buxxi:
- Adding logging when using deprecated weather-modules (had to add a NodeHelper for each)
- Updating sample config file to use the new weather module
- Added comment in code about not accepting new features
- Added warning in the old weather modules readme

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ _This release is scheduled to be released on 2021-04-01._
 - Added CodeCov badge to Readme.
 - Added CURRENTWEATHER_TYPE notification to currentweather and weather module, use it in compliments module.
 - Added `start:dev` command to the npm scripts for starting electron with devTools open.
+- Added logging when using deprecated modules weatherforecast or currentweather.
 
 ### Updated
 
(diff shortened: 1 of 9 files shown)


MichMich:
Thanks. I like this. Might give some issues with your other weather PR. Which one would you like me to merge first?

buxxi:
The other one, this one is easier to redo :)

MichMich:
Merged the other one. Is the weatherEndpoint still necessary in the config?

buxxi:
Yes, both before and after the other pull request.
Using forecast/daily will just get stuck on "Loading..." without specifying it, that's why I defined it like that in the sample config.

Could possibly make it so it makes sane guesses for weatherEndpoint based on type when it's not specified?

MichMich:
I think indeed it's a good idea to use a default one that works no matter what type of account you have. The simpler the necessary config, the better.

rejas (approved):

buxxi:
Removed the default value for weatherEndpoint and if it's not set have sane defaults.

MichMich:
Thanks for your great contribution!

Issue #2424, opened 2021-01-16
Calendar module enhancement: displaying full day events without time


uncleLem:
With relative timeformat, upcoming full day events are displayed like this: 

> Wednesday at 12:00 AM

which doesn't make much sense to me. I've tried to look into the documentation and the code, and it seems like there's no way to configure it at the moment. 

`calendar()` method from moments.js used in the relative time display can't format without time by default, but does allow using custom calendar formatters. They are a rather tricky, because due to different localizations one size won't fit all, but I believe I've nailed it by putting adapted locale-specific calendars into translation files.

Would you consider adding it to the main repository? The details of the change can be found [here](https://github.com/MichMich/MagicMirror/commit/5903adadf9af64b00fe9f21bebec98e58e773476). I've updated translation files for the languages I know, but others require at least minimal understanding of grammar, so I don't think I can't do it myself. Also, for configuring the calendar from translation files, I'm using coreTranslations directly because in this case I don't want the fallback to English version, but rather to default localized calendar from moments.js.

I am also thinking about using `dateFormat` from config to replace the default `"sameElse": "L"`, allowing customization in relative format too, what do you think of it?

MichMich:
If it's not too much work, would you mind creating a PR so we can see your proposed change?

uncleLem:
Created pull request #2434 

Sorry for the failed checks, I can't figure out what's wrong at the moment and I hope I'll be able to check them a bit later this weekend, but I hope that would be enough for now to explain the idea?

Alvinger:
@uncleLem you can use the config option **fullDayEventDateFormat** to set the format for full day events. That will allow you to skip the time part.

uncleLem:
@Alvinger `fullDayEventDateFormat` is used in absolute time format, and it's just a date format, while "calendars" used by moment.js in relative time format are more complex. So, even though this could be a workaround, it's not exactly what I'm going for, I personally find relative time format more appealing and I believe this aspect can be enhanced.

Alvinger:
@uncleLem I agree that there is always room for improvement. I also noticed that moment.js calendar() now has a config option for various formatting options. However, as you say, the options are tricky to implement due to localization. Also, it wouldn't solve your issue as full day events still would be displayed with time.

The way it currently works is "by design" to avoid issues with backward compatibility.

What I could do in the short term is to let "fullDayEventDateFormat" apply to "relative" events as well. How's that?

Alvinger:
[...]

PR #2434, opened 2021-01-23
Calendar module enhancement: displaying full day events without time

uncleLem:
Issue #2424 

Using localized calendar formats to display full day events without time in the label.

Before:
Before

After:
After


rejas:
Looks like you broke some automic tests for the calendar module, did you take a look at those yet?

modules/default/calendar/calendar.js
```diff
@@ -93,6 +93,22 @@ Module.register("calendar", {
...
+		if (this.config.timeFormat === "relative") {
+			this.fullDayEventRelativeCalendar = {};
+			let calendarTemplate = this.Translator.coreTranslations["FULL_DAY_EVENT_RELATIVE_CALENDAR"];
+			if (calendarTemplate) {
+				Object.entries(calendarTemplate).forEach(([key, value]) => {
+					if (value && value.startsWith("function")) {
+						value = "(" + value + ")";
+					}
+					if (value && value.startsWith("(function")) {
+						value = eval(value);
```
buxxi, line 105: Really think eval() should be avoided, if we actually want dynamic code in the translations a `new Function(...)` should probably be used instead and not let the function have access to global scope.

buxxi (commented):
I really think eval should be avoided and having code in each translation file feels like a bad solution.

translations/pl.json
```diff
@@ -32,5 +32,13 @@
 	"UPDATE_INFO_MULTIPLE": "Zainstalowana wersja odbiega o {COMMIT_COUNT} commitów od gałęzi {BRANCH_NAME}.",
 
 	"FEELS": "Odczuwalna {DEGREE}",
-	"PRECIP": "Szansa opadów"
+	"PRECIP": "Szansa opadów",
+	"FULL_DAY_EVENT_RELATIVE_CALENDAR": {
+		"sameDay": "[Dziś]",
+		"nextDay": "[Jutro]",
+		"nextWeek": "function () { switch (this.day()) { case 0: return '[W niedzielę]'; case 2: return '[We wtorek]'; case 3: return '[W środę]'; case 6: return '[W sobotę]'; default: return '[W] dddd'; } }",
```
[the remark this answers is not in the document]
rejas: IMO that seems very counterintuitive. Placeholders are fine, but this here should somehow be done in the js code

modules/default/calendar/calendar.js
rejas: Yes, eval-ing a function ís generally frowned upon. Hopefully this goes away when @uncleLem removes the function from the translations

uncleLem: Replacing `eval` with `Function` is a good idea, I didn't think about it because I took these calendars directly from moments.js as they were and only changed the time part.
The differences in grammar wouldn't allow to have one approach for all, and I believe having all different variants in the calendar.js (like the moments.js stores it) would increase its size significantly and make it harder to maintain - all the calendars for all the languages existing in translations at the moment would add at least 350 lines. That's why I thought it would be better to keep language-specific part of it separate from the main logic in existing language-specific files.
But I don't insist on using translation files, I also think it may be confusing, but it looked right at the moment. Perhaps, a separate js file for storing these calendars would be a better solution? What do you think?

uncleLem:
> 
> 
> Looks like you broke some automic tests for the calendar module, did you take a look at those yet?

Didn't have enough time to figure this out yet, sorry. Unfortunately, I'm not very familiar with this environment, so it takes some time to figure it out. I hope I'll be able to fix the tests this week.

modules/default/calendar/calendar.js
buxxi: Since I neither speak polish nor russian I need a little bit of help to understand what you are trying to do with the function.

I'm guessing the function is used for the fourth row in your example ("Wednesday") where it has some special grammar depending on the weekday and just presenting the name of the weekday isn't logical?

- Today and Tomorrow are already in the translation-files, would be better to not have duplicate translations for the same thing.
- The weather-module would have the same problem then since it displays "Today, Tomorrow, <weekday name>"

uncleLem: @buxxi Yes, the function is required for some grammatical nuances. Sometimes it's the difference in prepositions for different weekdays, sometimes there are different grammatical cases that won't match simple `dddd` format, and also some day names are feminine, some are masculine, and some are neutral, changing the declension of related words. 
Simply displaying the name of a weekday like in English would be grammatically correct (for the languages I know), but in uk/pl/ru it would sound kinda synthetic and official, which is not exactly the aim of the relative time format from my understanding, it would look more like the absolute date format with weekdays. It seems to me that the English calendar in moment.js implementation also could have had "on" before the weekday, but moment.js devs implemented it differently from others in this regard.

> Today and Tomorrow are already in the translation-files, would be better to not have duplicate translations for the same thing.

You are totally right, it can be replaced (at least, for these languages) and I'll do it.

> The weather-module would have the same problem then since it displays "Today, Tomorrow, "

Is it? I believe it shows simply the weekdays, and I couldn't find any mentions of today/tomorrow. Also, my changes shouldn't affect other modules in any way.

uncleLem:
Added some changes.

@Alvinger I've added configuration parameter allowing to enable this new behavior, now it won't change anything by default.

@rejas @buxxi I've moved calendars to the js file, so there's no more js code in translation files and no more `eval()`. If calendar.js becomes too large, we can think about extracting them later.

Also, added `fullDayEventDateFormat` to calendar definitions.

I have added some tests, but it doesn't look like codecov counted as covered some lines that are checked by new tests. Also. I'm not sure how to test calendars definitions. Any advice?

buxxi:
I'm not really sure about the this approach either and I've tried to look around to see how other applications handle this without success. I really think all translations should be located in the translation files and without having logic in it even if it means multiple duplicates of translation values or having some kind of rule for fallbacks.

Something like this (just examples, english wouldn't have any of these):
```json
{
   "NEXT_WEEK_THURSDAY" : "On thursday (next week)",
   "NEXT_WEEK_DAY" : "On {DAY} (next week)",
   "NEXT_DAY" : "On {DAY}",
}
```
`NEXT_WEEK_THURSDAY` could fallback to `NEXT_WEEK_DAY` if not found and to `NEXT_DAY` if that isn't found and finally to moment.js translation of the day if that isn't found either. The same would have to be repeated for `THIS_WEEK_THURSDAY` and `LAST_THURSDAY` etc.

I think you should have @MichMich comment on this before doing any more at least.

MichMich:
Thanks for your PR. I understand what you are trying to accomplish, but the implementation seems way to complicated. I'm certainly not a big fan of all the localised snippets in the code. The end result of the PR is relatively small but this change is a huge maintenance burden. So I'm not going to merge this PR. Sorry!

PR #2439, opened 2021-01-28
Added error to callback

fewieden:

After adding `lockStrings` in a merge request of a module we were running into issues because the callback wasn't triggered in case lock strings were active. Which means you only get notified on success. The way nodejs and any popular libraries that still uses callbacks implement them are like callback(error) or callback(error, data). This way the caller gets always notified and can handle based on the error argument.

@skuethe with this, we can avoid manually checking for active lock strings again.

https://github.com/fewieden/MMM-Modal/pull/7#issuecomment-768625722

- What does the pull request accomplish?

* callback for show function always gets called.
* If lockStrings are still active the callback contains an error

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -295,6 +295,7 @@ var MM = (function () {
 		// Otherwise cancel show action.
 		if (module.lockStrings.length !== 0 && options.force !== true) {
 			Log.log("Will not show " + module.name + ". LockStrings active: " + module.lockStrings.join(","));
+			callback("Active lock strings");
 			return;
 		}
 
@@ -321,13 +322,13 @@ var MM = (function () {
 			clearTimeout(module.showHideTimer);
 			module.showHideTimer = setTimeout(function () {
 				if (typeof callback === "function") {
-					callback();
+					callback(null);
 				}
 			}, speed);
 		} else {
 			// invoke callback
 			if (typeof callback === "function") {
-				callback();
+				callback(null);
 			}
 		}
 	};
--- a/js/main.js
+++ b/js/main.js
@@ -295,7 +295,7 @@ var MM = (function () {
 		// Otherwise cancel show action.
 		if (module.lockStrings.length !== 0 && options.force !== true) {
 			Log.log("Will not show " + module.name + ". LockStrings active: " + module.lockStrings.join(","));
-			callback("ERR_ACTIVE_LOCK_STRINGS");
+			callback(new Error("ERR_ACTIVE_LOCK_STRINGS"));
 			return;
 		}
 
--- a/js/main.js
+++ b/js/main.js
@@ -279,8 +279,9 @@ var MM = (function () {
 	 * @param {number} speed The speed of the show animation.
 	 * @param {Function} callback Called when the animation is done.
 	 * @param {object} [options] Optional settings for the show method.
+	 * @param {Function} errorCallback Called when the module failed to show.
 	 */
-	var showModule = function (module, speed, callback, options) {
+	var showModule = function (module, speed, callback, errorCallback, options) {
 		options = options || {};
 
 		// remove lockString if set in options.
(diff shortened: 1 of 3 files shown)


js/main.js
```diff
@@ -295,6 +295,7 @@ var MM = (function () {
 		// Otherwise cancel show action.
 		if (module.lockStrings.length !== 0 && options.force !== true) {
 			Log.log("Will not show " + module.name + ". LockStrings active: " + module.lockStrings.join(","));
+			callback("Active lock strings");
```
rejas, line 298: Maybe rename it to "ERR_ACTIVE_LOCK_STRINGS" so it is a little clearer that it is a error string?

MichMich: Er even better: allow for an optional error callback?

fewieden: I changed the error for the callback for an actual error object. If you log the error it gives you an actual stack trace from its origin which helps to debug.

image

I think adding a second callback just for the error isn't necessary. Also, @rejas just fixed the callback recently. Until now it was never triggered, so we are not breaking existing modules with it.

rejas (approved):

js/main.js
MichMich: The problem of this is older modules might not check if the callback contains an error. They might think the action was executed successfully because the callback was called. So this is a somewhat breaking change.

fewieden: @MichMich if you want me to change it I can, but as I said, @rejas just fixed the callback which means until now the callback was never executed at all https://github.com/MichMich/MagicMirror/pull/2385/files#diff-50260f11109f55b944da641c51de244dffd718cbd77f7e5ce7b9124ab18fe265R437

MichMich: Ah, true. But still I think a separate error callback is a bit cleaner (and easier to uses as a module developer). Or do you disagree?

js/main.js
```diff
@@ -279,8 +279,9 @@ var MM = (function () {
 	 * @param {number} speed The speed of the show animation.
 	 * @param {Function} callback Called when the animation is done.
 	 * @param {object} [options] Optional settings for the show method.
+	 * @param {Function} errorCallback Called when the module failed to show.
 	 */
-	var showModule = function (module, speed, callback, options) {
+	var showModule = function (module, speed, callback, errorCallback, options) {
```
rejas, line 284: woudln't just "error" be the most common name for such a function (here and in the other occurences too)?

fewieden: not sure error sounds more like an object instead of a function

rejas: yes, you are probably right :-)

rejas (approved):

MichMich:
Just a thought: What if we make the error callback part of the options object?

rejas:
> Just a thought: What if we make the error callback part of the options object?

It's a fine balance.... If we do it, why not also make the callback part of it? Not sure if this way or the current one one is more readable or less error prone...

MichMich:
Because that would be a change of API. The regular callback is also something you often use. Error is something you only need in very specific occasions so can be tucked away. IMHO it's good to limit the amount of arguments a method has. Since error will only be interesting for a hand full of devs, it's ok to be tucked away. 

Of course we could also allow the main callback to be put in the options object and check to see if the 2nd argument is an object or method. If it is an object the callback can be extracted.

rejas:
You had me at "change of API" :-) I agree, changing API would be some "v3" change that we dont need to put here. +1 for your putting the error callback in the options and have the normal callback for API reasons in the main function call

MichMich:
And are we going to call it `errorCallback` or `onError`?

rejas:
I'd go for `onError` since there are already occurences of that name in the code.

MichMich:
Agree! `onError` sounds best. :)

fewieden:
- Moved errorCallback into options
- Renamed to onError
- Created PR in the documentation project
- Resolved merge conflicts

rejas (approved):

PR #2442, opened 2021-01-29
Fix/translation fallbacks

fewieden:

If a module has the same translation label as the core project and the user configured the language to English (because it is the fallback language), then it takes the core translation instead of the module translation.

https://github.com/fewieden/MMM-NHL/pull/13#discussion_r559732740

- What does the pull request accomplish?

* Load English translations for 3rd party modules as main translations.
* Added tests
* Untangled the callback hell
* ES6


rejas (approved):
good idea & pr! lgtm!

PR #2452, opened 2021-02-10
Added translation for today and tomorrow

TheDuffman85:
I had misread the documentation of Moment.js. We've to provide the translation for today and tomorrow ourselves. For the translation I use the standard MM² translation mechanism.


TheDuffman85:
I don't think we need to include this in the changelog. Because it's a fix for code that is not released anyway.

PR #2454, opened 2021-02-11
Cleanup tests

rejas:
As @khassel mentions in https://github.com/MichMich/MagicMirror/pull/2448#issue-568870330 I took a look at the weather-compliment and cleaned it up as good as I can. 

Some other cleanups in the test section flowed into this PR too.

### DIFF
--- a/tests/configs/modules/calendar/auth-default.js
+++ b/tests/configs/modules/calendar/auth-default.js
@@ -26,7 +26,7 @@ var config = {
 				calendars: [
 					{
 						maximumNumberOfDays: 10000,
-						url: "http://localhost:8011/tests/configs/data/calendar_test.ics",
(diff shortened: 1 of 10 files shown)


tests/configs/modules/calendar/auth-default.js
```diff
@@ -26,7 +26,7 @@ var config = {
 				calendars: [
 					{
 						maximumNumberOfDays: 10000,
-						url: "http://localhost:8011/tests/configs/data/calendar_test.ics",
+						url: "http://localhost:8080/tests/configs/data/calendar_test.ics",
 						auth: {
```
khassel, line 30: the ports must also changed in the tests ...

rejas: Fixed those. not sure why they were in there, so I removed them where they didnt seem needed...

PR #2461, opened 2021-02-16
added locale to sample config

fewieden:
Some of my modules make use of the locale config option. Since this could be useful for others I thought why not adding it to the sample config file.


rejas (approved):

sdetweil:
anything use this that will be broke if people upgrade, and its not there? like we had with logLevels?

fewieden:
@sdetweil no currently it is only used by 3rd party modules

PR #2464, opened 2021-02-21
exposed logger as node module

fewieden:
This makes it easier for 3rd party modules to use the logger in the node helper with the according `logLevels` instead of console logs.

You only need to do `const Log = require('logger');` in a module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,8 @@
 		"valid-url": "^1.0.9"
 	},
 	"_moduleAliases": {
-		"node_helper": "js/node_helper.js"
+		"node_helper": "js/node_helper.js",
+		"logger": "js/logger.js"
 	},
 	"engines": {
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -89,7 +89,8 @@
 		"valid-url": "^1.0.9"
 	},
 	"_moduleAliases": {
-		"node_helper": "js/node_helper.js"
+		"node_helper": "js/node_helper.js",
+		"logger": "js/logger.js"
```
rejas, line 93: Good thinking, like the idea. Too bad node_helper is already widely used, otherwise I'd propose to add some prefix to those aliases so that develoeprs would know that they are from the MM-core.

rejas (approved):

MichMich:
Thanks! Might be a good idea to add info to the docs about this.

PR #2465, opened 2021-02-25
Weather canada

Crazylegstoo:
I have created a new weather module provider for Environment Canada, which is the government weather service in Canada. The new provider code is **envcanada.js**. I have made minor updates to existing weather module code in support of the new provider:

**weatherobject.js** - A new object element is included = precipitationUnits. It is initialized as null, and can be optionally used to specify a custom unit measurement to be displayed in a forecast. The new envcanada provider uses this (see forecast.njk). If this object element is not populated, then current default processing is used to select the correct unit to display (e.g. 'in', 'mm', etc.).

**forecast.njk** - Updated precipitation logic such that if weatherObject has precipitationUnits specified, they will be displayed instead of the default logic that specifies 'in' , 'mm', etc.  This allows the new envcanada provider to use the precipitation units included in the Environment Canada forecast data. It also better reflects that Canadian forecasts tend to measure snow in 'cm' and rain in 'mm' . This also allows each day within a forecast to use its own units (since it can snow on one day and rain the next).

**weather.js** - The getScripts input has been updated to include 'providers/xml2json.min.js'. This utility function is used by the new envCanada provider to convert XML documents provided by Environment Canada into JSON structures, which are easier to work with in the provider code.

**providers/xml2json.min.js** - Has been added to the module codebase

**Environment Canada (EC) provider usage:**

This module supports Canadian locations only and requires 2 additional config parms for weather module:

siteCode - the city/town unique identifier for which weather is to be displayed. Format is 's0000000'.

provCode - the 2-character province code for the selected city/town.

**Example:** for Toronto, Ontario, the following parms would be used:

siteCode: 's0000458',
provCode: 'ON'

To determine the siteCode and provCode values for a Canadian city/town, look at the Environment Canada document
at [](https://dd.weather.gc.ca/citypage_weather/docs/site_list_en.csv) (or site_list_fr.csv). There you will find a table
with locations you can search under column B (English Names), with the corresponding siteCode under 
column A (Codes) and provCode under column C (Province). 

There is further EC Documentation at following links:
  	[](https://dd.weather.gc.ca/citypage_weather/schema/)
 	[](https://eccc-msc.github.io/open-data/msc-datamart/readme_en/)


modules/default/weather/forecast.njk
```diff
@@ -19,10 +19,16 @@
                 <td class="align-right min-temp">
                     {{ f.minTemperature | roundValue | unit("temperature") }}
                 </td>
-                {% if config.showPrecipitationAmount %}
-                    <td class="align-right bright precipitation">
-                        {{ f.precipitation | unit("precip") }}
-                    </td>
+                {% if (config.showPrecipitationAmount) and (f.precipitation > 0) %}
```
rejas, line 22: Why add that extra case? Cant the precipation be 0 ?

modules/default/weather/forecast.njk
```diff
@@ -19,10 +19,16 @@
...
+                    {% if f.precipitationUnits %}	
+                        <td class="align-right bright precipitation">
+                            {{ f.precipitation }}{{ f.precipitationUnits }}
```
rejas, line 25: Why not translate it like the else case does?                             {{ f.precipitation | f.precipitationUnits }}

modules/default/weather/providers/xml2json.min.js
```diff
@@ -0,0 +1 @@
...ct" in window;if(D===undefined){return null;}var E;if(window.DOMParser){var G=new window.DOMParser();var B=null;if(!F){try{B=G.parseFromString("INVALID","text/xml").getElementsByTagName("parsererror")[0].namespaceURI;}catch(C){B=null;}}try{E=G.parseFromString(D,"text/xml");if(B!=null&&E.getElementsByTagNameNS(B,"parsererror").length>0){E=null;}}catch(C){E=null;}}else{if(D.indexOf("<?")==0){D=D.substr(D.indexOf("?>")+2);}E=new ActiveXObject("Microsoft.XMLDOM");E.async="false";E.loadXML(D);}return E;};this.asArray=function(B){if(B===undefined||B==null){return[];}else{if(B instanceof Array){return B;}else{return[B];}}};this.toXmlDateTime=function(B){if(B instanceof Date){return B.toISOString();}else{if(typeof(B)==="number"){return new Date(B).toISOString();}else{return null;}}};this.asDateTime=function(B){if(typeof(B)=="string"){return a(B);}else{return B;}};this.xml2json=function(B){return A(B);};this.xml_str2json=function(B){var C=this.parseXmlString(B);if(C!=null){return this.xml2json(C);}else{return null;}};this.json2xml_str=function(B){return e(B,"");};this.json2xml=function(C){var B=this.json2xml_str(C);return this.parseXmlString(B);};this.getVersion=function(){return t;};};}));
```
rejas, line 1: Some minified code that one can hardly check for integrity. Could we add the offical library to the vendor files if it is really needed @MichMich ? Or does some othe rmodules maybe need that too?

rejas:
Please run the prettier command "npm run lint:prettier" to fix the code styles issues

modules/default/weather/forecast.njk
Crazylegstoo: You are right and this is ugly. Basically I wanted to ensure precipitation would only be rendered is there actually is a value (and selfishly this was more for my own envcanada code). I will fix this in my envcanada code and this case will be removed.

modules/default/weather/forecast.njk
Crazylegstoo: I will apologize as this is my first serious bit of JS development, so my templating skills are very basic. But in this instance there is really no translation needed since the Provider code (e.g. envcanada) can explicitly set the precipitation and units values it wants displayed. Based on this, I thought it would be cleaner/simpler to do it as I've done.  I did look at making no changes to forecast.njk and pushing this logic into the filtering provided by weather.js, but this seemed really messy for what it required. I'm open to suggestions!

modules/default/weather/providers/xml2json.min.js
Crazylegstoo: I would be happy with whatever makes sense for MM overall. Really, I just need a mechanism to create a JSON object from a supplied XML doc and found this minified code being used by another MM 3rd-party module.

rejas: Which one also uses it?

modules/default/weather/forecast.njk
rejas: No worries about being fresh in JS development, we all were once :-) Anyway since I am not into weather stuff, what units can precipation have? 
If there are multiple units possible, I'd rather move that into the weather.js config, something like showPrecipitationAmountUnit (but only if they are not covered with existing units).

Crazylegstoo: I'm really trying to solve 2 issues:

1. Today, if the config units are metric, precipitation will be expressed as 'mm' for all forecast days. This does not really work for Canada in that we measure snow in 'cm' and rain in 'mm'. So... I need a way to override that for envcanada.
2. In Canada, it is possible to have snow on one day and rain on the next. So for envcanada I need a way to let each forecast day to have a unique precipitation unit if required.

Based on these 2 requirements, I came up with the solution to add a new element to weatherobject to hold a custom unit for precipitation. If that element gets populated in Provider code, then it will be displayed. If that element is not populated (i.e. left as null), then default processing/filtering for precipitation will take place. This lets envcanada have its custom units without breaking anything for the existing Providers.

For sure I could add a new config boolean - e.g showCustomPrecipitaionUnit or something like that (weather.js already supports showPrecipitationAmount to toggle precipitation presentation). The new boolean would basically replace my check for a non-null value in my new weatherobject element precipitationUnits. Maybe that is a safer method to determine whether or not display the custom vs. default precipitation units?

And thanks for for all the feedback. Who knew Canadian weather could be so much fun? ;)

modules/default/weather/providers/xml2json.min.js
Crazylegstoo: A few examples: 
MMM-WordOfTheDay
MMM-Reddito

I believe MMM-Bose-Soundtouch is using similar function (not the minified version however).

buxxi:
Some of my thoughts:
- That's a lot of comments, nicely done!
- Is the XML2JSON-conversion really needed, it's not a very complicated XML-structure and could easily be parsed with DOMParser and querySelector.
- It seems kinda bad having to depend on some external proxy for removing CORS-headers (I know there is another provider doing that too), could we use a node_helper to make the requests not in the frontend?
- There will be lots of duplicate code for each provider handling the conversion between imperial and metric, each provider should probably not be doing that and just return what unit the data is in and the weather-module should handle the conversion to what should be displayed.

Crazylegstoo:
> 
> 
> Some of my thoughts:
> 
>     * That's a lot of comments, nicely done!

Thank you! When I started my first development role (writing IBM mainframe Assembler about 100 years ago), my manager was super-strict about commenting code to help out whoever would next maintain it. So I've made it a habit to write (sometimes verbose) comments ever since. lol

>     * Instead of overriding fetchData maybe the base-method shouldn't expect JSON, maybe have the resulting format as an argument?

That's a valid design thought. I guess that would require some re-factoring of weatherprovider.js template and at least a few of the Provider modules.  That said, overriding the base method does allow for some flexibility for Providers to manage their access to backend data/function.

>     * Is the XML2JSON-conversion really needed, it's not a very complicated XML-structure and could easily be parsed with DOMParser and querySelector.

I have no experience with the DOMParser API, but I did seriously look at it for my Provider. In the end I decided to focus on JSON because I just found the code more familiar and a bit cleaner for my purposes and the conversion overhead is not a concern (I'm running my MM device on a Raspberry Pi 4).

>     * It seems kinda bad having to depend on some external proxy for removing CORS-headers (I know there is another provider doing that too), could we use a node_helper to make the requests not in the frontend?

I agree that the better design would be node_helper for weather to push the raw data access logic into the back-end. I suspect that would require some serious refactoring of the module and changes to all Providers (but maybe I'm wrong).

>     * There will be lots of duplicate code for each provider handling the conversion between imperial and metric, each provider should probably not be doing that and just return what unit the data is in and the weather-module should handle the conversion to what should be displayed.

That's a valid argument, for sure - keep a clean break between the source data measurement units and whatever it is that should be displayed in the UI. This would require a bit of work in the module to support preferred unit of measure. Example - windspeed in metric is currently expressed in m/s I believe, so it would also need to support km/h. And there would have to be consideration for the possibility that a Provider could be talking to a backend that supports metric or imperial on request - meaning that either the Provider needs to be aware of unit-of-measure, or else defaults to 1 unit-of-measure and lets the module sort out all the conversions for UI purposes.

I think you've raised a lot of valid points about the module design and how it could be improved. I'm not sure that was in my original scope to create my little Environment Canada Provider :)

buxxi:
Yeah, most of it stuff that I found previously when refactoring the other providers as there being a single special case for just one provider, but now there's two :). Agree that it is outside of the scope for this PR, but stuff that should probably be fixed before there are too many providers.

> I have no experience with the DOMParser API, but I did seriously look at it for my Provider. In the end I decided to focus on JSON because I just found the code more familiar and a bit cleaner for my purposes and the conversion overhead is not a concern (I'm running my MM device on a Raspberry Pi 4).

It wasn't about performance, more about adding an external dependency for something that already has an API included.

If line 121 was:
```js
resolve(this.responseXML)
```
Then you could do it like this:
```js
currentWeather.temperature = this.convertTemp(ECDoc.querySelector("siteData currentConditions temperature").textContent);
```

Crazylegstoo:
> 
> It wasn't about performance, more about adding an external dependency for something that already has an API included.
> 
> If line 121 was:
> 
> ```js
> resolve(this.responseXML)
> ```
> 
> Then you could do it like this:
> 
> ```js
> currentWeather.temperature = this.convertTemp(ECDoc.querySelector("siteData currentConditions temperature").textContent);
> ```

Okay, no problem. It's a valid argument to reduce dependencies in the codebase. I'll rework my code to remove the JSON dependancy and leverage DOM API. Thanks!

Crazylegstoo:
Closing this PR and will open a cleaner one...

PR #2466, opened 2021-02-27
Enable contextIsolation in Electron

rejas:
Currently we get this warning in the console when running the MM:

`[ERROR] (node:852808) electron: The default of contextIsolation is deprecated and will be changing from false to true in a future release of Electron.  See https://github.com/electron/electron/issues/23506 for more information`

This PR enables contextIsolation already so that if there are problems with it, we can identify them sooner.


khassel:
@rejas thanks, this error message was already on my todo-list, now you were faster 👍

PR #2474, opened 2021-03-05
Node fetch

### LINKED ISSUE #2468: replace request
The `request` package is deprecated, see https://www.npmjs.com/package/request

So we should replace it with e.g. `node-fetch`.

It is used in 
- calendarfetcher.js
- newsfeedfetcher.js
- and several e2e tests

Can submit a PR, but want to discuss this before if this is wanted and if `node-fetch` is the right replacement.

khassel:
closes #2468

This PR replaces the deprecated `request` package with `node-fetch`.

The replacement in the newsfeed module and the e2e tests are straight forward.

In the calendar module are 3 auth methods used: base, bearer and digest. For digest the package `digest-fetch` is used additionally.

Remark: There are no tests available for bearer and digest auth, so these cases should be reviewed in detail.

### DIFF
--- a/modules/default/calendar/calendarfetcher.js
+++ b/modules/default/calendar/calendarfetcher.js
@@ -44,33 +44,34 @@ const CalendarFetcher = function (url, reloadInterval, excludedEvents, maximumEn
 	 * Initiates calendar fetch.
 	 */
 	const fetchCalendar = function () {
-		function getFetcher(url, auth) {
-			const nodeVersion = Number(process.version.match(/^v(\d+\.\d+)/)[1]);
-			let headers = {
-				"User-Agent": "Mozilla/5.0 (Node.js " + nodeVersion + ") MagicMirror/" + global.version + " (https://github.com/MichMich/MagicMirror/)"
-			};
-			let httpsAgent = null;
-
-			if (selfSignedCert) {
-				httpsAgent = new https.Agent({
-					rejectUnauthorized: false
-				});
-			}
-			if (auth) {
-				if (auth.method === "bearer") {
-					headers.Authorization = "Bearer " + auth.pass;
-				} else if (auth.method === "digest") {
-					return new digest(auth.user, auth.pass).fetch(url, { headers: headers, httpsAgent: httpsAgent });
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -71,6 +71,7 @@
 	"dependencies": {
 		"colors": "^1.4.0",
 		"console-stamp": "^3.0.0-rc4.2",
+		"digest-fetch": "^1.1.6",
```
rejas, line 74: where is that package used in your changes? cant find any imports of it

khassel: in `calendarfetcher.js`, the diff is collapsed ...

rejas: Doh :-)

modules/default/calendar/calendarfetcher.js
```diff
@@ -41,387 +44,383 @@ const CalendarFetcher = function (url, reloadInterval, excludedEvents, maximumEn
 	 * Initiates calendar fetch.
 	 */
 	const fetchCalendar = function () {
-		clearTimeout(reloadTimer);
-		reloadTimer = null;
-
-		const nodeVersion = Number(process.version.match(/^v(\d+\.\d+)/)[1]);
-		const opts = {
-			headers: {
+		function getFetcher(url, auth) {
```
rejas, line 47: why make a function here that is only used once?

khassel: willt take a lokk at this later

khassel: function is removed

rejas:
> Remark: There are no tests available for bearer and digest auth, so these cases should be reviewed in detail.

I am wondering how difficult it would be to write tests for those?

khassel:
the current tests for basic auth are using an express server with `express-basic-auth`, so for testing for digest and bearer we would need similar servers (see https://github.com/MichMich/MagicMirror/blob/develop/tests/servers/basic-auth.js).

rejas:
I see, but couldnt find ones in a five minute search :-( Oh well, maybe we cant test everything?

PR #2477, opened 2021-03-11
Allowing openweathermap config lat and lon to function without onecall, overrides locationID and location since more specific.

PostLogical:
OpenWeatherMap in the weather module would not function with just latitude and longitude (not using onecall). Adding two lines of code to allow them and assuming they should take precedence over locationID and location. Without this change, the module just says "Loading…" even with valid lon and lat set in config. Sorry for mistakes – first pull request.


rejas:
Please base your PR against the develop branch and include an entry in the CHANGELOG :-)

PostLogical:
Thanks for your patience!

PR #2479, opened 2021-03-13
Timeout handling for updatenotification

### LINKED ISSUE #1907: updatenotification creates zombie processes
I noticed my server had lots of zombie-processes where the parent was the container running magicmirror in serveronly-mode. By enabling only one module at a time I managed to locate the problematic module which is the updatenotification.

The cause of this seems to be partly my fault due to cloning my own modules with ssh instead of https (using url = git@github.com:xxx/xxx.git) and I guess this fails in the container since it doesn't have access to my username/password/ssh-keys.

But it's probably also something missing in the code that doesn't close the connection after the update-check, since I've seen a few git-http zombies too.

Would also be great to have a blacklist for modules not to update-check or at least a flag to only update check for the main project.

buxxi:
Changes that hopefully fixes #1907.

- Refactored the module to use async/await to make the error handling easier.
- Upgraded simple-git to use the timeout settings.
- Added timeout as config-property to the module with the default of 10 seconds (per module check)
- Changed the notification parameter to be its own object and not piggybacking on the objects returned from simple-git.

I have tested it by:
- hard resetting a modules git to a previous version
- dropping packages with iptables to the target address to force a timeout


buxxi:
Uhm, can't find a test case for the updatenotification-module and not really sure how it would be written either.

Is it just that I increased the total lines which decreased the overall coverage?

modules/default/updatenotification/node_helper.js
```diff
@@ -93,13 +93,17 @@ module.exports = NodeHelper.create({
 			delay = 60 * 1000;
 		}
 
-		var self = this;
+		let self = this;
```
rejas, line 96: Can we get rid of this self?

buxxi: I'm not sure, as a Java developer I always hated that `this` is mutable in JavaScript which makes me do this overly defensive :)

Could change it if you think it's safe.

MichMich:
Would love to merge this, but there are some conflicts. Can you fix these?

rejas:
> Is it just that I increased the total lines which decreased the overall coverage?

Yes, thats the cause, no worries.

modules/default/updatenotification/node_helper.js
rejas: One has to check always if it doesnt have side effects. But you can leave it for now and I might check it out later in another PR

Issue #2471, opened 2021-03-01
The Docker image project is currently unmaintained


Legion2:
The docker image project is now over 6 months unmaintained, see https://github.com/bastilimbach/docker-MagicMirror/issues/48. @sthuber90, @khassel and I want to discuss the future of the docker image project and find new maintainers for the project. We think here (github issue) is the best place to discuss this.

khassel:
Its really strange, the owner did some merges concerning the docs, but is ignoring useful PR's since July.

MichMich:
Not sure if you already opened a forum topic for this, but if you want I can make it sticky.

khassel:
It was my idea to discuss this here, because the most technical contributors are here, not in the forum. There are several ideas to solve this problem e.g. forking the current repo or using my docker setup, but before we continue to discuss this there should be a decision from @MichMich to abandon the current one  - or if not, how to continue.

buxxi:
Couldn't we just update the docs to add the other docker image by @khassel too?
Since the documentation says this:
"The following installation methods are not maintained by the MagicMirror² core team. Use these scripts and methods at your own risk."

What are the differences between the one by @khassel and @bastilimbach?

khassel:
> What are the differences between the one by @khassel and @bastilimbach?

The images of @bastilimbach are only for the `server only` setup. Years ago I asked him to support running on raspi with electron too, but he declined. So I decided to do this on my own.

Comparison (size = compressed size on docker hub):

IMAGE     | TAG                | OS/ARCH     | RASPI WITH ELECTRON | SIZE IN MB | DESCRIPTION
------------ | ------------------ | ----------- | -------- | ------ | -------------------------------------------
bastilimbach/docker-magicmirror | latest (or v2.x.y) | linux/amd64 | no       | 499 | only `serveronly`-mode, based on debian buster
bastilimbach/docker-magicmirror | latest (or v2.x.y) | linux/arm | no       | 447 | only `serveronly`-mode, based on debian buster
bastilimbach/docker-magicmirror | latest (or v2.x.y) | linux/arm64 | no       | 489 | only `serveronly`-mode, based on debian buster
bastilimbach/docker-magicmirror | develop | linux/amd64 | no       | 509 | only `serveronly`-mode, based on debian buster
bastilimbach/docker-magicmirror | develop | linux/arm | no       | 455 | only `serveronly`-mode, based on debian buster
[...]

PR #2480, opened 2021-03-13
added possibility to move 3rd-party-modules into another directory

khassel:
Inspired by the discussion with @buxxi [here](https://github.com/MichMich/MagicMirror/issues/2471#issuecomment-798600037) this PR adds the possiblity to use another directory than `modules` as parent directory for 3rd-party-modules.

This change simplifies every docker setup because we can now map only the 3rd-party-modules to the host (at the moment the default modules are always included). The new folder should be a subdirectory of the magic mirror main directory.

I used the existing `paths.modules` variable and did minimal code changes.

I see only downsides for modules who are working as module manager for magicmirror:
- MMM-Remote-Control: Minimal backward compatibel code change, I would make a PR in this repo
- mmpm: Will inform the author

Will set this PR to draft and wait for reactions before merging.

### DIFF
--- a/js/loader.js
+++ b/js/loader.js
@@ -91,7 +91,7 @@ var Loader = (function () {
 			var moduleFolder = config.paths.modules + "/" + module;
 
 			if (defaultModules.indexOf(moduleName) !== -1) {
-				moduleFolder = config.paths.modules + "/default/" + module;
+				moduleFolder = "modules/default/" + module;
 			}
 
 			if (moduleData.disabled === true) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ _This release is scheduled to be released on 2021-04-01._
 - Added locale to sample config file
 - Added support for self-signed certificates for the default calendar module (#466)
 - Added hiddenOnStartup flag to module config (#2475)
+- Added possibility to move 3rd-party-modules into another parent directory than `modules` (should be a directory beside `modules`)
 
 ### Updated
 
--- a/js/app.js
+++ b/js/app.js
@@ -106,7 +106,7 @@ function App() {
 	function loadModule(module, callback) {
 		const elements = module.split("/");
 		const moduleName = elements[elements.length - 1];
-		let moduleFolder = `${__dirname}/../modules/${module}`;
+		let moduleFolder = `${__dirname}/../${config.paths.modules}/${module}`;
 
 		if (defaultModules.includes(moduleName)) {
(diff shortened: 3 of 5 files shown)


js/loader.js
```diff
@@ -91,7 +91,7 @@ var Loader = (function () {
 			var moduleFolder = config.paths.modules + "/" + module;
 
 			if (defaultModules.indexOf(moduleName) !== -1) {
-				moduleFolder = config.paths.modules + "/default/" + module;
```
rejas, line 94: is that path already configurable?

rejas: if so, why isnt it in the config sample mentioned?

rejas: (also: maybe we should add that there?)

khassel: path is already configurable, see defaults.js and I tested this. Found nothing documented about this, can add it to the config sample but I dont't know if this a good idea when users who don't know what they are doing are playing with this variable ...

rejas: I see. Maybe @MichMich should weigh in on this: if the ALL options should be documented in the config sample?

sdetweil:
be careful moving modules to another path. quite a few will break as they have hard coded modules  in the path to their files.

rejas:
I assume he will keep the current path so that shouldnt be a problem right?

khassel:
I will stop this PR.

Looked at several modules and I see no chance to get it working due to hardcoded things in the modules.

This PR is not the problem, it is backward compatibel if you don't change the modules folder in the config.

But if I would use another folder e.g. in the docker setup, several modules won't work out of the box. So I will stay with the current solution (copying default modules).

If there will be ever a release 3.0 with breaking changes we should discuss to separate default and foreign modules by separate parent folders.

khassel:
I reopened this due to the hint of @MichMich [here](https://github.com/MichMich/MagicMirror/issues/2471#issuecomment-802992169).

> be careful moving modules to another path. quite a few will break as they have hard coded modules in the path to their files.

@sdetweil can you may give some examples?
I had problems with MMM-Remote-Control and MMM-Strava but they are working now.

> I assume he will keep the current path so that shouldnt be a problem right?

@rejas you mean url path or folder path? See changes in `js/server.js`.

The changes in this PR should not affect current setups. Its only the basis for being able to move 3rd-party-modules into another folder.

MichMich:
I'm closing this PR due to inactivity. Feel free to reopen whenever you think it's ready! Thanks!

PR #2481, opened 2021-03-14
Fix basic auth in calendar module

rejas:
After the latest round of merges I updated my MM to the latest develop branch. Turned out that the calendar wasnt loading anything anymore.

The authentication failed due to the base64 encoding of user/password not being accepted.

This PR fixes it by using a different base64 encoding method that is recommended for node. Or where did you get yours from @khassel ?

So no CHANGELOG entry since it fixes something that was just introduced. Also downgraded husky which was accidently upgraded to the next major version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _This release is scheduled to be released on 2021-04-01._
 - Update documentation and help screen about invalid config files.
 - Moving weather provider specific code and configuration into each provider and making hourly part of the interface.
 - Bump electron to v11 and enable contextIsolation.
-- Dont update the DOM when a module is not displayed.
+- Don't update the DOM when a module is not displayed.
 - Cleaned up jsdoc and tests.
 - Exposed logger as node module for easier access for 3rd party modules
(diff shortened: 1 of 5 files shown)


khassel:
see `const base64 = require("base-64");`, this line should be removed.

Why did the tests work with the old method?

MichMich:
Is this one ready to merge?

rejas:
I say yes. Checking on the tests I will do later

hangorazvan:
[2021-03-23 11:26:21.665] [ERROR] (node:3964) UnhandledPromiseRejectionWarning:
TypeError: this.broadcastEvents is not a function
    at E:\OneDrive\MagicMirror\modules\default\calendar\calendarfetcher.js:94:10
    at process._tickCallback (internal/process/next_tick.js:68:7)
[2021-03-23 11:26:21.710] [ERROR] (node:3964) UnhandledPromiseRejectionWarning:
Unhandled promise rejection. This error originated either by throwing inside of
an async function without a catch block, or by rejecting a promise which was not
 handled with .catch(). (rejection id: 8)

PR #2491, opened 2021-03-20
Translation to galician language

qu1que:
Hi, 

I've created a new translation json file. It's for galician language (in the northwest of Spain). 

Regards,

Quique


rejas (changes requested):
Thx but please base your PR against the develop branch. You will find that there is at least one more string to translate there too.

PR #2518, opened 2021-04-08
Hide newsfeed description

### LINKED ISSUE #2517: Hide newsfeed description setting ignored

I'm using the latest master version on a Raspberri Pi. After the upgrade the showDescription setting is ignored for the newsfeed module.

oemel09:
Fixes #2517

showDescription setting for newsfeed module is checked and applied again.

### DIFF
--- a/modules/default/newsfeed/newsfeed.js
+++ b/modules/default/newsfeed/newsfeed.js
@@ -135,7 +135,7 @@ Module.register("newsfeed", {
 			sourceTitle: item.sourceTitle,
 			publishDate: moment(new Date(item.pubdate)).fromNow(),
 			title: item.title,
-			description: item.description
+			description: this.isShowingDescription ? item.description : null
 		};
 	},
 
--- a/modules/default/newsfeed/newsfeed.js
+++ b/modules/default/newsfeed/newsfeed.js
@@ -135,7 +135,7 @@ Module.register("newsfeed", {
 			sourceTitle: item.sourceTitle,
 			publishDate: moment(new Date(item.pubdate)).fromNow(),
 			title: item.title,
-			description: this.isShowingDescription ? item.description : null
+			description: item.description
 		};
 	},
 
--- a/modules/default/newsfeed/newsfeed.njk
+++ b/modules/default/newsfeed/newsfeed.njk
@@ -13,7 +13,7 @@
         <div class="newsfeed-title bright medium light{{ ' no-wrap' if not config.wrapTitle }}">
             {{ title }}
         </div>
-        {% if (config.showDescription) %}
+        {% if config.showDescription %}
             <div class="newsfeed-desc small light{{ ' no-wrap' if not config.wrapDescription }}">
                 {% if config.truncDescription %}
                     {{ description | truncate(config.lengthDescription) }}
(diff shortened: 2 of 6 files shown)


rejas:
Looks good. Do you feel brave enough to add a test case for this too?

modules/default/newsfeed/newsfeed.js
```diff
@@ -135,7 +135,7 @@ Module.register("newsfeed", {
 			sourceTitle: item.sourceTitle,
 			publishDate: moment(new Date(item.pubdate)).fromNow(),
 			title: item.title,
-			description: item.description
+			description: this.isShowingDescription ? item.description : null
```
rejas, line 138: On second though: Shouldnt this be rather checked in the njk-template like the other options too?

rejas (changes requested):

modules/default/newsfeed/newsfeed.js
MichMich: Agree with @rejas.

oemel09: Changed it, definitely better. Didn't notice the config being passed into the template as well...

tests/e2e/modules/newsfeed_spec.js
```diff
@@ -36,6 +36,10 @@ describe("Newsfeed module", function () {
 		it("should show the newsfeed article", function () {
 			return app.client.waitUntilTextExists(".newsfeed .newsfeed-title", "QPanel", 10000);
 		});
+
+		it("should show the newsfeed description", function () {
+			return app.client.waitUntilTextExists(".newsfeed .newsfeed-desc", "Para instalar esta nueva versión", 10000);
+		});
```
rejas, line 42: And maybe a test where the description is turned off *prettyplease* :-)

modules/default/newsfeed/newsfeed.njk
```diff
@@ -13,13 +13,15 @@
...
             {{ title }}
         </div>
-        <div class="newsfeed-desc small light{{ ' no-wrap' if not config.wrapDescription }}">
-            {% if config.truncDescription %}
-                {{ description | truncate(config.lengthDescription) }}
-            {% else %}
-                {{ description }}
-            {% endif %}
-        </div>
+        {% if (config.showDescription) %}
```
rejas, line 16: I am wondering whats occuring more often "if (<condition>)" or "if <condition>" ?

MichMich: I'd say we strip out all unnecessary chars. But that isn't specifically necessary for this PR.

tests/e2e/modules/newsfeed_spec.js
oemel09: Done. Although I'm not sure if this is the right way to test for the absence of a text

rejas: Mhhh, maybe wait till the title exists and then check if the description is there? Might also be nice to do in the other test

oemel09: I'm not too familiar with these tests. I tried but couldn't get that to work...

rejas:
I think the test should look like this:


```
it("should NOT show the newsfeed description", async () => {
			await app.client.waitUntilTextExists(".newsfeed .newsfeed-title", "QPanel", 10000);
			const events = await app.client.$$(".newsfeed .newsfeed-desc");
			return expect(events.length).equals(0);
		});
```

rejas (approved):

rejas:
Nice! thanks again for the PR!

MichMich:
Great PR. Thanks!

PR #2532, opened 2021-04-10
Alert Module not recognizing multiple alerts

### LINKED ISSUE #2522: [Alert Module] Multiple Alerts doesn't show as expected

**Platform**: Hyper-V Ubuntu

**Node Version**: v14.15.5

**MagicMirror Version**: 2.15.0

**Description**: According to issue https://github.com/Jopyth/MMM-Remote-Control/issues/256, when sending multiple alerts, the alert module can't show every alert and just hide both.

**Steps to Reproduce**: Send one alert (I tried with a 60 seconds timer), and then send the same alert.

**Expected Results**: 

**Actual Results**: 

**Configuration**: 
```
/*************** AUTO GENERATED BY REMOTE CONTROL MODULE ***************/

var config = 
{
  address: '0.0.0.0',
  electronOptions: {
    fullscreen: false
  },
  ipWhitelist: [],
  modules: [
    {
      module: 'MMM-Remote-Control',
      position: 'top_right',
      header: '',
      config: {
        showModuleApiMenu: true,
        customMenu: 'custom.json',
        secureEndpoints: false,
        classes: {
          test: {
            toggle: 'calendar'
          },
          'This is a test': {
            hide: [
              'calendar',
              'newsfeed'
            ],
            toggle: 'clock'
          },
          'this is another test': {
            show: [
              'calendar',
              'newsfeed'
            ],
            hide: [
              'clock'
            ]
          }
        }
      }
    },
    {
      module: 'updatenotification',
      position: 'top_bar',
      config: {}
    },
    {
      module: 'clock',
      position: 'top_left',
      header: '',
      config: {}
    },
    {
      module: 'calendar',
      header: 'US Holidays',
      position: 'top_left',
      config: {
        calendars: [
          {
            symbol: 'calendar-check',
            url: 'webcal://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics'
          }
        ]
      }
    },
    {
      module: 'compliments',
      position: 'lower_third',
      classes: 'test',
      config: {}
    },
    {
      module: 'newsfeed',
      position: 'bottom_bar',
      disabled: true, // Here it's the disabled module
      config: {
        feeds: [
          {
            title: 'New York Times',
            url: 'https://rss.nytimes.com/services/xml/rss/nyt/HomePage.xml'
          }
        ]
      }
    },
    {
      module: 'alert',
      header: '',
      config: {
        welcome_message: true
      }
    }
  ],
  useHttps: false,
  httpsPrivateKey: '',
  httpsCertificate: '',
  logLevel: [
    'INFO',
    'LOG',
    'WARN',
    'ERROR'
  ]
}

/*************** DO NOT EDIT THE LINE BELOW ***************/
if (typeof module !== 'undefined') {module.exports = config;}
```

**Additional Notes**: I made a fast hotfix, you could replace some lines inside https://github.com/MichMich/MagicMirror/blob/491f5aa7769a12de7f20f201de0078704c158a8d/modules/default/alert/notificationFx.js#L130

You'll have the following:
```
setTimeout(() => {
    this.ntf.classList.add("ns-hide");

    //callback
    this.options.onClose();
}, 25);
```

Select all that, and replace it with:
```
this.ntf.classList.add("ns-hide");
```

Of course that solves the issue. I tried to take the "ns-hide" part out of the function, but onClose seems to deny the normal function of the module.
I know that onClose shouldn't be dismissed, but right now it's the only way I found to bypass this bug.

Let me know if you need something else!
Cheers :D

ezeholz:
Fixes #2522

I added a condition, everything should work out normally.

onClose seems to have issues when is used with a timeout. The best way to hotfix this it's just not calling it inside the timeout.

### DIFF
--- a/modules/default/alert/alert.js
+++ b/modules/default/alert/alert.js
@@ -79,7 +79,7 @@ Module.register("alert", {
 
 		//If module already has an open alert close it
 		if (this.alerts[sender.name]) {
-			this.hide_alert(sender,false);
+			this.hide_alert(sender, false);
 		}
 
 		//Display title and message only if they are provided in notification parameters
@@ -114,7 +114,7 @@ Module.register("alert", {
 			}, params.timer);
 		}
 	},
-	hide_alert: function (sender,close=true) {
+	hide_alert: function (sender, close = true) {
 		//Dismiss alert and remove from this.alerts
 		if (this.alerts[sender.name]) {
 			this.alerts[sender.name].dismiss(close);
--- a/modules/default/alert/alert.js
+++ b/modules/default/alert/alert.js
@@ -79,7 +79,7 @@ Module.register("alert", {
 
 		//If module already has an open alert close it
 		if (this.alerts[sender.name]) {
-			this.hide_alert(sender);
+			this.hide_alert(sender,false);
 		}
 
 		//Display title and message only if they are provided in notification parameters
@@ -114,10 +114,10 @@ Module.register("alert", {
 			}, params.timer);
 		}
 	},
-	hide_alert: function (sender) {
+	hide_alert: function (sender,close=true) {
 		//Dismiss alert and remove from this.alerts
 		if (this.alerts[sender.name]) {
-			this.alerts[sender.name].dismiss();
+			this.alerts[sender.name].dismiss(close);
 			this.alerts[sender.name] = null;
 			//Remove overlay
 			const overlay = document.getElementById("overlay");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ _This release is scheduled to be released on 2021-04-01._
 - Fix calendar start function logging inconsistency.
 - Fix updatenotification start function logging inconsistency.
 - Checks and applies the showDescription setting for the newsfeed module again
+- Fix alert not recognizing multiple alerts (#2522)
 
 ## [2.15.0] - 2021-04-01
 
--- a/modules/default/alert/notificationFx.js
+++ b/modules/default/alert/notificationFx.js
@@ -123,7 +123,7 @@
 	/**
 	 * Dismiss the notification
 	 */
-	NotificationFx.prototype.dismiss = function () {
+	NotificationFx.prototype.dismiss = function (close=true) {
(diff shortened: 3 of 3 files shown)


rejas:
Thanks for the PR. Would you be so brave and write a test for the alert module (you would be the first if I see that correctly :-)

ezeholz:
Sure, no problem. How can I do that? @rejas

rejas:
I'd wait for this https://github.com/MichMich/MagicMirror/pull/2533 to be merged first, I added a basic test for the alert in there. You can look at it to get a feeling on how to write tests.
So maybe after that is merged we can push more tests.

modules/default/alert/alert.js
```diff
@@ -79,7 +79,7 @@ Module.register("alert", {
 
 		//If module already has an open alert close it
 		if (this.alerts[sender.name]) {
-			this.hide_alert(sender);
+			this.hide_alert(sender, false);
```
rejas, line 82: that would now behave differently to what the comment line above says? if so, wouldnt that mean we have change in behaviour for the alert module (which would surpise other users)?

ezeholz: It's not so different, actually. Still closing the previous notification, the only change is that doesn't send the close action twice.
onClose function does the same as hiding it. And it's called as a delayed callback when the next notification take place, that's hiding not only the previous one, but the new one as well.

I would like to iterate above the alert module code when I get some time, because there are some functions that doesn't do much. Right now, this is more like a hotfix than an actual fix to the problem.

rejas: Good to hear that you want to clean it up :-)

rejas: Maybe add some explanation to the code would be nice for this PR too. But I am not opposed on merging it if it helps fix your bugs and doesnt break any exisiting behaviour

rejas (approved):

MichMich:
@ezeholz @rejas, It looks like this one is good to go? Merge?

ezeholz:
Sure! Go ahead, thanks for the vote of confidence. @rejas @MichMich :D

Issue #2530, opened 2021-04-10
[Weather Module] decimalSymbol has no effect


KristjanESPERANTO:
I set the `decimalSymbol` in the config of the new weather module to `,`, but that has no effect. The values still have points. I saw that the PR #1510 should have fix that, but it doesn't work (anymore?).

I noticed this with a just freshly installed MagicMirror instance.

Or am I doing something wrong in my config?

```
    {
      module: "weather",
      position: "top_right",
      header: "Wetter in ?????",
      config: {
        appendLocationNameToHeader: false,
        decimalSymbol: ",",
        weatherProvider: "openweathermap",
        type: "forecast",
        location: "?????????",
        locationID: "?????????",
        maxNumberOfDays: 5,
        apiKey: "?????????",
      }
    }
```

rejas:
Hi @KristjanESPERANTO I dont think you are doing somethign wrong. I took a quick look at the code and it seems to me the decimal point isnt used (yet) in the forecast part (only in the current weather part) of the module.

Maybe you want to write a fix for it? ;-)

rejas:
Nevermind, I think I can fix that if you give me a day :-)

KristjanESPERANTO:
I'm on it :slightly_smiling_face:

rejas:
Nice. So lets see if your PR fixes my tests :-D

rejas:
Did you have the time to test the latest develop branch to see if it fixes your problem for good now? So that we can close this ticekt?

KristjanESPERANTO:
Yes, with the develop branch it works fine. Thank you :slightly_smiling_face:

PR #2534, opened 2021-04-11
Fix decimalSymbol for forecast

KristjanESPERANTO:
With this, the `decimalSymbol` config also works at the forecast part of the new weather module.

This solves the issue #2530.

### DIFF
--- a/modules/default/weather/forecast.njk
+++ b/modules/default/weather/forecast.njk
@@ -14,10 +14,10 @@
                 {% endif %}
                 <td class="bright weather-icon"><span class="wi weathericon wi-{{ f.weatherType }}"></span></td>
                 <td class="align-right bright max-temp">
-                    {{ f.maxTemperature | roundValue | decimalSymbol | unit("temperature") }}
+                    {{ f.maxTemperature | roundValue | unit("temperature") | decimalSymbol }}
                 </td>
                 <td class="align-right min-temp">
-                    {{ f.minTemperature | roundValue | decimalSymbol | unit("temperature") }}
+                    {{ f.minTemperature | roundValue | unit("temperature") | decimalSymbol }}
                 </td>
                 {% if config.showPrecipitationAmount %}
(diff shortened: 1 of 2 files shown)


modules/default/weather/forecast.njk
```diff
@@ -14,10 +14,10 @@
                 {% endif %}
                 <td class="bright weather-icon"><span class="wi weathericon wi-{{ f.weatherType }}"></span></td>
                 <td class="align-right bright max-temp">
-                    {{ f.maxTemperature | roundValue | unit("temperature") }}
+                    {{ f.maxTemperature | roundValue | decimalSymbol | unit("temperature") }}
```
rejas, line 17: Not sure if the order is important here... In the current njk the decimalSymbol is applied last...

KristjanESPERANTO: I tried that first, but it didn't work.

KristjanESPERANTO: I checked it again. In the forcast part it doesn't work if `decimalSymbol` is at the end.

rejas: How doesnt it work on your side? It looks ok if I move the dcimalSymbol to the end.

KristjanESPERANTO: Apparently I made the same mistake twice (insert the decimalSymbol inside the unit brackets). That couldn't work 🤦‍♂️. So it (decimalSymbol at the end) works if you do it right. Sorry.

Should I adjust my PR so that it is consistent with the current part?

rejas: Yes, I think consistency is king :-) Have a good monday!

KristjanESPERANTO: I changed it :slightly_smiling_face:

rejas (approved):
Looks good, thx for the quick PR

PR #2535, opened 2021-04-11
Fix weather tests and add test for Issue 2530

rejas:
While looking at #2530 I noticed that some async calls in the weather-module tests were not used correctly so the tests would ALWAYS pass :-)

This PR fixes that and also adds a test for #2530 which will fail until #2534 is merged.

So I propose to merge this first and then see if #2534 really fixes it :-)


MichMich:
Tests seem to be failing.

rejas:
It is supposed to since it tests for the thing that got fixed with #2534 
Will merge develop into this branch, then it should run fine

rejas:
Told ya ;-)

PR #2536, opened 2021-04-13
Develop

B1gG:
> 1) Base your pull requests against the `develop` branch.
yep

> 2) Include these infos in the description:
This proposal is to rise the format in `dateFormat` when using `timeFormat` relative.
I started using the module, but I couldn't find the way to change the format on the dates from American style MM/DD/YYYY to anything else unless you use `absolute` as `timeFormat`

> 3) Please run `npm run lint:prettier` before submitting so that
> style issues are fixed.
done this time ;)

> 4) Don't forget to add an entry about your changes to
done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _This release is scheduled to be released on 2021-04-01._
 - Added French translations for "MODULE_CONFIG_ERROR" and "PRECIP".
 - Added German translation for "PRECIP".
 - Added first test for Alert module.
+- Added support for `dateFormat` when not using `timeFormat: "absolute"`
 
 ### Updated
 
--- a/fonts/package-lock.json
+++ b/fonts/package-lock.json
@@ -1,12 +1,26 @@
 {
-  "name": "magicmirror-fonts",
-  "requires": true,
-  "lockfileVersion": 1,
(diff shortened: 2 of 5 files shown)


MichMich:
@sdetweil since you put a lot of work in the calendar module. What's your view on this PR?

sdetweil:
I've seen quite a number of users struggling to get the dates displayed how they want. I havent done much with the UI, but never understood why this was not possible already. I agree with this change

MichMich:
Thanks for your PR @B1gG 
Thanks for your feedback @sdetweil!

B1gG:
Thanks @MichMich @sdetweil, will also. Update the docs later today

rejas:
You seemed to have added the lock files that were generated with a newer version of npm / node. This might lead to confusion for usersif they see the warning on the console. Could we revert those files?

MichMich:
@rejas I always generate e new lock file upon release. So I don't think this is an issue.

B1gG:
I saw that but thought you have tweak the gitignore if you don't want those. Perhaps is another little thing to do. 🙂

rejas:
No, I would leave them in the repository, they are valuable for making sure that the dependencies are the same everywhere.

PR #2544, opened 2021-04-18
Convert code to look a little more like es6 

rejas:
This PR
- converts 99% of the var-uages into let/const
- removes `var self=this`code 
- cleans up for-loops

so that my inner es6 fanatic is a little more satisfied

The remaining var usages would require more rework since they contain variables that are passed in the global scope (like module and translation)

Also, the deprecated weather modules were not converted, sometimes I am lazy too :-)


rejas:
That was quick :-D

Issue #565, opened 2016-12-12
Broken, unmaintained upstream "rrule" fails parsing calendar EXDATEs and UNTIL correctly


berlincount:
MagicMitch² develop, Electron, Node 6.9.1, Ubuntu

When having a recurring Google Calendar event, e.g. 3 times daily repeat:

BEGIN:VEVENT
DTSTART;TZID=Europe/Berlin:20161213T060000
DTEND;TZID=Europe/Berlin:20161213T070000
RRULE:FREQ=DAILY;COUNT=3
DTSTAMP:20161212T231711Z
UID:0sjm1ra23qcop8taelg1fttu1k@google.com
CREATED:20161212T231602Z
DESCRIPTION:
LAST-MODIFIED:20161212T231624Z
LOCATION:
SEQUENCE:1
STATUS:CONFIRMED
SUMMARY:Testrepeated
TRANSP:OPAQUE
END:VEVENT

Everything works as expected, i.e. three events show up correctly.

When deleting only middle occurance:

BEGIN:VEVENT
DTSTART;TZID=Europe/Berlin:20161213T060000
DTEND;TZID=Europe/Berlin:20161213T070000
RRULE:FREQ=DAILY;COUNT=3
EXDATE;TZID=Europe/Berlin:20161214T060000
DTSTAMP:20161212T231911Z
UID:0sjm1ra23qcop8taelg1fttu1k@google.com
CREATED:20161212T231602Z
DESCRIPTION:
LAST-MODIFIED:20161212T231624Z
LOCATION:
SEQUENCE:1
STATUS:CONFIRMED
SUMMARY:Testrepeated
TRANSP:OPAQUE
END:VEVENT

The removed event described by the EXDATE doesn't disappear correctly.

There is an additional issue that "UNTIL" statements in the RRULE doesn't seem to be honored correctly either, i.e. if there's a recurring calendar item that was defined three years ago and I skip an occurance next week it'll still be shown by the calendar module.

The functionality seems to be provided by https://github.com/jkbrzt/rrule, which is passed through by the vendored ical.js, and used by calling rrule.between().

The rrule package doesn't seem to be maintained anymore https://github.com/jkbrzt/rrule/issues/160, and the bug is present in both the latest release and the master.

Not sure whether we should fork & fix, or correct the data we're getting from rrule.between instead.

Ideas?

berlincount:
[...]

PR #2561, opened 2021-05-04
update dependencies

khassel:
- update dependencies in package.json
- require node >= v12 in package.json
- remove rrule-alt in package.json (no require found)
- update next release date in changelog

Tested this with node v16, we will see ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [2.16.0] - Unreleased (Develop Branch)
 
-_This release is scheduled to be released on 2021-04-01._
+_This release is scheduled to be released on 2021-07-01._
 
 Special thanks to the following contributors: @B1gG, @codac, @ezeholz, @khassel, @KristjanESPERANTO, @rejas, Faizan Ahmed.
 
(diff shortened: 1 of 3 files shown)


rejas:
Nice, the tests run through...

Regarding rrule-alt, I found this in the v2.1.1 changelog entry

- Switched out `rrule` package for `rrule-alt` and fixes in `ical.js` in order to fix calendar issues. ([#565](https://github.com/MichMich/MagicMirror/issues/565))

Not sure if that is still relevant?

rejas:
Do we need rrule at all? Cant find a require for that neither...

khassel:
> Switched out rrule package for rrule-alt and fixes in ical.js in order to fix calendar issues. (#565)

thats what i found, but nothing in the code.

> Do we need rrule at all?

Didn't search for `rrule` but there is no require in the code. It is used by `node-ical` so may it was used a long time ago.
So should I update this PR and remove `rrule`?

Or may @sdetweil can tell us more?

rejas:
Looking at the code I think it is a left over entry from the time when ical was in the code, not as a library but commited directly into the code. Back then the dependency to rrule was needed (and later replaced with rrule-alt)
So I say, remove it, I dont think anything will break :-D famous last words: What could possibly go wrong? ;-)

PR #2563, opened 2021-05-09
Add warning message

Maxbever:
Hello,
I have added a warning message if the height or the width of a div who contains modules is greater than the screen size. The message appears if a part of a module is out of the screen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Special thanks to the following contributors: @B1gG, @codac, @ezeholz, @khassel,
 - Added support for `dateFormat` when not using `timeFormat: "absolute"`
 - Added custom-properties in css of core and modules for improved styling experience, see `custom.css.sample` file
 - Added test case for recurring calendar events
+- Added warning message if the size of the modules overflows the size of the screen
 
 ### Updated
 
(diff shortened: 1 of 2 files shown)


sdetweil:
you should check if overflow is hidden
main.css
```
html {
  cursor: none;
  overflow: hidden;
  background: #000;
  user-select: none;
}
```

one of my modules (myCovid19) I can draw many charts, too many to be on one page.. I turn off hidden and I can scroll to the other charts.
its intentional to be beyond the fold..

rejas:
If I remember correctly, the newsfeed also overflows when set to fullarticle....

Maxbever:
My goal was more to display a warning when I set up the mirror and I can't scroll when the mirror is installed

MichMich:
This could cause a lot of log entries. And I don't really see the added benefit of hevig this in the log. Could you give more info about the use case?

Maxbever:
When I was adding different modules to my mirror, I try the MMM-soccer module. I wanted the display the module on the lower_third position, but there wasn't enough place to display all the teams. I don't understood why so I tried different config. The problem came because I put other modules in that position. So, I decided to create a warning when that problem happens.

MichMich:
I'm going to close this issue because it creates a bigger code burden than the issue that it solves. Thank for your time.

PR #2567, opened 2021-05-19
fix husky setup in prod environment

khassel:
After my husky upgrade yesterday my docker build is broken.

I'm using `npm install --only=prod` and in this context husky is not installed (which is o.k. as dev dependency) but the prepare script (containing `husky install`) is executed and fails.

They provide a [solution](https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker) for disabling husky in a ci environment but I think `npm install --only=prod` is a valid install also for other usecases.

So I chose the pragmatic way to run only `husky install` when the husky executable exists.

I'm open for better solutions.


sdetweil:
i use only=prod in the scripted install to reduce space consumption

khassel:
@rejas can you please take a look at this because you brought in the husky stuff, thanks!

rejas:
@khassel took a look and found some issues (not with your stuff but rather with husky and pretty-quick) Created a PR against yours so it can get merged together

rejas (approved):

khassel:
Thanks @rejas , so this is now ready for merge.

PR #2568, opened 2021-05-21
New 'Environment Canada' provider for default WEATHER module

Crazylegstoo:
A new provider has been created for the default WEATHER module to pull data from Environment Canada - pertains to Canadian locations only. This provides more accurate weather info for Canadian locations vs. the existing providers. This provider supports Current, Forecast, and Hourly data display.

There are no changes to the WEATHER module UI, with one exception: A new feature has been added to the module to allow a provider to _optionally_ specify a custom Precipitation unit-of-measure. In the current module, the Metric unit-of-measure for Precipitation is 'mm'. This does not really work well for Canada since we measure rain in 'mm' and snow in 'cm' - and this is also reflected in the data provided by Environment Canada. By allowing for a custom Precipitation unit-of-measure, the Environment Canada provider can ensure the correct data is displayed depending on the nature of the forecast precipitation. NOTE that if a provider does not specify a custom Precipitation unit-of-measure, the module will use current default rules to display precipitation.

The changes included in this PR:

weather/providers/**envcanada.js** - new Environment Canada provider

weather/**weatherobject.js** - add new element called PrecipitationUnits to hold an optional custom unit-of-measure

weather/**forecast.njk** - updated code that displays Preciptation to consider custom PrecipitationUnits value

weather/**hourly.njk** - updated code that displays Preciptation to consider custom PrecipitationUnits value

**CHANGELOG.md** - added bullet to highlight this additional provider

NOTE that WEATHER module documentation updates are still to come (to explain how to use the new provider).

### DIFF
--- a/modules/default/weather/forecast.njk
+++ b/modules/default/weather/forecast.njk
@@ -24,7 +24,7 @@
                         <td class="align-right bright precipitation">
                             {{ f.precipitation }}{{ f.precipitationUnits }}
                         </td>			
-		    {% else %}
+		                {% else %}
                         <td class="align-right bright precipitation">
                             {{ f.precipitation | unit("precip") }}
                         </td>
--- a/modules/default/weather/forecast.njk
+++ b/modules/default/weather/forecast.njk
@@ -14,17 +14,17 @@
                 {% endif %}
                 <td class="bright weather-icon"><span class="wi weathericon wi-{{ f.weatherType }}"></span></td>
                 <td class="align-right bright max-temp">
-                    {{ f.maxTemperature | roundValue | unit("temperature") }}
+                    {{ f.maxTemperature | roundValue | unit("temperature") | decimalSymbol }}
                 </td>
                 <td class="align-right min-temp">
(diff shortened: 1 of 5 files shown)


modules/default/weather/forecast.njk
```diff
@@ -20,9 +20,15 @@
...
                 </td>
                 {% if config.showPrecipitationAmount %}
-                    <td class="align-right bright precipitation">
-                        {{ f.precipitation | unit("precip") }}
-                    </td>
+                    {% if f.precipitationUnits %}	
+                        <td class="align-right bright precipitation">
+                            {{ f.precipitation }}{{ f.precipitationUnits }}
+                        </td>			
+		    {% else %}
```
rejas, line 27: Please check for intendation. Did you also run "npm run lint:prettier" before submitting?

Crazylegstoo: Thanks for catching this! I did run prettier but screwed up committing the updates back to my branch. This should be corrected now.

PR #2573, opened 2021-05-24
missing modules will be cloned (and installed) on start

khassel:
I thought about simplifying the module installation.

The only piece missing in `config.js` is the url of the git repo.

So what about introducing a new param `url` in the module section of the `config.js`, e.g.

```js
        {
                module: "MMM-RepoStats",
                position: "top_left",
                url: "https://gitlab.com/khassel/MMM-RepoStats.git",
                config: {
                        ...
                }
        },
```

If the `url` is set mm checks on start if the directory under `modules` exists and if not it will clone the repo. If there is a `package.json` after cloning it will also run `npm install` (or may better `npm install --only=prod`).

This is a small and backward compatibel change which would simplify module installation.

I provided this PR as draft to discuss if such a change is wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Special thanks to the following contributors: @B1gG, @codac, @ezeholz, @khassel,
 - Added custom-properties for colors and fonts for improved styling experience, see `custom.css.sample` file
 - Added custom-properties for gaps around body and between modules
 - Added test case for recurring calendar events
+- If (new) param `url` is provided in module config a missing module will be cloned (and installed) on start
 
 ### Updated
 
--- a/js/app.js
+++ b/js/app.js
@@ -14,6 +14,7 @@ const Log = require("logger");
 const Server = require(`${__dirname}/server`);
 const Utils = require(`${__dirname}/utils`);
 const defaultModules = require(`${__dirname}/../modules/default/defaultmodules`);
+const child_process = require("child_process");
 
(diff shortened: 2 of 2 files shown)


rejas:
The idea is nice, but I am wondering if the user should be informed about this and / or asked if he wants to have that done?

khassel:
> but I am wondering if the user should be informed about this and / or asked if he wants to have that done?

At the moment I see this more as an advanced feature for people who know what they are doing.

The clone/install only happens if you set the `url` in the `config.js`, so people doing this should know what will happen. I see no chance to ask because it happens on server side, you find only infos in the log.

My goal is to avoid custom scripting around mm when it is simple to have this as core feature.

With this feature you can start from scratch only with a predefined `config.js`.

It will not replace shortly the current module installation procedure because the old procedure is not only mentioned in the mm docs but in every module repository.

MichMich:
I appreciate the idea. But I forsee a lot of inexperienced users not understanding it (ie: pasting the wrong URL) creating a lot of support questions. 

It seems like we're trying to solve a problem which isn't there while generating a lot of code we need to maintain in the process. 

So not going to merge for now. Hope you understand.

MichMich:
By the way: this could simply be converted to a "3rd party" module that does the same. That way it doesn't need to be part of the core.

khassel:
still not convinced that this is no core topic but have it now running as module ...

PR #2574, opened 2021-05-26
Add flag that removes today entry in forecast

daniel-windsor:
Hello!

Just a wee addition to the weather module here.

I added a new config flag - `ignoreToday` - that removes the entry for today in weather forecast mode.  This is so if you have two weather modules, one just for today and one for the forecast, there isn't duplicate data between the two modules.

Thanks!


MichMich:
Thanks! Could you please also send a PR to the docs?

PR #2581, opened 2021-06-03
Fix to Environment Canada weather provider

Crazylegstoo:
After doing some further playing around with the new Env Canada provider for the weather module, I realized there were the following 2 bugs:

**hourly.njk** - logic for displaying custom vs. default precipitation units was incorrect and would have broken OpenWeatherMap module if 'hourly forecast' was being displayed with precipitation

**envcanada.js** - related to the previous point, an incorrect precipitation unit-of-measure was being set under the condition that (1) Env Canada provided an accumulation amount in a forecast and (2) module or MM was set to Imperial units

These bugs have been corrected and tested against the Env Canada provider and the OpenWeatherMap provider (both Imperial and Metric).


modules/default/weather/providers/envcanada.js
```diff
@@ -576,32 +587,6 @@ WeatherProvider.register("envcanada", {
 			return kilo;
 		}
 	},
-	//
-	// Convert cm or mm to inches
-	//
-	convertPrecipAmt(amt, units) {
```
rejas, line 582: Are those used nowhere?

rejas:
Thx for the PR. Could you add a CHANGELOG entry for it?

modules/default/weather/providers/envcanada.js
Crazylegstoo: That's correct. Conversion of precipitation amount and unit-of-measure was moved into the setPrecipitation function as part of cleaning up the bug I found. I think this is cleaner code since setPrecipitation already has the overall responsibility to determine how precipitation will be expressed based on the way the Env Canada service decides to provide the data - (can be an accumulation amount (mm or cm) and/or POP depending on their forecast).

Crazylegstoo:
CHANGELOG entry has been made (referencing custom precipitation unit-of-measure). I think I am getting a little bit tired of precipitation :)

rejas (approved):

MichMich:
There seems to be a conflict (which I can not fix from my phone.) Is it possible to fix this conflict so I can merge?

Crazylegstoo:
Apologies for this! I am camping for a few days with only my phone and crappy cell coverage. I will be back in civilsation on Thursday and will gets this addressed right away. Does that sound good to you?

MichMich:
No worries. I might be able to fix it before that time. No rush.

MichMich:
Merged! Thanks for your contribution!

PR #2586, opened 2021-06-10
Todoist module

Axenntio:
It is now possible to add a [Todoist](https://todoist.com/) module

Following the template in `config` folder, and providing your API key, you can view your today's task.


MichMich:
This doesn't need to be part of the core. This can just be a third party module. No need to make a or for this.

PR #2603, opened 2021-06-21
workaround for dev_console test using getWindowCount

khassel:
Found a workaround to get the dev_console test running again (but still not the real cause).

This needs an updaten in `electron.js` so I don't know if that really makes sense because with this change we are not testing the "originial" dev setup.

So this is only a draft, waiting for comments, no problem if we throw it away ...


rejas:
Can't really offer any advice on the whole jest thing so from my side I am okay with this PR.

MichMich:
Honestly, I don't know a lot about frontend testing either. So as long as this doesn't break automated testing, I'm ok with this.

Issue #2408, opened 2021-01-06
Improve test suite by starting electron only when needed


rejas:
Speaking of: the test suite could be so much improved. Currently for every test we start electron, while in most cases we could just "virtually" start the module and test what getDom returns.

_Originally posted by @MichMich in https://github.com/MichMich/MagicMirror/issues/2401#issuecomment-753959303_

MichMich:
And while were at it. It might be worth rethinking the test suite we are using. I'm not a specialist on (javascript) testing. But the output of our current test is rather underwhelming. I like the output of [Jest](https://jestjs.io). But maybe there is a better alternative.

fewieden:
I had to work with multiple test frameworks for javascript in my company such as mocha, ava, jest, ...

My favorite test framework of all (the ones I used) is definitely Jest.

I used it for the last few days to test one of my modules from 0 to 100% coverage [MMM-soccer](https://github.com/fewieden/MMM-soccer). Since this worked quite nicely I intend to apply this also to my other modules.

The tests are mainly unit tests with a few integration tests. I didn't create any e2e tests like the ones we have for the default modules with electron. Instead, I created simple mocks for NodeHelper and Module, to avoid cloning magic mirror in the CI like in this [example](https://github.com/ianperrin/MMM-Strava/blob/ef0855979c0870c01511e8041bbcd41ece66e101/.travis.yml#L8-L11).

My suggestion would be to cover most logic as unit and integration test, which doesn't require to spin up electron, but using mocks where needed. In addition, electron can be used for specific tests like rendering or related to electron itself.

Maybe @roramirez can add some thoughts to it, as I think he created the initial setup.

MichMich:
This sounds so much better than what we currently have. It would also simplify building test, especially if we create some test tools users can use.

roramirez:
Hi guys!

The main idea to create the e2e was try the cover all most cases of the principal modules and find and get it  the bug in configuration and start of the app in Electron. At in this time (4 year ago) the tests for new feature/bugfix are 0 and the way most easily to cover from the module to the display in the html render for Electron was build a e2e test. 

The unit test are good but we try get and test what the user can see in the Mirror. The problem here the issue is for run a heavy load and many tests in the e2e in CI. Issue like dependencies between run/start/stop of app in middle of every test case have a lot of cost of resources and sometime the stop or start not run the best way.

[...]

PR #2609, opened 2021-07-04
Electron13

khassel:
Managed to get the e2e tests running with electron v13 and spectron v15.

Tested new electron version on pi3 and ubuntu desktop.

Related to #2408.


MichMich:
Great work!

I notice that almost all changes are the same electron options object in all different tests. Maybe it's worth it to extract that to a separate file so it can be reused?

khassel:
> Maybe it's worth it to extract that to a separate file so it can be reused?

You mean something like

```js
	defaultElectronOptions: {
		webPreferences: {
			nodeIntegration: true,
			enableRemoteModule: true,
			contextIsolation: false
		}
```

in a seperate file and then 

```js
	electronOptions: defaultElectronOptions
```

in every test config?

MichMich:
Jean, or even better. A default test config that can be extended. 
   
    let config = testConfig({foo: bar})

The testConfig method will return the default config (including the electronOptions) and it will extend/overwrite with whatever you add as an argument. 

Of course this can be a separate PR. But it might save some work in the future and makes PR's like these much cleaner.

khassel:
o.k., will try this in a new PR.

rejas (approved):
LGTM!

PR #2611, opened 2021-07-05
refactor test configs

khassel:
see discussion [here](https://github.com/MichMich/MagicMirror/pull/2609#issuecomment-873647689).

I created a `default.js` test case which is now included in every other test case. So in these test cases only the properties are changed which are different to `default.js`.

Now you can see directly the differences to `default.js` and the defaults are in one place (e.g. electronOptions).


tests/configs/default.js
```diff
@@ -1,8 +1,3 @@
-/* Magic Mirror Test config sample environment
- *
- * By Rodrigo Ramírez Norambuena https://rodrigoramirez.com
- * MIT Licensed.
- */
```
rejas, line 5: I'd keep some kind of comment header so that people know what this default js does

khassel: The comments have been lost, will paste them back.

MichMich:
Thanks for your awesome effort! 

I think it might be better to go for a 'factory' approach. In this case you opt to use a lot of `Object.assign(require("../../default.js"), config);` statements. Using a factory approach, we could do something like:


````js
const configFactory = require('../../default.js')

const config = configFacory({
    modules: [
        ...
    ]
})
````

It is a bit more verbose, but a lot cleaner, and gives a lot of opportunities for extension and validation. 

What do you think?

**Comment by khassel** [comment]
will take a look (still thinking about how can I do this with search and replace ...)

**Comment by MichMich** [comment]
I understand. We can always keep this PR open for someone to continue (and do some manual labor). :)

**Comment by khassel** [comment]
something is missing here

```js
const configFactory = require('../../default.js')

const config = configFactory({
    modules: [
        ...
    ]
})
```
getting `Error found: TypeError: configFactory is not a function`

MichMich:
Ha, yeah sorry. That configFactory is an example. It's the function that we need to build. :)

khassel:
I'm stopping here. You can take a look at my `develop` branch where I [wrote the function](https://github.com/khassel/MagicMirror/blob/develop/tests/configs/default.js) but half of the tests are failing with this construction and I don't see why.

rejas:
Cool that you came back to it. I can report the tests work on my mac and ubuntu laptops with no problem now too.

MichMich:
Awesome work! Thanks!

PR #2613, opened 2021-07-14
Actually test all js and css files when lint script is run

rejas:
So I noticed that the notation in the package.json was inccorect therefor not all js and css files were linted and tested. 

This PR fixes that and also all errors that appeared.

Most noteworthy to look at is the conversion of
"var Module = "
in module.js and a similar conversion in translator.js


khassel (approved):

rejas:
Not sure why the coverage drops like this. Any idea @khassel ?

khassel:
No. But I will open a new issue concerning code cov later. I think @MichMich will merge it nevertheless.

MichMich:
Thanks!

Issue #2462, opened 2021-02-20
Add ability to hide time in clock module


prapicault:
Hi, I would like for the clock module to have the ability to hide the time (so it only displays the date). I have already changed the code in my installation but it makes keeping up with master a bit annoying. 
Would you be interested in a contribution that would add a `showTime` configuration option to the default clock module? This option when set to `false` would not render the time at all.

sdetweil:
you could use css to turn off the time by overriding the class set to disolay:none;

prapicault:
I probably could (though tbh, it did not even occur to be me try) but it does not seem to be as discoverable as a configuration for other users.

MichMich:
Feel free to send a PR. 👍🏻

rejas:
I looked into the clock module to see if hiding the time was possible.
Yes it is, but the module is loaded with different options for all kind of things which made me think if it could be reworked a little. But then one or two options might change so I wanted forst to ask @MichMich what the official stance on removing / chaning options in the default module is?

MichMich:
I don't think we should overbloat the clock module (I think it already is a bit). If someone really wants super specific features, it's better to opt for a third party module. 

I'm not a big fan for removing options and essentially introducing breaking changes.

rejas:
@prapicault @AndyHazz can this issue be closed since your problems should be fixed in the last year ;-) ? if yes @MichMich can do it

AndyHazz:
Fine by me, I've had no issues since switching to use showDate: false

PR #2627, opened 2021-08-07
Refactor clock module 

rejas:
Lots of code cleanup resulting in less lines :-)

- Added showTime parameter to clock module for enabling/disabling time display in analog clock
- analogShowDate is obsolete (should we deprecate it?) since showTime and analogPlacement does the job too
- Uses css grid for displaying the clocks
- Also fixes https://github.com/MichMich/MagicMirror/issues/2462

Normally I would have refactored more and removed a lot of parameters that could be replaced with more general ones but maybe thats for another day? Whats you opinions @khassel @MichMich ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2021-10-01._
 
 ### Added
 
+- Added showTime parameter to clock module for enabling/disabling time display in analog clock
+
 ### Updated
 
 - Bump electron to v13 (and spectron to v15) and update other dependencies in package.json.
--- a/modules/default/clock/clock.js
+++ b/modules/default/clock/clock.js
@@ -12,11 +12,14 @@ Module.register("clock", {
 		displayType: "digital", // options: digital, analog, both
 
 		timeFormat: config.timeFormat,
+		timezone: null,
+
 		displaySeconds: true,
 		showPeriod: true,
 		showPeriodUpper: false,
 		clockBold: false,
 		showDate: true,
+		showTime: false,
 		showWeek: false,
 		dateFormat: "dddd, LL",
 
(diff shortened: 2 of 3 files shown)


khassel:
LGTM.

> Normally I would have refactored more and removed a lot of parameters that could be replaced with more general ones but maybe thats for another day? Whats you opinions @khassel @MichMich ?

The problem is removing such paramters will generate support because the clock is not worling anymore as expected with the next release. On the other side the project must move on and get rid of old stuff. May it is sufficient to log a warning if the old parameter is used with a short message how to solve it (may with a link).

Another idea is to replace such old parameters automatically on start up with the new ones in the `config.js` (including saving the old config).

MichMich:
I think showing a warning (and alert on the UI?) is the best option. It is easy to implement and helps the user to understand things change.

rejas:
> Another idea is to replace such old parameters automatically on start up with the new ones in the `config.js` (including saving the old config).

That would be optimal but writing logic for replacing parameters with other would be very hard if it is more than just a simple "replace nameY with nameX"

Adding a UI warning and in the console would be a good start. So maybe have it deprecated for one or two releases and print a warning and then remove it after that?

rejas:
But of course any of that stuff would happen in another PR ;-)

MichMich:
Maybe the easiest way is to create a list of json paths that are deprecated and if one of those paths occur in the config it shows the alert to check the log, and fix the config. 

The alert/messages can be displayed as long as people have it in their config. I don't think it is a good idea to only do this for two releases since a lot of people only update one a year. 

Automatically change the config requires A LOT of code, generates a big risk of messing things up, and doesn't teach the user a thing. I don't think we should take this route, IMHO.

MichMich:
The json paths can be supplied by a method of the module. This way all modules can hook into this feature. A nice additional feature for the core.

khassel:
> Automatically change the config requires A LOT of code, generates a big risk of messing things up, and doesn't teach the user a thing. I don't think we should take this route, IMHO.

I'm aware that the automatically change of the config is a bigger thing and I won't code it ;)

So the json paths idea is my favorite at the moment.

rejas:
> Maybe the easiest way is to create a list of json paths that are deprecated and if one of those paths occur in the config it shows the alert to check the log, and fix the config

Sounds good, will you create an enchancment-ticket for this? So people with time can take a stab at it ;-)

> The alert/messages can be displayed as long as people have it in their config. I don't think it is a good idea to only do this for two releases since a lot of people only update one a year.

Well, one could always declare the spring release as the one that cleans stuff up and removes deprecated stuff ;-)
 
> Automatically change the config requires A LOT of code, generates a big risk of messing things up, and doesn't teach the user a thing. I don't think we should take this route, IMHO.

Agreed.

Issue #2630, opened 2021-08-10
weathergov forecast not loading


apiontek:
I found that regardless of config, the weathergov provider for the weather module would stall at "loading" for the `type: 'forecast'` weather config, although it worked fine for the `'current'` weather config.

After searching issues I couldn't see any mention of it nor a solution, so I went through the code.

The issue seems to lie in the `fetchWxGovURLs()` function. For weather.gov, the provider first needs to determine, from the API, what the correct API URLs are for various kinds of data. It waits until the required URLs are populated, and _then_ goes to fetch the actual weather data.

However, the `finally()` only calls `this.fetchCurrentWeather()` which does nothing if the user has configured the weather module to get the forecast.

I found that replacing the simple call to `this.fetchCurrentWeather()` as below fixes the issue for me:

```js
        // handle both 'current' and 'forecast' config types
        if (config.type == 'forecast') {
          this.fetchWeatherForecast();
        } else if (config.type == 'current') {
          this.fetchCurrentWeather();
        }
```

This seems good enough. The provider doesn't support the further types of 'hourly' or 'daily' as far as I can tell, and I don't need them.

Cosmetically, the default `apiBase` is also wrong in the module, which shouldn't matter as long as users correctly configure the apiBase in config.js, but it ought to be "https://api.weather.gov/points/" instead of "https://api.weatherbit.io/v2.0"

I've can do a pull request for both of the above, just want to post the issue first.

sdetweil:
change the initialLoadDelay config to 100?

the code does this, but timer of 0 is bad 

scheduleUpdate: function (delay = null) {
		let nextLoad = this.config.updateInterval;
		if (delay !== null && delay >= 0) {
			nextLoad = delay;
		}

I had opened issue back in may, 
https://github.com/MichMich/MagicMirror/issues/2505

apiontek:
Changing the load delays isn't necessary if the `fetchWxGovURLs()` method is fixed. It already properly calls `this.fetchCurrentWeather()` when it has the correct URLs. But if user config is for 'forecast' rather than 'current' it should call `this.fetchWeatherForecast()` instead, which the above modification handles.

sdetweil:
the problem is that the fetch and construction of the proper endpoint urls is variable length time.. 

SOMETIMEs your solution will work..  there is no callback to make it certain the have completed before attempting to use the constructed urls

apiontek:
[...]

PR #2631, opened 2021-08-10
update weathergov provider to try fetching not just current but also …

apiontek:
…forecast when API URLs available

Trying to address seeming shortcoming I reported in issue #2630 

It seems that configuring the "initialLoadDelay" to a high enough number also can solve the issue where API data might take a while, but the code already tries to handle this by executing current weather fetch once the API URLs are ready. Seems to make a lot of sense to me to try to load the forecast here instead if that's what user has configured.

default apiBase url was also incorrect.


rejas (approved):
Makes sense and looks good to me :-)

PR #2643, opened 2021-08-30
Add custom switches for electron mainWindow

eouia:
I need some custom electron switches for my purpose. (e.g., Desktop background playing)

This feature enables adding custom switches on Electron starts by a user.

You can add switches in `config.js` like;
```js
electronSwitches: ['enable-transparent-visuals'],
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ _This release is scheduled to be released on 2021-10-01._
 ### Added
 
 - Added showTime parameter to clock module for enabling/disabling time display in analog clock
+- Added custom electron switches from user config (`config.electronSwitches`)
 
 ### Updated
 
--- a/js/electron.js
+++ b/js/electron.js
@@ -19,7 +19,8 @@ let mainWindow;
  *
  */
 function createWindow() {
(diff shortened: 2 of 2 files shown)


MichMich:
What's the difference between this and the already implemented `electronOptions` config option?

eouia:
@MichMich
This is not about `electronOption`. It is about `switch`.
With `electronOption` (https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions), you can control the window created by electron, but not the electron app itself.
`switch (commandLine.switch)` is rather for App(and some node features) itself.  (https://www.electronjs.org/docs/latest/api/command-line-switches) These could not be achieved with `electronOption`

MichMich:
Ok. If you can fix the conflict so I can merge it. Also, please create a PR in the docs repo (and if possible reference the PR in this PR). 

Thanks!

eouia:
Thanks. I'll do it in a few hours.
Anyway, with this feature, your MM could be located as Desktop wallpaper on the Mac or Linux machine. (That's my purpose of this PR. :D )

MichMich:
Cool! Thanks for your PR!

PR #2648, opened 2021-09-01
Add common methods to weatherobject

rejas:
Looking at the different weather providers I noticed some methods were repeated.

So I refactored some of them into the weatherobject to be used from the providers.

Added a unit test too, so hopefully it is correct. But better to have more than one pair of eyes look at the changes since the providers are hard to test :-)

Question though: Some wind / temp conversions seem to be doubled in there too, should I take a look at them too?

### DIFF
--- a/modules/default/weather/weatherobject.js
+++ b/modules/default/weather/weatherobject.js
@@ -1,3 +1,5 @@
+/* global SunCalc */
+
 /* Magic Mirror
  * Module: Weather
  *
@@ -112,5 +114,17 @@ class WeatherObject {
 		}
 
 		return this.tempUnits === "imperial" ? feelsLike : ((feelsLike - 32) * 5) / 9;
+	}
+
+	/**
+	 * Update the sunrise / sunset time depending on the location
+	 *
+	 * @param {number} lat latitude
+	 * @param {number} lon longitude
+	 */
+	updateSunTime(lat, lon) {
+		let times = SunCalc.getTimes(new Date(), lat, lon);
+		this.sunrise = moment(times.sunrise, "X");
+		this.sunset = moment(times.sunset, "X");
 	}
 }
--- a/modules/default/weather/weatherobject.js
+++ b/modules/default/weather/weatherobject.js
@@ -13,11 +13,12 @@
  */
 class WeatherObject {
 	/**
+	 * Constructor for a WeatherObject
 	 *
-	 * @param units
-	 * @param tempUnits
-	 * @param windUnits
-	 * @param useKmh
+	 * @param {string} units what units to use, "imperial" or "metric"
+	 * @param {string} tempUnits what tempunits to use
+	 * @param {string} windUnits what windunits to use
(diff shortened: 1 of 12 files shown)


khassel:
test errors may caused by `node-fetch`: https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md
Will take a look at this.

modules/default/weather/weatherobject.js
```diff
@@ -10,6 +12,13 @@
  * As soon as we start implementing the forecast, mode properties will be added.
  */
 class WeatherObject {
+	/**
+	 *
+	 * @param units
```
khassel, line 17: what about all the following warnings `Missing JSDoc @param ...`?

rejas: yeah, will fix those with the next commits over the weekend. stay tuned :-)

MichMich:
Thanks! I love refactors like these!

khassel:
@rejas after this was merged I'm getting errors in one unit test.

I know these tests passed here but I'm running the tests additionally after building the docker images (here the develop image).

Tested now locally and with my `TZ=Europe/Berlin` everything works fine.

But if I set e.g. `TZ=America/New_York` the unit tests fail:

```bash
 FAIL   unit  tests/unit/functions/weather_object_spec.js
  WeatherObject
    ✕ should return true for daytime at noon (2 ms)
    ✕ should return false for daytime at midnight (1 ms)

  ● WeatherObject › should return true for daytime at noon

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      14 |              weatherobject.date = moment(12, "HH");
      15 |              weatherobject.updateSunTime(-6.774877582342688, 37.63345667023327);
    > 16 |              expect(weatherobject.isDayTime()).toBe(true);
         |                                                ^
      17 |      });
      18 |
      19 |      it("should return false for daytime at midnight", function () {

      at Object.<anonymous> (tests/unit/functions/weather_object_spec.js:16:37)

  ● WeatherObject › should return false for daytime at midnight

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

      20 |              weatherobject.date = moment(0, "HH");
      21 |              weatherobject.updateSunTime(-6.774877582342688, 37.63345667023327);
    > 22 |              expect(weatherobject.isDayTime()).toBe(false);
         |                                                ^
      23 |      });
      24 | });
      25 |

      at Object.<anonymous> (tests/unit/functions/weather_object_spec.js:22:37)
```

So it seems these tests are now depending on which timezone they are running.

rejas:
> @rejas after this was merged I'm getting errors in one unit test.

Yes, I can confirm that when changing my tz too. ALready have a fix for that, question is @MichMich, should I make a seperate PR for that fix or can I put it in https://github.com/MichMich/MagicMirror/pull/2650 ?

MichMich:
Just put it in the same. That's ok.

rejas:
Done :-)

Issue #2644, opened 2021-08-30
Update notifications not working as intended


Richard238:
From - https://forum.magicmirror.builders/topic/15425

`git remote -v`

origin https://github.com/MichMich/MagicMirror.git (fetch)
origin https://github.com/MichMich/MagicMirror.git (push)

config is set to get update notifications:

`{
 module: "updatenotification", position: "top_bar" 
}, `

I received no notification of the upgrade to 2.15, or to 2.16.

When I manually applied an update, I was some 4130 commits behind.

rejas:
A quick check seems to indicate that the updatenotification worked until 2.15.0 
With that release something broke so that only module updates are shown but not the core updates.

rejas:
Maybe @buxxi can take a look at it since he worked on that module during the developemnt cycle.

khassel:
This is caused by https://github.com/MichMich/MagicMirror/pull/2100.

Running with `--dry-run` the property `fetchData.behind` is always = 0.

Running without `--dry-run` works as expected.

So if there is no better idea we have to revert the above PR.

rejas:
So @bee-mar should weigh in on this topic :-)

buxxi:
I didn't receive a notification for the last release, have been thinking about trying to debug the cause but @khassel seems to have been quicker.

I have one module that hasn't been updated that I tested an alternative way to find out the amount of commits:
```sh
git fetch --dry-run
```
> From https://github.com/timdows/MMM-JsonTable
>   8059e8d..3051433  master     -> origin/master
```sh
git rev-list --ancestry-path 8059e8d..3051433 | wc -l
```
> 4

haven't checked how this would work in the module, but it doesn't actually update the local git with the remote. Wouldn't want every module to auto update when a commit is pushed to the repo :)

khassel:
This can be closed, fixed in new release `v2.17.1`.

PR #2651, opened 2021-09-04
fix update notification

khassel:
see #2644 

We have to handle several cases which I was not able to get running with the `simple-git` package. So now the git command line is used.

First run `git status -sb`.

If the repo was already fetched (hard, not with dry run) the output contains the `behind` number, so use this.

If not run `git fetch --dry-run` and extract the `from..to` hashes. With this run `git rev-list --ancestry-path --count from..to` to get the `behind` number.

### DIFF
--- a/modules/default/updatenotification/git_helper.js
+++ b/modules/default/updatenotification/git_helper.js
@@ -1,0 +1,132 @@
+const util = require("util");
+const exec = util.promisify(require("child_process").exec);
+const fs = require("fs");
+const path = require("path");
+const Log = require("logger");
+
+class gitHelper {
+	constructor(){
+		this.gitRepos = [];
+		this.baseDir = path.normalize(__dirname + "/../../../");
+	}
+
+	async execShell(command) {
+		let res = { stdout: "", stderr: "" };
+		const { stdout, stderr } = await exec(command);
+
+		res.stdout = stdout;
+		res.stderr = stderr;
+		return res;
+	}
+
+	async isGitRepo(moduleFolder) {
+		let res = await this.execShell("cd " + moduleFolder + " && git remote -v");
+		if (res.stderr) {
+			Log.error("Failed to fetch git data for " + moduleFolder + ": " + res.stderr);
+			return false;
+		} else {
+			return true;
+		}
+	}
+
+	add(moduleName) {
+		let moduleFolder = this.baseDir;
+		if (moduleName !== "default") {
+			moduleFolder = moduleFolder + 'modules/' + moduleName;
+		}
+		try {
+			Log.info("Checking git for module: " + moduleName);
+			// Throws error if file doesn't exist
+			fs.statSync(path.join(moduleFolder, ".git"));
+			// Fetch the git or throw error if no remotes
+			if (this.isGitRepo(moduleFolder)) {
+				// Folder has .git and has at least one git remote, watch this folder
+				this.gitRepos.unshift({ module: moduleName, folder: moduleFolder });
+			}
+		} catch (err) {
+			// Error when directory .git doesn't exist or doesn't have any remotes
+			// This module is not managed with git, skip
+		}
+
+	}
+
+	async getRepoInfo(repo) {
+		let gitInfo = {
+			module: repo.module,
+			// commits behind:
+			behind: 0,
+			// branch name:
+			current: "",
+			// current hash:
+			hash: "",
+			// remote branch:
+			tracking: ""
+		};
+		let res;
+		if (repo.module === "default") {
+			// the hash is only needed for the mm repo
+			res = await this.execShell("cd " + repo.folder + " && git rev-parse HEAD");
+			if (res.stderr) {
+				Log.error("Failed to get current commit hash for " + repo.module + ": " + res.stderr);
+			}
+			gitInfo.hash = res.stdout;
+		}
+		res = await this.execShell("cd " + repo.folder + " && git status -sb");
+		if (res.stderr) {
+			Log.error("Failed to get git status for " + repo.module + ": " + res.stderr);
+			// exit without git status info
+			return;
+		}
+		// only the first line of stdout is evaluated
+		let status = res.stdout.split("\n")[0];
+		// examples for status:
+		// ## develop...origin/develop
+		// ## master...origin/master [behind 8]
+		status = status.match(/(?![.#])([^.]*)/g);
+		// examples for status:
+		// [ ' develop', 'origin/develop', '' ]
+		// [ ' master', 'origin/master [behind 8]', '' ]
+		gitInfo.current = status[0].trim();
+		status = status[1].split(" ");
+		// examples for status:
+		// [ 'origin/develop' ]
+		// [ 'origin/master', '[behind', '8]' ]
+		gitInfo.tracking = status[0].trim();
+		if (status[2]) {
+			// git fetch was already called before so `git status -sb` delivers already the behind number
+			gitInfo.behind = parseInt(status[2].substring(0, status[2].length - 1));
+			return gitInfo;
+		}
+		res = await this.execShell("cd " + repo.folder + " && git fetch --dry-run");
+		// example output:
+		// From https://github.com/MichMich/MagicMirror
+		//    e40ddd4..06389e3  develop    -> origin/develop
+		// here the result is in stderr (this is a git default, don't ask why ...)
+		if (res.stderr === "") return;
+		let refs = res.stderr.match(/s*([a-z,0-9]+[.][.][a-z,0-9]+)s*/g)[0];
+		if (refs === "") {
+			// if match fails set behind to a number greater 0 and return
+			gitInfo.behind = 1;
+			return gitInfo;
+		}
+		// get behind with refs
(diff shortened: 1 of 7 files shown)


modules/default/updatenotification/node_helper.js
```diff
@@ -54,34 +56,78 @@ module.exports = NodeHelper.create({
...
+			Log.error("Failed to get current commit hash for " + repo.module + ": " + res.stderr);
+		}
+		gitInfo.hash = res.stdout;
+		res = await this.execShell("cd " + repo.folder + " && git status -sb");
+		if (res.stderr) {
+			Log.error("Failed to get git status for " + repo.module + ": " + res.stderr);
+			// exit without git status info
+			return;
+		}
+		// get branch and remote
```
rejas, line 101: would be a nice bonus in the commentsto see an example of what the res.stdout is (for better undertsanding what happens)

modules/default/updatenotification/node_helper.js
```diff
@@ -54,34 +56,78 @@ module.exports = NodeHelper.create({
...
+		gitInfo.tracking = status[0].trim();
+		if (status[2]) {
+			// git fetch was already called before so `git status -sb` delivers already the behind number
+			gitInfo.behind = parseInt(status[2].substring(0, status[2].length - 1));
+			return gitInfo;
+		}
+		res = await this.execShell("cd " + repo.folder + " && git fetch --dry-run");
+		// here the result is in stderr
+		if (res.stderr === "") return;
+		// set default > 0
```
rejas, line 115: default means what?

rejas (approved):

modules/default/updatenotification/node_helper.js
```diff
@@ -54,34 +56,78 @@ module.exports = NodeHelper.create({
...
-						behind: fetchData.behind,
-						current: fetchData.current,
-						hash: logData.latest.hash,
-						tracking: fetchData.tracking
-					});
-				}
-			} catch (err) {
-				Log.error("Failed to fetch git data for " + sg.module + ": " + err);
+		for (let repo of gitRepos) {
+			const gitInfo = await this.getRepoInfo(repo);
```
rejas, line 128: I'm wondering if that git status message can be mocked so that we could write some tests....

khassel: not sure what you are talking about here, the `git status -sb` output (see new comments) or the `gitInfo` object.

rejas: Sorry for my late night mumbling: Just ment to ask, if it is possible in your opinion to write unit tests for your new code?

rejas (approved):

modules/default/updatenotification/node_helper.js
khassel: will put it on my list (not in this PR).

khassel:
@rejas I moved the git stuff into an own class so I was able to write some minimal tests.

@MichMich we have a problem with the new node version v16.9 because this breaks the e2e tests. With this PR node is nailed to v16.8 (as hotfix).

So while this is not merged every e2e test from any PR will fail because of this problem.

As info the error message using node v16.9 (I have no idea what's going on):

```bash
#
# Fatal error in , line 0
# Check failed: !holder_map.has_named_interceptor().
#
#
#
#FailureMessage Object: 0x7ffd78d05d90
 1: 0xb691f1  [node]
 2: 0x1bf3094 V8_Fatal(char const*, ...) [node]
 3: 0x10ac4a1 v8::internal::ConcurrentLookupIterator::TryGetPropertyCell(v8::internal::Isolate*, v8::internal::LocalIsolate*, v8::internal::Handle<v8::internal::JSGlobalObject>, v8::internal::Handle<v8::internal::Name>) [node]
 4: 0x1c5aa55  [node]
 5: 0x1c5b9eb v8::internal::compiler::JSGlobalObjectRef::GetPropertyCell(v8::internal::compiler::NameRef const&, v8::internal::compiler::SerializationPolicy) const [node]
 6: 0x1e61ba3 v8::internal::compiler::JSNativeContextSpecialization::ReduceNamedAccess(v8::internal::compiler::Node*, v8::internal::compiler::Node*, v8::internal::compiler::NamedAccessFeedback const&, v8::internal::compiler::AccessMode, v8::internal::compiler::Node*) [node]
 7: 0x1e623c4 v8::internal::compiler::JSNativeContextSpecialization::ReducePropertyAccess(v8::internal::compiler::Node*, v8::internal::compiler::Node*, v8::base::Optional<v8::internal::compiler::NameRef>, v8::internal::compiler::Node*, v8::internal::compiler::FeedbackSource const&, v8::internal::compiler::AccessMode) [node]
 8: 0x1e62f0c v8::internal::compiler::JSNativeContextSpecialization::ReduceJSLoadProperty(v8::internal::compiler::Node*) [node]
 9: 0x1df5e0a v8::internal::compiler::Reducer::Reduce(v8::internal::compiler::Node*, v8::internal::compiler::ObserveNodeManager*) [node]
10: 0x1c92604  [node]
11: 0x1df60be v8::internal::compiler::GraphReducer::Reduce(v8::internal::compiler::Node*) [node]
12: 0x1df7895 v8::internal::compiler::GraphReducer::ReduceTop() [node]
13: 0x1df7c48 v8::internal::compiler::GraphReducer::ReduceNode(v8::internal::compiler::Node*) [node]
14: 0x1ca03fd v8::internal::compiler::InliningPhase::Run(v8::internal::compiler::PipelineData*, v8::internal::Zone*) [node]
15: 0x1ca1291 v8::internal::compiler::PipelineImpl::CreateGraph() [node]
16: 0x1ca1668 v8::internal::compiler::PipelineCompilationJob::PrepareJobImpl(v8::internal::Isolate*) [node]
17: 0xd94f5d v8::internal::OptimizedCompilationJob::PrepareJob(v8::internal::Isolate*) [node]
18: 0xd96358  [node]
19: 0xd97e70 v8::internal::Compiler::CompileOptimized(v8::internal::Isolate*, v8::internal::Handle<v8::internal::JSFunction>, v8::internal::ConcurrencyMode, v8::internal::CodeKind) [node]
20: 0x11f4463 v8::internal::Runtime_CompileOptimized_Concurrent(int, unsigned long*, v8::internal::Isolate*) [node]
21: 0x15e67f9  [node]
Trace/breakpoint trap (core dumped)
Error: Process completed with exit code 133.
```

khassel:
[node v16.9 issue](https://github.com/nodejs/node/issues/40030)

modules/default/updatenotification/node_helper.js
```diff
@@ -1,9 +1,5 @@
-const SimpleGit = require("simple-git");
-const simpleGits = [];
-const fs = require("fs");
-const path = require("path");
+const git_Helper = require(__dirname + "/git_helper.js");
```
rejas, line 1: I'd rename that variable to GitHelper to stay consistent with the other helpers (the filename is ok)

modules/default/updatenotification/git_helper.js
```diff
@@ -0,0 +1,161 @@
+const util = require("util");
+const exec = util.promisify(require("child_process").exec);
+const fs = require("fs");
+const path = require("path");
+
+class gitHelper {
+	constructor() {
+		this.gitRepos = [];
+		this.baseDir = path.normalize(__dirname + "/../../../");
+		if (process.env.JEST_WORKER_ID === undefined) {
```
rejas, line 10: I dont like having special cases for tests in normal code, but I guess you couldnt find a better way neither? (not sure if other modules have something done like this before)

rejas:
Looks nice (except the logger special case but got no idea myself, maybe those sandboxes do something but as I always say, I am bad at the whole jest testing setup). 
Anyway, the 16.9 isssue might soone be resolved if I read the ticket correctly, who knows when they release 16.9.1
Until then no harm in testing with 16.8 explicitly

modules/default/updatenotification/git_helper.js
khassel: > but I guess you couldnt find a better way neither? 

yes ...

modules/default/updatenotification/node_helper.js
khassel: done.

khassel:
> Looks nice (except the logger special case but got no idea myself, maybe those sandboxes do something but as I always say, I am bad at the whole jest testing setup).

I will look at this in a new PR, I got it not running with the sandboxes (the sandboxes are not something introduced with jest, they were already there with mocha). Another approach could be to do the logger special case in `logger.js` (including the code which is now in mock logger.js). And may we could then do the other tests without this sandbox stuff (which has to be tested).

rejas (approved):

MichMich:
After I merged this, all the tests failed. Any idea?

khassel:
in the following test run everything is green, so I would wait if this happens again.

khassel:
@MichMich could reproduce it on my side, I have to adjust the new tests of updatenotification, they are tricky because they use the current state of the mm git repo ...

MichMich:
Isn't there a way to mock that state?

khassel:
I have to think about how to do this, meanwhile I will delete the 2 test with the next PR.

MichMich:
Thanks!

Issue #2649, opened 2021-09-03
replace node-fetch with native fetch module (coming with node v18)


khassel:
We are using `node-fetch` for fetching url's in 3 places:
- calendar
- newsfeed
- e2e tests

With the new version v3 they decided to [convert `node-fetch` to ES module](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#converted-to-es-module) so we can not use `require` anymore. I testet the alternative syntax using `const fetch = (...args) => import('node-fetch').then(({default: fetch}) => fetch(...args));` instead.

The modules calendar and newsfeed are working with this change, but `jest` is failing with a `Segmentation fault`:
```bash
PID 2750 received SIGSEGV for address: 0x10
/opt/magic_mirror/node_modules/segfault-handler/build/Release/segfault-handler.node(+0x3206)[0x7f18049f3206]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7f18046b9730]
node[0xa15ec0]
node(_ZN2v88internal7Isolate38RunHostImportModuleDynamicallyCallbackENS0_6HandleINS0_6ScriptEEENS2_INS0_6ObjectEEENS0_11MaybeHandleIS5_EE+0xad)[0xe4bb5d]
node(_ZN2v88internal25Runtime_DynamicImportCallEiPmPNS0_7IsolateE+0xb1)[0x11fd7d1]
node[0x15ce134]
```

AFAIS this error comes from the headless chromium browser used for the e2e tests.

At this point I stopped my PR for using `node-fetch` v3. They still support v2 with security patches, so we have no problem yet.

For the future we need a solution for this or have to use antoher library.

In `package.json` `node-fetch` must stay on v2.

rejas:
Thanks for the investigation. I agree to keep v2. 
Converting to es modules .... That might help the overall project but has a loooot of implications :-) Future selfs problem I'd say ;-)

fewieden:
@khassel I had a look into your MR. I'm not sure why we spin up the full electron app for those tests. It should be sufficient there to only run in server mode only.

I applied your recent changes from https://github.com/MichMich/MagicMirror/pull/2655/files to your branch `node-fetch`.

I added the import `const app = require("../../js/app");` to the vendor_spec.js file.

Changed the `beforeAll` hook to:

```
beforeAll(function (done) {
  app.start(() => {
    done();
  });
});
```
I received the error:
```
[...]

PR #2664, opened 2021-09-17
Test setup

khassel:
This is a first step to use electron only when needed in the tests, see #2408.
The idea comes from @fewieden see [this comment](https://github.com/MichMich/MagicMirror/issues/2649#issuecomment-917600122).

All tests using node-fetch and the translations test can run in server-only-mode instead using electron. So I make them work for this and moved all the remaining electron tests to a new category `electron` (from `e2e`) hoping we can shift more electron tests to `e2e` in the future.

I also changed `logger.js` again to be silent in the tests.

Tested all changes running mm in server-only mode and running with electron on a pi3.

Some dependencies were updated too.


tests/e2e/global-setup.js
```diff
@@ -1,53 +1,16 @@
-/*
```
rejas, line 1: Dont remove the header, update it :-)

khassel: This is a new file, the old one still exists under `tests/electron`.

tests/electron/env_spec.js
```diff
@@ -0,0 +1,40 @@
+const helpers = require("./global-setup");
```
rejas, line 1: Here a header comment with license is missing

rejas (approved):

tests/electron/env_spec.js
khassel: this is missing in the most test files, I did split this file but the original also didn't contain a header comment.

PR #2670, opened 2021-09-28
New e2e

khassel:
Related to #2408

Managed to move more tests using `electron` to use `server only`.

Because this new tests are using `jsdom` and therefore running in the browser all test configs used must be changed because `require` cannot be used anymore.

There are 2 module tests still using electron (calendar and weather). They are complex (using basic-auth or mock-data), will look into this in a following PR.

Beside the test stuff only `logger.js` and `server.js` are changed and I tested this in `server only` and with `electron` on a pi3.


tests/configs/modules/alert/default.js
```diff
@@ -3,7 +3,13 @@
  * By rejas
  * MIT Licensed.
  */
-let config = require(process.cwd() + "/tests/configs/default.js").configFactory({
+let config = {
+	port: 8080,
```
rejas, line 7: Do these config infos need to be in each file? Cant they be refactored into a common file?

khassel: tell me how, `require` cannot be used in the browser ...

khassel: Solution is simple, we can remove all the default values which are defined in `js/default.js`. Will add a commit.

khassel:
@MichMich the upload error to Codecov seems unrelated to this PR

tests/configs/without_modules.js
```diff
@@ -3,10 +3,11 @@
...
  * MIT Licensed.
  */
-let config = require(process.cwd() + "/tests/configs/default.js").configFactory({
-	ipWhitelist: ["127.0.0.1", "::ffff:127.0.0.1", "::1", "::ffff:192.168.10.1"]
-});
-delete config.modules;
+let config = {
+	ipWhitelist: ["127.0.0.1", "::ffff:127.0.0.1", "::1", "::ffff:192.168.10.1"],
+
+	units: "metric"
```
rejas, line 9: do we need that unit config entry?

rejas (approved):

tests/configs/without_modules.js
khassel: no, did an automatic search/replace, will check again and update.

rejas:
Looks good to me and though it doesnt change much in the non-test code, maybe we wait until the october release is out to merge it? Juuuust to be on the safe side :-D

tests/configs/modules/alert/default.js
rejas: Yeah, I thought about using import but then a lot more would have to be refactored IIRC

khassel:
> maybe we wait until the october release

let we decide @MichMich , have no problem to wait ...

MichMich:
> Looks good to me and though it doesnt change much in the non-test code, maybe we wait until the october release is out to merge it? Juuuust to be on the safe side :-D

Whoops! Missed that one. Well, let's start living on the edge. 😂

PR #2676, opened 2021-10-01
module animation

eouia:
# Module show/hide Animation restructuring
- Delegation of the animation from MM `main.js` to module class `module.js`
- It enables the module to handle and decide its own animation on showing/hiding self by overriding.
- 3rd-party module could give custom animation by `.show()` and `.hide()`

## new Module methods `.showAnimation()` and `.hideAnimation()`
The module developer can decide how the module would be showing/hiding by `.showAnimation()` and `.hideAnimation()` 

For backward compatibility and lower-power device, default animation behaviours of module class is defined like with the legacy transition of original MM.
```js
showAnimation(moduleWrapper, speed) {
  moduleWrapper.style.opacity = 1;
}
hideAnimation(moduleWrapper, speed) {
},
```
But a module developer can override these methods when he needs;

```js
hideAnimation: function (moduleWrapper, speed) {
  let keyframe = [
    { transform: "rotate(0deg) scale(1, 1)", easing: "ease-in" },
    { transform: "rotate(1080deg) scale(0, 0)",  easing: "ease-out" }
  ];
  let animation = moduleWrapper.animate(keyframe, {
    duration: speed
  });
  /*
  animation.onfinish  = () => {
    console.log("Hiding animation finished");
  };
  */
},
```
In this example, I showed to use CSS animation, but there could be other approaches to show or hide. (I can imagine iconizing or application bar style approach..)

## new animation callback option for `.show()` and `.hide()`
For paging modules or similar working modules which control other module's show/hide, `.show(speed, options)` and `.hide(speed, options)` can get new `animation` callback property for `options` of method parameter.

```js
const myAnimation = ({ moduleWrapper, module, speed }) => {
  // own animation logic...
}

MM.getModules().enumerate((m) => {
  m.show(1500, {
    lockstring: "lockstring",
    animation: myAnimation
  })
})
```
`animation` property is a callback function that has an `object` as an argument. That object could have `moduleWrapper` , `module` and `speed` as properties. A developer can assign a custom animation for that target module for this time of calling.

## Performance

The RPI3 or older might not have enough power to handle various animations. But at this moment, RPI4 or other stronger SBC/Devices are also used for MM. The developer, who wants to use custom animation, needs to be careful to use or provide fallback together. Anyway, that would be the charge of the developer.

## Demo

https://user-images.githubusercontent.com/1720610/135678350-9ac55640-7b25-4719-ad0a-f4a7041751f8.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,11 @@ _This release is scheduled to be released on 2022-01-01._
 
 ### Added
 
+- `showAnimation()` and `hideAnimation()` to override default show/hide animation of the module. And `options` of `.show()` and `.hide()` can give the custom animation to modules.
+
 ### Updated
+
+- Delegate show/hide animation from MM `main.js` to `module.js`. (Module can handle its or other's show/hide animation.)
 
 ### Fixed
 
--- a/js/loader.js
+++ b/js/loader.js
@@ -107,7 +107,8 @@ const Loader = (function () {
 				header: moduleData.header,
 				configDeepMerge: typeof moduleData.configDeepMerge === "boolean" ? moduleData.configDeepMerge : false,
 				config: moduleData.config,
-				classes: typeof moduleData.classes !== "undefined" ? moduleData.classes + " " + module : module
+				classes: typeof moduleData.classes !== "undefined" ? moduleData.classes + " " + module : module,
+				customAnimation: moduleData.customAnimation
 			});
 		});
 
--- a/js/main.js
+++ b/js/main.js
@@ -243,8 +243,11 @@ const MM = (function () {
 
 		const moduleWrapper = document.getElementById(module.identifier);
 		if (moduleWrapper !== null) {
-			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
-			moduleWrapper.style.opacity = 0;
+			if (typeof options.animation === "function") {
+				options.animation({ moduleWrapper, module, speed });
+			} else {
+				module.hideAnimation(moduleWrapper, speed);
+			}
 
 			clearTimeout(module.showHideTimer);
 			module.showHideTimer = setTimeout(function () {
@@ -307,7 +310,6 @@ const MM = (function () {
 
 		const moduleWrapper = document.getElementById(module.identifier);
 		if (moduleWrapper !== null) {
-			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			// Restore the position. See hideModule() for more info.
 			moduleWrapper.style.position = "static";
 
@@ -315,7 +317,12 @@ const MM = (function () {
 
 			// Waiting for DOM-changes done in updateWrapperStates before we can start the animation.
 			const dummy = moduleWrapper.parentElement.parentElement.offsetHeight;
-			moduleWrapper.style.opacity = 1;
+
+			if (typeof options.animation === "function") {
+				options.animation({ moduleWrapper, module, speed });
+			} else {
+				module.showAnimation(moduleWrapper, speed);
+			}
 
 			clearTimeout(module.showHideTimer);
 			module.showHideTimer = setTimeout(function () {
--- a/js/main.js
+++ b/js/main.js
@@ -231,6 +231,10 @@ const MM = (function () {
 	 * @param {object} [options] Optional settings for the hide method.
 	 */
 	const hideModule = function (module, speed, callback, options) {
+		const defaultAnimation = function (moduleWrapper, speed) {
+			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+			moduleWrapper.style.opacity = 0;
+		};
 		options = options || {};
 
 		// set lockString if set in options.
@@ -246,7 +250,7 @@ const MM = (function () {
 			if (typeof options.animation === "function") {
 				options.animation({ moduleWrapper, module, speed });
 			} else {
-				module.hideAnimation(moduleWrapper, speed);
+				module.data.customAnimation ? module.hideAnimation(moduleWrapper, speed) : defaultAnimation(moduleWrapper, speed);
 			}
 
 			clearTimeout(module.showHideTimer);
@@ -280,6 +284,11 @@ const MM = (function () {
 	 * @param {object} [options] Optional settings for the show method.
 	 */
 	const showModule = function (module, speed, callback, options) {
+		const defaultAnimation = function (moduleWrapper, speed) {
+			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+			moduleWrapper.style.opacity = 1;
+		};
+
 		options = options || {};
 
 		// remove lockString if set in options.
@@ -321,7 +330,7 @@ const MM = (function () {
 			if (typeof options.animation === "function") {
 				options.animation({ moduleWrapper, module, speed });
 			} else {
-				module.showAnimation(moduleWrapper, speed);
+				module.data.customAnimation ? module.showAnimation(moduleWrapper, speed) : defaultAnimation(moduleWrapper, speed);
 			}
 
 			clearTimeout(module.showHideTimer);
--- a/js/module.js
+++ b/js/module.js
@@ -202,6 +202,32 @@ const Module = Class.extend({
 		Log.log(this.name + " is resumed.");
 	},
 
+	/**
+	 * Called when the module needs the animation of hide.
+	 *
+	 * @param {HTMLDom} moduleWrapper Usually this module's HTML Dom
+	 * @param {number} speed animation speed
+	 *
+	 */
+
+	hideAnimation(moduleWrapper, speed) {
+		moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+		moduleWrapper.style.opacity = 0;
+	},
+
+	/**
+	 * Called when the module needs the animation of show.
+	 *
+	 * @param {HTMLDom} moduleWrapper Usually this module's HTML Dom
+	 * @param {number} speed animation speed
+	 *
+	 */
+
+	showAnimation(moduleWrapper, speed) {
+		moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+		moduleWrapper.style.opacity = 1;
+	},
+
 	/*********************************************
 	 * The methods below don"t need subclassing. *
 	 *********************************************/
--- a/js/module.js
+++ b/js/module.js
@@ -207,10 +207,9 @@ const Module = Class.extend({
 	 *
 	 * @param {HTMLDom} moduleWrapper Usually this module's HTML Dom
 	 * @param {number} speed animation speed
-	 *
 	 */
 
-	hideAnimation(moduleWrapper, speed) {
+	hideAnimation: function (moduleWrapper, speed) {
 		moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 		moduleWrapper.style.opacity = 0;
 	},
@@ -220,10 +219,9 @@ const Module = Class.extend({
 	 *
 	 * @param {HTMLDom} moduleWrapper Usually this module's HTML Dom
 	 * @param {number} speed animation speed
-	 *
 	 */
 
-	showAnimation(moduleWrapper, speed) {
+	showAnimation: function (moduleWrapper, speed) {
 		moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 		moduleWrapper.style.opacity = 1;
 	},
--- a/js/module.js
+++ b/js/module.js
@@ -205,7 +205,7 @@ const Module = Class.extend({
 	/**
 	 * Called when the module needs the animation of hide.
 	 *
-	 * @param {HTMLDom} moduleWrapper Usually this module's HTML Dom
+	 * @param {object} moduleWrapper Usually this module's HTML Dom
 	 * @param {number} speed animation speed
 	 */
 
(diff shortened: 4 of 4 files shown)


MichMich:
I looks awesome. But I'm curious if I understand it correctly:

Is this a way for every module developer to change the hide/show animation of their module? Or is this a way to easily overwrite the hide/show method for each method?

The issue is that if every module Strat implementing their own animation style, it quickly becomes a mess. I'm not sure if this is something we should promote. 

What do you (and others) think?

sdetweil:
Yeh, I'm not sure I understand the value to MM..  is this an info panel or a game platform?  why would we want to encourage different components to act differently

skuethe:
> Yeh, I'm not sure I understand the value to MM.. is this an info panel or a game platform? why would we want to encourage different components to act differently

I think this comment is a bit short-sighted. Just because it does not fit your use-case, it does not mean others couldn't benefit from it.

That being said, I do feel that giving this option to module developers would potentially create more problems than it would "fix". A lot of people would probably start asking support for "fixing that weird animation module developer XYZ has introduced".

Instead of the module developers, how about giving this as a new addition to everybodys `config.js`? Wouldn't that be possible?
Don't pinpoint me on specifics here, as I currently don't have insights to MM² code, but maybe it is possible to give a "library" of basic animations ("keyframes") and ppl can then reference these for each and every module in their `config.js`?

rejas:
Formally the PR is superb, lots of explanation and screenshort, excellent work. The example animation @eouia provides looks very nice. 

As for the usefulness, I do agree with michmichs concern that it would fragment the look&feel of the MagicMirror modules if every module developer adds his own style.

Then again, even today modules can implement their own styles for some stuff, that dont match anything else: For example, some modules fade out long text lines, some cut them of with "...", some add a marquee scrolling effect on them. 

Since the MM allows overwriting the styles, there is nothing that can be done here to enforce a consistent styling (or can it?)
so why not allow adding custom animations?

So, in the end I would prefer to add different animations as a config options ("fade", "scroll", "whatever" as @skuethe suggest. This way, the user can decide how his mirror looks/feels and not the module developer.

eouia:
Thanks for all your interest.

I agree entirely with all your concerns. But still, my answer is;
**"Possible but not doing" is not **"impossible"**, it is my basic mind.
 
First point;
Due to the limit of SBC power or aesthetic point of view or whatever reason, many users will prefer a default behaviour or at least harmonized behaviours. Developers will respect their user's wishes. If some module has too weird behaviour breaking the harmony of the MM screen, users will not use it. (or  will request a more humble behaviour to the developer) I think most Developers will provide options to the user with config.

Second point;
Fortunately(?) `module object` of the MM is not encapsulated. So we can **OVERWRITE** (not override) the behaviours whenever we want at runtime.
```js
MM.getModules().enumerate((m) => {
  m.showAnimation = (moduleWrapper, speed) => {
  moduleWrapper.style.opacity = 1;
})
```
This code could make returning the show animation of the module back to the default. 
Maybe the `MMM-Humble` module can make a calm down the crazy modules.
My original intention of this feature is, "Commander" module to control other module's behaviours. To make a slave,  he needs to be a free man at first. :)
Anyway, I hope to see "really sliding contents right to left in the screen" when I use a paging module.

eouia:
@skuethe @rejas 
For the user's config instead of developer, Well, good idea. At the first time, I considered that. 
But I guess it would be enabled more easily by A `Commander` module instead of each configuration of modules. Maybe the Commander will be some paging modules or something similar modules which want to control other module's show/hide. And naturally, the Commander will keep the harmony of entire modules and will permit the exception only he wants. I love dictatorship. My feature is never to give uncontrolled freedom to the users, but rather to rule them. 👿

MichMich:
Just to make it clear: I like the idea of extending the transitions very much. You did some great work. Thanks for that!

As said, I think we should try our best to prevent all module developers from adapting their own style. 

What would be an idea is to allow the user to configure a transition module that handles all the transitions for the mirror. By default we use our own transition module which just does the current simple fade. 

But a more complex third party module could allow the transitions you showed. 

This way all the modules will use the same style. 

What do you think of that?

How can we implement this best? Let's discuss!

rejas:
> What would be an idea is to allow the user to configure a transition module that handles all the transitions for the mirror. By default we use our own transition module which just does the current simple fade.

Exactly what I was thinking :-). This way, other people, who are interested in writing cool animations but not in creating a whole module, can contribute too.

eouia:
https://github.com/MMRIZE/public_ext_storage/blob/main/showHideAnimation/test.js

This is the test module code that I've tested.

https://github.com/MMRIZE/public_ext_storage/blob/ba30b5e6ff5a87767d23dfdd178aa8c33d711446/showHideAnimation/test.js#L124-L146

Here I made a weird animation (crazy spiral movements)
Maybe this is too kinky for most of the users, and breaking the harmony.

https://github.com/MMRIZE/public_ext_storage/blob/ba30b5e6ff5a87767d23dfdd178aa8c33d711446/showHideAnimation/test.js#L23-L37

And here, I made the `compliments` module also being crazy too, by **OVERWRITE** default behaviours.

https://github.com/MMRIZE/public_ext_storage/blob/ba30b5e6ff5a87767d23dfdd178aa8c33d711446/showHideAnimation/test.js#L42-L122

But here, these two crazy behaviours are controlled by force.  And the result was above DEMO video.

I believe this would show the whole thing could be under control by other modules. So that is why I'm not worrying too much and the current PR is enough (at my think).


However, changing of point of view, 
We can consider drop `.hideAnimation()` and `.showAnimation()`, and leave only `.show(options.animation)` and `.hide(options.animation)`. This could prevent the self-expression of module, but still controllable from other modules. I think this could be also considerable.

eouia:
> > What would be an idea is to allow the user to configure a transition module that handles all the transitions for the mirror. By default we use our own transition module which just does the current simple fade.
> 
> Exactly what I was thinking :-). This way, other people, who are interested in writing cool animations but not in creating a whole module, can contribute too.

For this; I think it would be done in the specific module better than 'config.js',

MichMich:
I'm not sure if we are on the same page. But it seems we are slowly getting a good idea of wat could be the best option.

In my opinion we should aim for the following:

- The default behaviour remains the same. 
- No module should be able to (only) override it's own animation style without the users explicit consent. This way we prevent fragmentation in the style. It should remain clean and uniform.
- A module can be the designated module which handles the animation for all modules on the mirror. This can be a module that ONLY does the animation. But it could also add other features. For instance, a paging module could also include advanced animations to make the paging mechanism a bit cooler.

Making it the designated animation module should be part of the module settings, not part of the module config:

```
modules: [
		{
			module: "clock",
			position: "top_left"
		},
		{
			module: "weather",
			position: "top_right",
			config: {
			    ...
			}
		},
		{
			module: "MMM-AdvancedAnimation",
			designatedAnimator: true,
			config: {
			    enable3DEffects: true
			}
		},
]
``` 

So the important part here is: `designatedAnimator`. Although I'm not sure this is the best config name. 

What could ba a nice addition is to allow `designatedAnimator` to be an (array of) class names of the modules which it is allowed to control. This way an animation module can be limited to specific modules if the user wants this. If it's set to true, all animation will be handled by this module. 

If multiple modules are set as `designatedAnimator`, a warning should be shown, and only the first one in the list should be used for the animation.

Your thoughts on the above?

eouia:
Ok. I added `customAnimation` module setting.  (I think the term `customAnimation` is more intuitive than `designedAnimator`. But this is less important matter, so I don't care any naming)
It enables explicitly the user to allow the module's own show/hide animation or not. 
- `customAnimation: true` : apply the module's animation.
- `customAnimation: false` or undescribed : apply default behaviours (transition opacity 1s)

This option doesn't prevent `.show(options.animation)` and `.hide(options.animation)`, so a 3rd-party module still can give instant animation when it call `.show()` and `.hide()`.  So, Paging module can "slide all module" regardless of user's `customAnimation` options. (Anyway the user himself have decided to use that paging module.)

Here is a DEMO



https://user-images.githubusercontent.com/1720610/135751404-94888b72-985d-4db2-828a-0a142b6a0cf9.mov

I injected all modules to have spiral showing animation,
But in `config.js`;
image

I explicitly allow only the `clock`  and `compliments` modules to use their own animation.

MichMich:
The proposed name was `designatedAnimator` in stead of `designedAnimator` what an other meaning. But as said, the name is not of any importance right now. 

I do think we both have a different idea about the implementation. If I'm correct, your proposal is to allow a module to change the way *it* animates. 

My proposal is not to allow a module to change *its* transitions. But to allow a module to be in charge of ALL animations. 

This will allow "theme builders" to build a module that adds a total new way all the modules on a mirror transition. 

So to make things more clear, we end up with:

- Modules: a module adds content to the mirror. These already currently exist: default modules as well as third party modules. 
- Animators: this is a new type of plugin which does not add content to the mirror, but ads a new transition style to a mirror. 

To keep things simple, both the module functionality as well as the animator functionality can be out in the same kind of package. This way they can be combined. 

Does this make sense to you?

eouia:
@MichMich 
> My proposal is not to allow a module to change _its_ transitions. But to allow a module to be in charge of ALL animations.
> 
> This will allow "theme builders" to build a module that adds a total new way all the modules on a mirror transition.
> So to make things more clear, we end up with:
> 
> * Modules: a module adds content to the mirror. These already currently exist: default modules as well as third party modules.
> * Animators: this is a new type of plugin which does not add content to the mirror, but ads a new transition style to a mirror.
> 
> To keep things simple, both the module functionality as well as the animator functionality can be in the same package. This way they can be combined.
> 

Now I understand your intention halfly. But still hard to understand how the shape it can be in real world.
 
- How the user can inject the `Animator` into the MM? I think `Animator` will be a some `callback function or class object` which would be called when the animation is needed.

In my guess, the imaginable user scenario might be;

1. Store some `Animator.js` into the `animators` directory.  For example; 
```
/MagicMirror
-- /animtors
  -- default_transit.js (single file or package directory whatever)
  -- horizontal_sliding.js
  -- flash_bounce.js
```
Or could it be in the module together;
```
/MagicMirror
-- /modules
  -- /MMM-Something
  -- MMM-Something.js
  -- /animation
    -- /flash_bounce.js 
```
2. Then user load Animators which he want to use explicitily
```js
//config.js
...
address: "localhost",
animators: ["default_transit", "flash_bounce", ...],

// or with some options properties;
animations: [
  "flash_bounce": {
    defaultSpeed: 1000,
    direction: "up-down",
    ...
  }, ...
]
...

// or just toggle using on/off then dynamically loading at runtime. 
animators: true,
```
Anyway, you might prefer the explicit allowance from user be needed.


3. Can user assign animator to the specific moule? I'm not sure you intention is this or not.
```js
modules: [
  {
    module: "MMM-Something",
    animator: "flash_bounce",
    position: "top_left"
  }
  ...
  ```
  
  4. How can 3rd party module activate or control the animation of other modules?
  ```js
  MM.setAnimator({hide: "horizontal_sliding"})
  MM.getModules().enumerate((m) => { m.hide(1000) }
  
  // or
  MM.getModules().enumerate((m) => {m.hide(1000, {animator: "horizontal_sliding"})
  ```
  
  It will be better to see concrete usecase to understand.

MichMich:
Sorry, I don't think you understand correctly. I used the name "animator" to explain that it has an other purpose than a default module. What I meant is that a module can either be something that adds content, or something that controls animation. (Or both, but ignore this for now.)

In the config the user has to state explicitly that a module is allowed to control animation (and with animation, I mean hide/show transitions). Without this explicit setting, the animation part of the module will never be called. The built in system will be used.

What happens internally is the following. 

1. The MagicMirror needs to transition a module.
2. The MagicMirror searches for the module that is marked as the designated (please look up that word so you understand correctly :)) animator.
3. The MagicMirror core says to that animator module: "Hey, please hide module X for me. And let me know when you're done.": `animatorModule.hideTransition(moduleX, settings, callback)` (or better: we could use promises).
4. The animator module executes the animation (using the setting, which includes stuff like desired animation duration).
5. The animator module tells the core it's done, and the cure continues with the stuff that is normally done after the hide transition.

Now, a few important things:

- If a module wants to hide/show, it is the same as always. The module just says `m.hide(1000)`. It doesn't say what type of animation is desired. That's up to the core (based on the config). From the module's standpoint, nothing changes.
- If there is no module set as designated animator, it just uses the default built in animation (which is the simple fadeInOut).
- Using class names, we could allow to only set designatedAnimators for specific modules. Which also allows the module to be it's own animator But only if the users allows this in the config.
- In some cases you might not want fancy transitions. For example: when content updates, a simple fade is nicer than a full spinning animation. This could be part of the settings property. Is this because it is a content update? Or is it just a request to hide/show it. The core knows this, and could simply put this in the settings.
- If the animatorModule for any reason doesn't want to do the animation, it could hand it off to the built in animator. This could be done by letting the `animatorModule.hideTransition(moduleX, settings, callback)` return false. Saying "Do it yourself, I don't want to do it for you!". For example, if the animation is because it's a content change (see previous bullet), it could just return false. 

Does this make my idea clear?

@rejas @khassel @sdetweil Your thoughts?

eouia:
Ok. Let's forget this PR itself. Let's discuss from baseline.

Plz, check what I understand (I thought)

1) `Animator` is a kind of specific MM module.
2) The user can explicitily decide to use the `Animator` by adding the module definition in `config.js`
3) The original behaviours should be respected because they also be used on contents updating not only show/hide from `Paging module`.
4) So, when the `core` needs extended animation, `core` will call Animator's extended method. Only in that case the decorated animation would be executed.
5) If `Animator` reject a specific animation of module at any reason, default behaviours should be done. 

Am I right?

MichMich:
Yes, i think you are right. This might be the best startingpoint for further discussion. 

I thing the biggest goal is to make this all work without adding lots of code. We should keep it to an absolute minimum.

eouia:
@MichMich 
Please see #2681.
Finally, I conclude that Only the `Animator` module needs the custom effects on the other module's transition. Just `.show()` and `.hide()` with optional animation options would be enough for that. It will not break the current structure and legacy modules.

PR #2679, opened 2021-10-02
Clean up updatenotifications and nunjuck templating

fewieden:
Does the pull request solve a **related** issue?

No

What does the pull request accomplish? Use a list if needed.
- Switch from manual DOM creation to nunjuck templating
- Fix async/await issues
- Removed exceptions in code and replaced them with jest mocks

If it includes major visual changes please add screenshots.
  - Top is nunjuck template
  - Bottom is current rendering
updatenotifications


fewieden:
I bumped ESLint to support ECMAScript 2018. It complained about the object literal spread operator which is supported since node v8.3

image

MichMich:
Sorry, am I missing something in the image? I don't see a difference between the top and bottom.

fewieden:
@MichMich that's good, it should look the same.

I just added it as a reference, since I changed the whole rendering of the module.

rejas (approved):
Nice cleanup!

MichMich:
If you feel like it ... the alert module could use the same TLC. 😅

PR #2680, opened 2021-10-02
Use feels_like data from openweathermap instead of calculating it

### LINKED ISSUE #2678: weather module: Feels Like temperature doesn't obey units correctly

**Platform**: Raspberry Pi3

**Node Version**: v14.17.6

**MagicMirror Version**: 2.16.0

**Description**: 
When using the weather module with the 'current' type the feelsLike temperature is incorrect unless you explicitly set the tempUnits. It appears to use the Fahrenheit temperature but believes that it is a Celsius temperature and therefore tries to convert it to Fahrenheit. 

**Steps to Reproduce**: 
Configure the weather module similar to this:
```
{
	module: "weather",
	position: "top_left",
	config: {
		weatherProvider: "openweathermap",
		type: "current",
		units: "imperial",
		apiKey: "xxx",
	}
}
```

**Expected Results**: Feels like temperature displays correctly. 

**Actual Results**: Feels like temperature displays a number in the upper 1 or 200's

**Configuration**: See above

**Additional Notes**: 
This is fixed if the weather module is configured as follows:
```
{
	module: "weather",
	position: "top_left",
	config: {
		weatherProvider: "openweathermap",
		type: "current",
		units: "imperial",
		tempUnits: "imperial",
		windUnits: "imperial",
		apiKey: "xxx",
	}
}
```

rejas:
Fixes #2678

When getting the current weather from openweathermap, the provider ignored the feels_like data from the response, and instead calculated the information himself.

Due to how the units were configured, the result was way off (due to imperial and metric units being mixed together). This weird logic is not changed in this PR. That's for another day...


kevinsivic (approved):
This change does resolve the issue.

PR #2681, opened 2021-10-04
module animation

eouia:
This is a new PR for the result of the previous #2676 discussion.

DOT: `Animator`: A specific MM module that wants to make other module's transitions animated with custom effects.

- Enable an `Animator` module to give special transition effects only when it calls `target.show()` and `target.hide()` with proper options.
- In other cases, original behaviours will be applied.
- Animation effects in each call are instant, not permanently applied. 
- Minimum changing of original core codes. 
- No need to modify current legacy modules including 3rd-party by this change.
- The only module that wants to use this feature, needs an additional option property `animation` in `.show()` and `.hide()`

### quick modifying of MMM-Pages
```js
lockStringObj.animation = async ({ moduleWrapper, module, speed }) => {
  let keyframe = [
    { transform: "translate(0, 0)", easing: "ease-out" },
    { transform: `translate(-50vw, 0)`, easing: "ease-in" }
  ];
  let animation = moduleWrapper.animate(keyframe, { duration: speed });
}

MM.getModules()
  .exceptWithClass(modulesToShow)
  .enumerate(module => module.hide(animationTime, lockStringObj));
```
RESULT

https://user-images.githubusercontent.com/1720610/135819704-762b6d61-5e17-4786-99c7-1d0c9e0743b7.mov

### quick modifying of MMM-Carousel
```js

        var hideOptions = {
            lockString: "mmmc",
            animation: async ({ moduleWrapper, module, speed }) => {
                let keyframe = [
                  { transform: "translate(0, 0)", easing: "ease-out" },
                  { transform: `translate(-50vw, 0)`, easing: "ease-in" }
                ];
                let animation = moduleWrapper.animate(keyframe, {
                  duration: speed
                });
            }
        }
        
...
this[I].hide(this.slideTransitionSpeed, hideOptions);
...
```
RESULT

https://user-images.githubusercontent.com/1720610/135820103-8b043acf-f2f4-4fc3-a231-e0d9f77372f0.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2022-01-01._
 
 ### Added
 
+- Added `animation` option property to `.show()` and `.hide()` for enabling custom animation effects in show/hide transition.
+
 ### Updated
 
 ### Fixed
--- a/js/main.js
+++ b/js/main.js
@@ -243,9 +243,12 @@ const MM = (function () {
 
 		const moduleWrapper = document.getElementById(module.identifier);
 		if (moduleWrapper !== null) {
-			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
-			moduleWrapper.style.opacity = 0;
-
+			if (typeof options.animation === "function") {
+				options.animation({ moduleWrapper, module, speed });
+			} else {
+				moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+				moduleWrapper.style.opacity = 0;
+			}
 			clearTimeout(module.showHideTimer);
 			module.showHideTimer = setTimeout(function () {
 				// To not take up any space, we just make the position absolute.
@@ -307,7 +310,6 @@ const MM = (function () {
 
 		const moduleWrapper = document.getElementById(module.identifier);
 		if (moduleWrapper !== null) {
-			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			// Restore the position. See hideModule() for more info.
 			moduleWrapper.style.position = "static";
 
@@ -315,7 +317,12 @@ const MM = (function () {
 
 			// Waiting for DOM-changes done in updateWrapperStates before we can start the animation.
 			const dummy = moduleWrapper.parentElement.parentElement.offsetHeight;
-			moduleWrapper.style.opacity = 1;
+			if (typeof options.animation === "function") {
+				options.animation({ moduleWrapper, module, speed });
+			} else {
+				moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
+				moduleWrapper.style.opacity = 1;
+			}
 
 			clearTimeout(module.showHideTimer);
 			module.showHideTimer = setTimeout(function () {


MichMich:
It's a nice implementation. But not really what we discussed, IMHO. With this implementation it still means every module can set arbitrary animations per hide/show action. This means that the whole mirror will be filled with multiple different animations without the user's consent. I'm not sure this is desirable. 

I think its worth to have a thorough discussion about this in an issue before we implement something. (It's easy to implement a feature. It's impossible to take out a feature without breaking stuff).

Is it a good idea to first open an issue so we can settle on the desirable concept and api?

eouia:
> It's a nice implementation. But not really what we discussed, IMHO. With this implementation it still means every module can set arbitrary animations per hide/show action. This means that the whole mirror will be filled with multiple different animations without the user's consent. I'm not sure this is desirable.
> 
> I think its worth to have a thorough discussion about this in an issue before we implement something. (It's easy to implement a feature. It's impossible to take out a feature without breaking stuff).
> 
> Is it a good idea to first open an issue so we can settle on the desirable concept and api?

IDEALLY and TOTALLY, agree with you.

But in real use-case, I conclude Only some paging modules(or similar commanding modules) needs custom animated transition.  I believe that the activity of a user adding this kind of module is explicit consent itself.

As our previous discussion, 
- MM core delegates animation to `Animator` only when core thinks it needs,
- `Animator` decide to express custom animation or not 

To achieve that; 
0) A user explicitly select Animator to apply.
1) Inject Animator method (callback, or ...) into core `main.js` at bootup time or runtime
2)  core `main.js` decide which behaviours are adopted. If it thought custom animation is needed, injected Animator method will be called.
3) Animator method will try to express animation or reject it for some reason.
4) If rejected, the core will do default behaviour.

It is a Not TOO complex or expensive job to implement, though...

What is the purpose of this implementation really? 

1) To prevent uncontrolled self-expression of the module itself;
Frankly, a module can override its hide/show without this PR.
```js
this.hide = function (speed, options) {
   // own animation logic synchronised during speed.
   ...
   MM.hideModule(this, 0, options) // then instantly hide modue.
}
```
With this kind of code, A module already can do self-expression whatever they want. And even further, Other module's `.show()` and `.hide()` could be reassignable also with `bind()`. 
This thing is a real uncontrolled situation. So with or without this PR, this purpose is unreachable. Rather providing official ways to do might be better than uncontrolled.

and for self-expression itself; If A module wants really to have weird behaviours, Should the system prevent or allow it? In my thought, A user should decide whether or not. And if a user chooses to use that weird ugly module, it has to be allowed unless it crashes the whole system's running, not just aesthetic point of view.
Anyway, uncontrolled self-expression is not the issue anymore of this PR. 

2) Animator module has a charge of the whole custom animations.
But what needs custom animated transition really? Maybe just some paging/commanding modules will. Without paging/commanding modules, custom animations are meaningless. (for self-updating content, It is already achievable by the module itself)
I think those modules will become the `Animator` finally.  Most users are using only one or two paging/commanding modules. It makes sense the paging/commanding module itself would be regarded as an Animator then allow them to control modules' animations for their needs.  
Then, It might be enough only to give the ability to those kinds of modules.  The decision of using those modules probably becomes the explicit consent from the user.
And for other modules, even though they are opened to use `.show()` and `.hide()` by themselves, what happens wrong? they will use that feature if they need it. And the decision would be a charge of users who select that module's behaviours.

After writing above sentences, I finally realized this PR is not needed indeed to achieve my wants, because I can overwrite all the module's behaviours without modifying cores at run-time unless the module object becomes encapsulted.  So I don't care this PR is accepted or not any more.

Thanks @MichMich, It was a really interesting topic.

MichMich:
What might be an idea is to add setting `allowCustomAnimations` for each module which defaults to true. This way a user can always disable the custom animations. If it's set to false, the animation callback will just be ignored.

eouia:
> What might be an idea is to add setting `allowCustomAnimations` for each module which defaults to true. This way a user can always disable the custom animations. If it's set to false, the animation callback will just be ignored.

Yup, good idea. 
Anyway, I'm trying proof of concept of my idea. Then I might make a new PR (or not... ^_^)

PR #2682, opened 2021-10-04
fix language Code kr->ko and add missed translation

eouia:
The right languageCode(ISO 639-1) of the Korean language is `ko` or `ko-kr`, not `kr`. 
For backward compatibility; `kr.json` needs to be kept together for a while, but in the future, `kr.json` should be eliminated. 

`kr` is Kanurian language in Africa. (https://en.wikipedia.org/wiki/Kanuri_language)


rejas (approved):

MichMich:
When adding a new translation, the [translations.js](https://github.com/MichMich/MagicMirror/blob/develop/translations/translations.js) needs to be updated as well. In this case this would allow you to make the `kr` an alias of the `ko` translation.

translations/translations.js
```diff
@@ -34,7 +34,8 @@ let translations = {
 	hu: "translations/hu.json", // Hungarian
 	is: "translations/is.json", // Icelandic
 	et: "translations/et.json", // Estonian
-	kr: "translations/kr.json", // Korean
+	ko: "translations/ko.json", // Korean
+	kr: "translations/kr.json", // Wrong File for Korean, This code is reserved for Kanurian, but for backward compatibility; it needs keeping a while; this would be eliminated or replaced with Kanuri in future.
```
MichMich, line 38: `kr` could just point to `translations/ko.json` as well. This way we don't need to keep the two files.

And to be honest. You can just remove the 'ko' reference as well if we make note of it in the changelog. It's an easy fix for users who use the wrong language code.

rejas (changes requested):

PR #2694, opened 2021-10-17
Update github actions

rejas:
Time to update some github actions, since 
- codecov will deprecate their v1 action (see https://github.com/codecov/codecov-action)
- github setup node allows caching of the npm modules
- the enforcechangelog got updated too

Lets see what happens if I create this PR first with no CHANGELOG entry :-)


rejas:
Changelog Enforcer still works (and fails :-) Lets add a CHANGELOG

MichMich:
Seems to work. 😬

PR #2695, opened 2021-10-18
removed duplicate "weather" entry in config.sample.js

Bee-Mar:
Removed the duplicate "weather" entry in config.sample.js. According to the documentation for the new "weather" module, there should only be one entry. 

This pull request was primarily created to solve a issue on the MagicMirror Package Manager repo.

https://docs.magicmirror.builders/modules/weather.html#usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ _This release is scheduled to be released on 2022-01-01._
 - Fixed wrong file `kr.json` to `ko.json`. Use language code 'ko' instead of 'kr' for Korean language.
 - Fixed `feels_like` data from openweathermaps current weather being ignored (#2678).
 - Fixed chaotic newsfeed display after network connection loss thanks to @jalibu (#2638).
+- Removed duplicate "weather" entry in `config.sample.js`
 
 ## [2.17.1] - 2021-10-01
 
--- a/config/config.js.sample
+++ b/config/config.js.sample
@@ -78,18 +78,6 @@ let config = {
 			}
 		},
 		{
-			module: "weather",
-			position: "top_right",
-			header: "Weather Forecast",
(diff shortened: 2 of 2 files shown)


rejas:
> This pull request was primarily created to solve a issue on the MagicMirror Package Manager repo.

What issue are you refeerring to? The config is ok, so maybe there is something else that should be fixed?

Bee-Mar:
@rejas @MichMich sorry, it wasn't clear to me in the documentation the weather module could be used for both types of weather. A user of my software reported an issue caused by the weather module being listed twice. I'll close this pull request.

sdetweil:
from the doc 

```
#Weather Module
This module is aimed to be the replacement for the current currentweather and weatherforcast modules. The module will be configurable to be used as a current weather view, or to show the forecast. **This way the module can be used twice to fulfill both purposes.**
```

Bee-Mar:
@sdetweil It's mentioned the module is intended to be a replacement for both of the aforementioned modules, but not that it ought to be listed twice to replace both. The sample shows a single entry, which is the reason for the confusion. I've already closed the pull request.

sdetweil:
@Bee-Mar 
but THIS sentence is part of the doc too

**This way the module can be used twice to fulfill both purposes.**

Bee-Mar:
@sdetweil I've already closed the pull request.

sdetweil:
i understand, but I want to correct the misimpression that it was incorrect for some reason.,

your statement 
```
it wasn't clear to me in the documentation the weather module could be used for both types of weather.
```
is confusing. as the doc CLEARLY says its for BOTH AT THE SAME TIME and then shows it in the sample

SO, if we need to fix the doc, how could we do that?  you are a pretty savy guy and missed it somehow.

PR #2698, opened 2021-10-20
Update scripts to run under windows

rejas:
Some people want to run MM under windows, so this PR adds a new start script entry without the display stuff.

It also makes the tests runnable under windows thanks to the cross-env library.


rejas:
The alert test still seems unstable after https://github.com/MichMich/MagicMirror/pull/2687 @khassel  ... Any idea on why it is so picky?

sdetweil:
is install fixed?  vendor and fonts do not get installed as the script is wrong for windows batch

sdetweil:
if we went back to a script for start, then there could be two scripts 
start_it   (sh, but no extension)
start_it.cmd 

then launch would the 
  "start":"./start_it"

rejas:
Argh, you're right @sdetweil windows-npm-install doesnt traverse into the font/vendor dirctories.

Having scripts seems like the only viable solution. But I am not sure if @MichMich wants more files in the repo. 

But it would help for those who would want to develop on windows. So, he might say yes or no ;-) ?

khassel:
If @MichMich `s decision is to merge this, we should also update the documentation.

AFAIS there is no word about windows in the docs, so it should may mentioned [here](https://docs.magicmirror.builders/getting-started/requirements.html#operating-system) and also [here](https://docs.magicmirror.builders/getting-started/installation.html#manual-installation).

May there are other places but [searching does not work](https://github.com/MichMich/MagicMirror-Documentation/issues/91) at the moment.

Tests are failing here due to:
```bash
npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64
```

rejas:
Oh, this isnt ready to be merged with the windows install being not proper. Will convert it into a draft until we know how to proceed.

MichMich:
I'm a bit conflicted with this PR. A while ago we made the decision to remove any non Raspberry support because it is simply not maintainable. The problem of adding these features is that people will rely on them and expect from us we make sure it keeps working. I personally have no interest or intention to maintain any windows support. 

Since this PR does mainly add some convenience helpers it might be a better option to just add info to the docs. 

What do you guys think?

khassel:
> I personally have no interest or intention to maintain any windows support.

+1

new dependency `cross-env`:

`🚨 NOTICE: cross-env still works well, but is in maintenance mode. No new features will be added, only serious and common-case bugs will be fixed, and it will only be kept up-to-date with Node.js over time.`

I would have no problem with adding this so far but adding additional `bash`/`cmd` scripts to get the installation working on multiple platforms would be contrary to #1860

rejas:
> Since this PR does mainly add some convenience helpers it might be a better option to just add info to the docs.

Funny, I looked in the docs and found something regarding this already in there: https://docs.magicmirror.builders/getting-started/installation.html#common-installation-issues

Maybe this section needs to be clearer about why it is there (non.raspberry installation)?

rejas:
Created a PR in the MM Docs repo. Maybe that might help?

sdetweil:
but the doc IMPLIES that you SUPPORT running on Windows, which you "don't",  so I would remove the doc too ..

you have to make up your mind.

rejas:
I didnt thought about that since stuff about other (not officially supported) installation methods is in the docs too

khassel:
> Maybe this section needs to be clearer about why it is there (non.raspberry installation)?

yes, I saw this section and obviously the windows workaround is described there but windows is not mentioned ...

MichMich:
I think the best thing is to close this issue and to improve the docs. I've applied for Agolia DocSearch to make searching the docs better.

rejas:
Closing in favor of https://github.com/MichMich/MagicMirror-Documentation/pull/92

Issue #2700, opened 2021-10-22
[Feature Request] Calendar module: Hide recurringCountTitleForSpecificEvents


jupadin:
Hi,

**TL;DR**
Given a calendar with birthday entries, not knowing the date of birth for each entry, I want to hide the recurring count title.
*In case you don't like it due to the given reasons, just close this issue and related PR.*

According to post in https://github.com/MichMich/MagicMirror/issues/2639#issuecomment-909013378, I thought about a solution to this and came up with the following:

**Use case**:
You want to have a birthday calendar, based on the default calendar module, because it is such a great module and has a lot of functionality \o/.
**Problem:**
I do not know the year of birth for all of my entries, which would result in having a recurring count title like seen here (knowing the year of birth for the event _My birthday_ and not knowing the year of birth for the event _Another birthday_ - just chose a random year since I have to specify one in the ics file...):
![Bildschirmfoto 2021-10-22 um 10 57 10](https://user-images.githubusercontent.com/84035374/138427459-6536a290-c2d0-498d-ab2a-46d1094e6748.png)

**Solution:**
In the config, you can specify a year (under the given option: `hideRepeatingCountTitleOnYear`, which default value is `false`, such that it does not harm anything on default), for which the recurringCountTitle will be hidden.
If this config parameter is set, the situation is as follows (since I specified that the recurring count title for all events having a start year of `1920` should be hidden):
![Bildschirmfoto 2021-10-22 um 10 57 44](https://user-images.githubusercontent.com/84035374/138427673-6d2ea724-6e6f-411e-b37a-6c0b55f8eb92.png)

**Advantage:**
I can specify all my events with a given year of birth (or call it starting year), for which the recurring event title is hidden and I don't have to use a third party module, do some hacky things in the configuration for each individual calendar url or have to use two separate calendar ics-files (one for knowing the year of birth and one for the other case) - with btw. would lead to copying one event to the other file when knowing the year of birth.

I would appreciate your thoughts and ideas on this :-)

PR #2701, opened 2021-10-22
Proposal to hide the recurring count title in the calendar module

jupadin:
This PR is a first idea to implement the feature in #2700.

Done.

>

Done (In the issue #2700).

>    style issues are fixed.

Done.

> 4. Don't forget to add an entry about your changes to
>    the CHANGELOG.md file.

Done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2022-01-01._
 
 ### Added
 
+- Functionality to hide the repeatingCountTitle from the calendar module for events starting at a specific year.
+
 ### Updated
 
 - ESLint version supports now ECMAScript 2018
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -17,6 +17,7 @@ Module.register("calendar", {
 		showLocation: false,
 		displayRepeatingCountTitle: false,
 		defaultRepeatingCountTitle: "",
+		hideRepeatingCountTitleOnYear: false,
 		maxTitleLength: 25,
 		maxLocationTitleLength: 25,
 		wrapEvents: false, // wrap events to multiple lines breaking at maxTitleLength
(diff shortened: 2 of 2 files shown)


MichMich:
I think this PR still needs some thought. It is such an edgecase that we might want to rethink wether or not we want to bloat the current module. Maybe it makes more sense to built a nice (third party) birthday calendar module?

jupadin:
The definition of edgecase depends on the point of view I would say.
Regarding bloating the module I would say that adding this small adaption, wouldn't be that much change - since "developing" a 3rd party module would be 99,9% similar to the calendar module, since it brings everything with it already.
This modification can also be used as an anniversary calendar for yearly recurring wedding days, name days, jour fixes and sprints.

In the course of this, I also noticed that there is a bug in the calculation of times regarding all-day events, as the time from 2007 is used here to calculate the summer and winter time changeover (DST), instead of the time of the current year - these times change from year to year.
That would probably be worth another PR, though.

I would be happy in any case about this functionality in the Calendar module, because I use this currently already very much - If the PR would be merged, I would be happy but otherwise it's absolutely understandable - I would add it then just locally and if there is still someone who faces the same problem I'm happy to share the modifications.

MichMich:
Maybe a bit cleaner solution is the option to set a maximum for the reoccurring count. So let's say you set it to 100, everything higher then 100 will be hidden. That way we don't need to specify an arbitrary year.

MichMich:
If you are okay with that, please send a new pr. Closing this one for now.

jupadin:
> Maybe a bit cleaner solution is the option to set a maximum for the reoccurring count. So let's say you set it to 100, everything higher then 100 will be hidden. That way we don't need to specify an arbitrary year.

What do you mean by "higher than 100 will be hidden" ?
I want to hide the recurring count title when I don't know the exact starting year such that I can specify a year (e.g., birthday or wedding dates - whiteout knowing the exact year) such that there will no recurring title will be displayed for exactly those events - all other (even if higher or lower a magic number like 100) will be displayed..

jupadin:
Would be happy to provide a new PR but currently the requirements aren't that clear for me, sorry!! ...

MichMich:
Let say you set 100 as max, in case the year for that event is 1900, the reoccurring count would be 121 and therefor be hidden.

jupadin:
Ah, now I got it - thanks!
Gonna prepare another PR next week.

PR #2704, opened 2021-10-24
Relative full day events

### LINKED ISSUE #2424: Calendar module enhancement: displaying full day events without time

AmpioRosso:
Fixes #2424
- Don't display start time of full date events when using relative dates.
- Use the fullDateEventDateFormat when showing full day events more than a week out

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Fixed
 
 - Fixed error when accessing letsencrypt certificates
+- Fixed Calendar module enhancement: displaying full events without time (#2424)
 
 ## [2.17.0] - 2021-10-01
 
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -370,7 +370,7 @@ Module.register("calendar", {
 					// Show relative times
 					if (event.startDate >= now) {
 						// Use relative  time
-						if (!this.config.hideTime) {
+						if ((!this.config.hideTime) && (!event.fullDayEvent)) {
 							timeWrapper.innerHTML = this.capFirst(moment(event.startDate, "x").calendar(null, { sameElse: this.config.dateFormat }));
 						} else {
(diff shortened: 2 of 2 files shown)


josselynL:
Is this PR will be implemented ? :) Thank you.

MichMich:
The tests are failing. Not sure why though.

rejas:
Maybe after a merge from develop those tests run fine?

khassel:
the tests failed because they run before my latest test-fix, so I think thats unrelated to this PR, but test details are not available anymore. May @MichMich can restart the tests? Otherwise @AmpioRosso could add a fake commit (e.g. with a minimal change in changelog) to start a new pipeline.

khassel:
my first answer in not correct, after I read @rejas answer again the actual `develop` branch must be merged to get the fixed tests ...

rejas:
Merging develop into this PR does help as can be seen in #2737

Maybe @AmpioRosso can do this here or @MichMich merges the other one :-)

MichMich:
Merged https://github.com/MichMich/MagicMirror/pull/2737

PR #2716, opened 2021-11-28
Revert node ical update

khassel:
With my last dependency update node-ical was updated from 0.13.0 to 0.14.1.

Starting mm with the calendar module results now in `module luxon not found` when using `npm install --no-optional`.

So reverting this, must be solved on the node-ical side.


rejas (approved):

khassel:
changelog check fails but this time I didn't forget it ...

PR #2736, opened 2021-12-21
Allow dangerouslyDisablingAutoEscaping in newsfeed items

### LINKED ISSUE #2712: Newsfeed module cannot show description right. Wrong HTML handling?
Hi together,
somehow, the newsfeed module cannot handle the RSS feed right. I tried different sources but every time the same at my MagicMirror:

image

Thank you in advance.

rejas:
Sometimes the newfeed might encode html tags and such (german umlauts for eaxmple) which the nunjuck templating would escape again so that the text looks like garbage.

This is helpful in preventing malicious code to be send to the mirror but the user might now disable this behaviour with the new config option "dangerouslyDisableAutoEscaping"

Fixes #2712


MichMich:
Love the variableName. 😂

PR #2745, opened 2021-12-22
Add missing shebang

KristjanESPERANTO:
I noticed the missing shebang when I checked it with ShellCheck. I don't have an old installation to test, but the change shouldn't cause any problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ _This release is scheduled to be released on 2022-01-01._
 - Added lithuanian language to translations.js
 - Show info message if newsfeed is empty (fixes #2731)
 - Added dangerouslyDisableAutoEscaping config option for newsfeed templates (fixes #2712)
+- Added missing shebang to `installers/mm.sh`.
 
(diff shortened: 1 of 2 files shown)


khassel:
test failure is unrelated to this PR, will look into this within the next days ...

@MichMich because of your comment in the changed file `# This file is still here to keep PM2 working on older installations.`

Is the file still needed? What is meant with "older installations"?

MichMich:
I've got no idea. Can't recall it. @sdetweil knows a lot about PM2. Maybe he can shine a light on this file.

sdetweil:
pm2 won't launch npm start directly
so mm.sh is used, but directly launched with sh, not using the shebang approach to relaunch

pm2 needs the .sh extension to know how to launch. , maybe that's the missing shebang problem..

khassel:
so if I understand @sdetweil  correctly the file is still needed and then this should me merged.

sdetweil:
yes, correct on both points

rejas (approved):

Issue #2740, opened 2021-12-21
update templates and documentation concerning node versions


khassel:
I found following relevant mentioned node versions in this repo:
- https://github.com/MichMich/MagicMirror/blob/5a4fbbf48a55666fb851d75b5ce8a19dd2eee874/.github/workflows/automated-tests.yml#L18
- https://github.com/MichMich/MagicMirror/blob/5a4fbbf48a55666fb851d75b5ce8a19dd2eee874/.github/CONTRIBUTING.md#L35
- https://github.com/MichMich/MagicMirror/blob/5a4fbbf48a55666fb851d75b5ce8a19dd2eee874/.github/ISSUE_TEMPLATE.md#L36

In the documentation repo:
- https://github.com/MichMich/MagicMirror-Documentation/blob/07199539b7a42ce5b94eb4f59524e7de18e9465b/getting-started/requirements.md#L20

Since 26. Oct. Node v16 is the active [release of Node](https://nodejs.org/en/about/releases/). End of life of v12 is Apr. 21.

So I think we should update the above docs to:
- min. version v14
- recommended version v16
- automated tests to use v14, v16, v17

Should be done before next release so that these infos become available on `master` next year.

Want to discuss this here before making a PR.

I found no problems so far with v16/v17, I'm using v16 in my `latest` docker images since nearly 2 month and v17 in my `develop` images (which I use running my own mm).

PR #2746, opened 2021-12-22
node version update in templates and github workflows

khassel:
see #2740

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 17.x]
     steps:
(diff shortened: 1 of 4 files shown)


khassel:
failed test unrelated to this PR, hopefully fixed in the other one ...

.github/workflows/automated-tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 17.x]
```
rejas, line 18: so once v18 is released we switch to that instead of v17 right?

rejas (approved):

.github/workflows/automated-tests.yml
khassel: yes, release date 2022-04-19, we can switch everytime because this is only important on `develop`.

PR #2750, opened 2021-12-25
Upgrade to node-fetch 3.x

KristjanESPERANTO:
This is an attempt to tackle issue #2649.

Since `node-fetch` is [converted into an ES module](https://github.com/node-fetch/node-fetch/blob/HEAD/docs/v3-UPGRADE-GUIDE.md), a simple require statement no longer works.


KristjanESPERANTO:
Since Node 18 will probably get native fetch support, I think that we don't need to continue this topic any further. See #2649

PR #2754, opened 2021-12-26
Broadcast custom symbols in CALENDAR_EVENTS

kolbyjack:
I added support for displaying event symbols to MMM-MonthlyCalendar, and received a bug report that custom symbols aren't being displayed.  Move the custom event logic into symbolsForEvent so that they are properly set in the CALENDAR_EVENTS message.

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -237,18 +237,6 @@ Module.register("calendar", {
 				symbolWrapper.className = "symbol align-right " + symbolClass;
 
 				const symbols = this.symbolsForEvent(event);
-				// If symbols are displayed and custom symbol is set, replace event symbol
-				if (this.config.displaySymbol && this.config.customEvents.length > 0) {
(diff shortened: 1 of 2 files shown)


modules/default/calendar/calendar.js
```diff
@@ -237,18 +237,6 @@ Module.register("calendar", {
 				symbolWrapper.className = "symbol align-right " + symbolClass;
 
 				const symbols = this.symbolsForEvent(event);
-				// If symbols are displayed and custom symbol is set, replace event symbol
-				if (this.config.displaySymbol && this.config.customEvents.length > 0) {
```
rejas, line 241: This check isnt in the added code. Dont we need it there?

kolbyjack: I don't believe so, the check was redundant in the old location, it was already inside an if (this.config.displaySymbol), and the other call should have the custom symbol information even if symbols aren't being displayed, since it's populating the notification message

rejas (approved):

Issue #2749, opened 2021-12-25
is there any doc on useHttps?  I can't find any


sdetweil:
user trying to set it up... but not referenced in config docs

and the config sample 

path to  implies, NOT filename

sdetweil:
this post  says the code is broken 
https://forum.magicmirror.builders/topic/14091/issues-enabling-https-on-rp-4?_=1640398160931
```
useHttps: false -> useHttps: true

Nevertheless I only got a black screen afterwards.

I tried everything I could think of, but, being that Im not much of a programmer, it basically meant hitting the keyboard with one hand while having another coffee hoping that would do it.

Eventually, I managed to find out that I could connect to the https://localhost:port via a browser, which made me realise the keyboard-hitting works… and that it was running… just not with electron.

Well, tonight, after many an hour in front of this, I was able to get it to work, so I can narrow it down to the following:

Create your certificates correctly
Make sure you put the right address where they are located
Not enough coffee
There is a piece of code in electron.js that seems to be checking the wrong variable:
// if (config[“tls”] !== null && config[“tls”]) {
if (config.useHttps) {
prefix = “https://”;
} else {
prefix = “http://”;
}

The line in “//” was commented out by me. I couldnt find much of a reference to tls anywhere.

Finally, as the certificate is self signed, these lines had to be added also to electron.js:

app.on(‘certificate-error’, (event, webContents, url, error, certificate,
callback) => {
event.preventDefault();
callback(true);
});

```

I see electron.js in develop is still checking config['tls'], not config.useHttps

khassel:
yes, this is a mess.

AFAIK there were 2 use cases implemented using `https` so far:

- option `useHttps` for `serveronly` mode: parameter `useHttps` and certs in `config.js`
- option `tls` for `clientonly` mode: runtime parameter `--use-tls` needed

But I never understood any of the https use cases. Why encrypt things when the whole config per default is public and so readable by erveryone (including config secrets)? You only have to call `http(s)://<ip-of-mm>:<port-of-mm>/config/`.

So if anyone wants to have a public MagicMirror-Url he should protect this url with some sort of authentification and therefore simplest solution is a reverse proxy which also can handle the https part out of the box.

sdetweil:
apparently home assistant in https mode cannot access mm in http mode

khassel:
to get this running without breaking the existing code we could change the `if` statement in `electron.js` to

```js
[...]

PR #2756, opened 2021-12-26
enable useHttps for full electron version

sdetweil:
Fixes: #2749 
updated  CHANGELOG.md file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,9 @@ _This release is scheduled to be released on 2022-01-01._
 - Fixed User-Agent-Header for newsfeed and calendar module (#2729).
 - Replace broken shields in Readme and use https for links.
 - Fixed electron tests with retry.
-- Fix useHttp for full version #2749
+- Fixed Calendar recurring cross timezone error (add/subtract a day, not just offset hours) (#2632)
+- Fixed Calendar showEnd and Full Date overlay (#2629)
+- Fix useHttps for full version (#2749)
 
 ## [2.17.1] - 2021-10-01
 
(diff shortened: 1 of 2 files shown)


sdetweil:
what does the failing codecov/patch test mean?

khassel:
this means nothing, looks like adding code lines without adding tests produces this, will be merged anyway ...

khassel (approved):

sdetweil:
how i tested

with the ~/MagicMirror folder as the current folder
i did this 
https://stackoverflow.com/questions/10175812/how-to-generate-a-self-signed-ssl-certificate-using-openssl
search for "You can do that in one command:"
adding -nodes at the end of the listed command

I completed  all the fields except section, which I let default to  blank ([]),
I moved the generated key.pem and cert.pem files to a new folder ~/MagicMirror/keys
I added those pem files to the confg.js, key and cert, and changed useHttps to true
```
    useHttps: true,         // Support HTTPS or not, default "false" will use HTTP
    httpsPrivateKey: "keys/key.pem",     // HTTPS private key path, only require when useHttps is true
    httpsCertificate: "keys/cert.pem",     // HTTPS Certificate path, only require when useHttps is true
```
I started MM, and used chrome to access the pi at https://localhost/:mm_port (as configured in config.js)// note  this is not https://localhost/ (using default port 443)
I clicked advanced at the warning and then proceed on the next page

MM came up...  I was using a local ICS file, which failed to load as it was http:
I changed to https and then added 
                selfSignedCert:true,  // <- this
                calendars: [

to the calendar module config

and restarted MM, and it came up, and was able to access the calendar ics file

rejas (approved):

Issue #2742, opened 2021-12-22
[Feature Request] Sync the Newsfeed `updateInterval` between separately configured modules


winstonma:
The `updateInterval` in [newsfeed](https://github.com/MichMich/MagicMirror/blob/master/modules/default/newsfeed/newsfeed.js) works perfectly. However the problem arise when there are multiple modules running on the screen and the e-ink screen is being used.

Current implementation:
```
	scheduleUpdateInterval: function () {
		this.updateDom(this.config.animationSpeed);

		// Broadcast NewsFeed if needed
		if (this.config.broadcastNewsFeeds) {
			this.sendNotification("NEWS_FEED", { items: this.newsItems });
		}

		this.timer = setInterval(() => {
			this.activeItem++;
			this.updateDom(this.config.animationSpeed);

			// Broadcast NewsFeed if needed
			if (this.config.broadcastNewsFeeds) {
				this.sendNotification("NEWS_FEED", { items: this.newsItems });
			}
		}, this.config.updateInterval);
	},
```

I tried this:
```
	scheduleUpdateInterval: function () {
		const nextLoad = this.config.updateInterval - (moment().currentTime.valueOf() % this.config.updateInterval);
		this.updateDom(this.config.animationSpeed);

		// Broadcast NewsFeed if needed
		if (this.config.broadcastNewsFeeds) {
			this.sendNotification("NEWS_FEED", { items: this.newsItems });
		}

		setTimeout(() => {
			this.activeItem++;
			this.scheduleUpdateInterval();
		}, nextLoad);
	},
```

This would ensure that the modules are updating in the same time, even the modules are configured separately. And it works best on the e-ink MagicMirror devices because they did less refresh.

rejas:
This sounds reasonable, would you like to create a PR for this feature since you already have some code? Or do you need help with that?

sdetweil:
but then u overload the network all at once with concurrent fetches...  which may impact other modules as well. 
i think the objective is a bad idea.,   

I think setTimeout should be used instead of setInterval, regardless(because if u run off the end of the interval time while processing, oops, another cycles starts, etc,etc,etc..) , but you will still end up with different cycles.  because the content is likely different between the two instances.

rejas:
> but then u overload the network all at once with concurrent fetches... which may impact other modules as well.

how about a config option for this behaviour then? something like "syncFeedUpdates" with the default being false (would be the current behaviour)? and in the docs the possible network problems could be mentioned?

[...]

PR #2758, opened 2021-12-28
Replace setInterval with setTimeout to ensure that multiple modules with same updateInterval would be updated at the same time

winstonma:
The new implementation method uses Unix Time as a reference to sync the visual updates between separately configured modules. Instead of using `setInterval`, which causes random update time between modules, the modification uses `setTimeout` with dynamic timeout period force all modules to execute the scheduled action at a commonly defined time.

It is very useful when e-ink display device is used because it can ensure that all modules would update at the same time. So the e-ink display would not be so flashy.

#2742 is linked with this


MichMich:
Closed. (See issue for explanation)

PR #2764, opened 2022-01-02
Add absolute dates to weather forecast

oraclesean:
No

-  If so, can you reference the issue like this `Fixes #<issue_number>`? 
N/A

- What does the pull request accomplish? Use a list if needed.
Add a Boolean configuration option `absoluteDates` to `modules/default/weather` to allow weather forecast dates to be formatted as absolute (MON, TUE) or relative (TODAY, TOMORROW).

- If it includes major visual changes please add screenshots.
N/A

### DIFF
--- a/modules/default/weather/forecast.njk
+++ b/modules/default/weather/forecast.njk
@@ -8,9 +8,9 @@
         {% set forecast = forecast.slice(0, numSteps) %}
         {% for f in forecast %}
             <tr {% if config.colored %}class="colored"{% endif %} {% if config.fade %}style="opacity: {{ currentStep | opacity(numSteps) }};"{% endif %}>
-                {% if (currentStep == 0) and config.ignoreToday == false %}
+                {% if (currentStep == 0) and config.ignoreToday == false and config.absoluteDates == false %}
                     <td class="day">{{ "TODAY" | translate }}</td>
-                {% elif (currentStep == 1) and config.ignoreToday == false %}
+                {% elif (currentStep == 1) and config.ignoreToday == false and config.absoluteDates == false %}
                     <td class="day">{{ "TOMORROW" | translate }}</td>
                 {% else %}
                     <td class="day">{{ f.date.format('ddd') }}</td>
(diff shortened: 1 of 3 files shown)


modules/default/weather/forecast.njk
```diff
@@ -8,9 +8,9 @@
         {% set forecast = forecast.slice(0, numSteps) %}
         {% for f in forecast %}
             <tr {% if config.colored %}class="colored"{% endif %} {% if config.fade %}style="opacity: {{ currentStep | opacity(numSteps) }};"{% endif %}>
-                {% if (currentStep == 0) and config.ignoreToday == false %}
+                {% if (currentStep == 0) and config.ignoreToday == false and config.absoluteDates == false %}
```
rejas, line 11: this config option should also be set in the defaults of the weather.js module

MichMich: Indeed!

MichMich:
Nice idea! @rejas made a valid point. 

Also: since this is a new feature, don't forget to send a PR to the docs repo after merge. :)

oraclesean (commented):
Change made to weather.js, thanks for catching that!

rejas:
Nice. Now it would be awesome if you could add a testcase for that. Think you can do it or do you need help?

oraclesean:
@rejas yes I'd appreciate some guidance for a test case. JS is not my ninja area 😄

modules/default/weather/weather.js
```diff
@@ -43,7 +43,8 @@ Module.register("weather", {
 		onlyTemp: false,
 		showPrecipitationAmount: false,
 		colored: false,
-		showFeelsLike: true
+		showFeelsLike: true,
+		absoluteDates: false
```
khassel, line 47: if this is a property for the forecast, shouldn't this inserted in `weatherforecast.js` instead?

oraclesean: I thought the current and forecast modules were deprecated in favor of the standalone weather module.

khassel: ups, sorry for the noise, was irritated when looking in the tests (theoretically I should know this ...)

MichMich: 😅👍🏻

khassel:
wrote a test for this, will provide a PR when this is merged.

MichMich:
Done!

PR #2774, opened 2022-01-08
remove old weather modules

khassel:
`currentweather` and `weatherforecast` were deprecated with `v2.15` so I think we should remove them with the next release which is one year later.

This is still a draft to get opinions on whether this is okay.


rejas:
> It might be an idea to temporary (one year again?) replace them with dummy modules which show an error/explanation.

Would definitly help those users that still run the modules

khassel:
The 2 modules are now replaced with fake modules, see screenshot:

grafik

MichMich:
Could you please make the docs link white. I have an unreasonable hate for purple and blue links. 

Other than that. Awesome work! :)

rejas:
> Could you please make the docs link white. I have an unreasonable hate for purple and blue links.
> 
> Other than that. Awesome work! :)

I agree with both statements! :-)

rejas (approved):

khassel:
done:

grafik

MichMich:
Thanks. Great work.

PR #2780, opened 2022-01-11
automatically add/remove a hidden/shown class to the module wrappers

Tom-Hirschberger:
Hi,

i need a possibility to change the z-index of modules based on the information if they are hidden or shown currently. In my setup i do have overlapping modules which are hidden/shown based on profiles. The modules are controlled by touch.
I added two small changes to the show/hide function which add/remove the classes shown/hidden to the module wrappers.
This way i can control which modules need to handle the current events by setting different z-index values in my custom.css file.

```
.module.MMM-Bring.shown,
.module.MMM-SynologySurveillance.shown {
  z-index: 1000;
}

.module.MMM-Bring.hidden,
.module.MMM-SynologySurveillance.hidden {
  z-index: -1;
}
```

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -245,7 +245,6 @@ const MM = (function () {
 		if (moduleWrapper !== null) {
 			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			moduleWrapper.style.opacity = 0;
-			moduleWrapper.className = moduleWrapper.className.split(" shown").join("");
 			moduleWrapper.className += " hidden";
 
 			clearTimeout(module.showHideTimer);
@@ -313,7 +312,6 @@ const MM = (function () {
 			// Restore the position. See hideModule() for more info.
 			moduleWrapper.style.position = "static";
 			moduleWrapper.className = moduleWrapper.className.split(" hidden").join("");
-			moduleWrapper.className += " shown";
 
 			updateWrapperStates();
 
--- a/js/main.js
+++ b/js/main.js
@@ -245,6 +245,8 @@ const MM = (function () {
 		if (moduleWrapper !== null) {
 			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			moduleWrapper.style.opacity = 0;
+			moduleWrapper.className = moduleWrapper.className.split(" shown").join("");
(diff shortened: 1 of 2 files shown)


MichMich:
Hi, thanks for your contribution. 

Isn't this something your module can do itself? I don't think a change to the core is nessecery for this. 

What am I missing here?

Tom-Hirschberger:
Hi, basically i can implement in my modules. Some modules like MMM-CalenderExt2 implement it already. But in my opinion it is much easier to implement it in the core and do not need to touch the code of every touch enabled module i use in my setup.

Is there any chance to the detect the hidden state in css without my change?

MichMich:
I've got no problem adding a class `hidden` to any hidden module. But the details of that class can better reside in your own module. So the core adds and removed the hidden class. You module configures the z index for those classes.

Is that a workable option?

MichMich:
Sorry, I just noticed that that is exactly what your code does. 🤦

MichMich:
But I think we can leave out the "shown" class. You can use the `not()` selector to select any module that doesn't have a hidden class.

Tom-Hirschberger:
Sounds great to me. I will remove the shown part

js/main.js
```diff
@@ -245,6 +245,8 @@ const MM = (function () {
 		if (moduleWrapper !== null) {
 			moduleWrapper.style.transition = "opacity " + speed / 1000 + "s";
 			moduleWrapper.style.opacity = 0;
+			moduleWrapper.className = moduleWrapper.className.split(" shown").join("");
+			moduleWrapper.className += " hidden";
```
MichMich, line 249: I think it might be better to use `classList.add("hidden")` and `classList.remove("hidden")` in stead of doing string manipulation.

MichMich:
Great! Bonus points if you add a test for this new feature :)

js/main.js
Tom-Hirschberger: did not know this functions. Changed to them now

Tom-Hirschberger:
I will try to write an test but currently i do not have a glue how to do that. Did not write an JavaScript test till now

MichMich:
Great moment to start. ;)

PR #2784, opened 2022-01-13
improved and speedup e2e tests

khassel:
In the e2e tests I used an artifical wait after mm start because the elements where loaded in jsdom but still undefined when accessing them. Now I managed to wait for the element so I could remove this artifical wait. So now the e2e tests are faster and hopefully there are no fails anymore (in the past most fails were because the artifical wait was to low).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ _This release is scheduled to be released on 2022-04-01._
 
 ### Fixed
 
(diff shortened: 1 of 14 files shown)


MichMich:
If this works, this would be the greatest PR of all times. 😬👍🏻

rejas:
37s vs 228s on my machine, very impressive @khassel ! great work !!!

PR #2787, opened 2022-01-17
Return full list of events for broadcasting

kolbyjack:
I have several modules that provide alternate displays for calendar events, and currently, if I want the calendar module to return all of the future events for the current month (for my MMM-MonthlyCalendar module), then I need to configure maximumEvents to be large enough that the calendar module itself needs to be hidden.  Instead, the configured display limits should only be imposed when generating the list of events to display, not when fetching events from configured calendars.


MichMich:
I understand the need and it's fine to implement. But I'm not a big fan of the `forDisplay` attribute since it doesn't tell what will happen if it's set to true. Maybe rename it to something like `limitNumberOfEntries`?

khassel:
@MichMich  the test fail is unrelated to the code changes, same test ran locally on my machine and in all tests except node 14 (by now I think the vm running the tests was extremly slow but if this occurs again we have to increase timeout).

MichMich:
I'll rerun the jobs to check if that helps.

PR #2788, opened 2022-01-17
change husky setup

khassel:
This is a convenience PR for developers when working with this git repo (or a fork) without having dependencies installed.

I'm doing this e.g. when adding a changelog entry on my windows machine where I don't want to install the whole node/npm stuff.

So I changed the husky pre-commit to run the commands inside only if they are installed.

Hopefully @rejas can live with this change ...

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,7 @@
 #!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
 
-npm run lint:staged
+[ -f "$(dirname "$0")/_/husky.sh" ] && . "$(dirname "$0")/_/husky.sh"
+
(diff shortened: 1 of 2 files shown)


sdetweil:
note that the scripted install does not install the dev dependencies, so they will not be there,
how does this affect first time users trying to create a pr?

khassel:
current situation is, trying to commit something without dev dependencies installed will fail.

With this PR this will work. The pre-commit runs normally `npm run lint:staged` so this is may not executed with this PR but checked later running the tests.

So its easer to contribute with this PR ...

sdetweil:
ok, thx, now understand.. i support your idea and changes

MichMich:
Nice one. Thanks!

PR #2794, opened 2022-01-22
Newsfeed module: Add option to display title as url

khassel:
Added new config option `showTitleAsUrl` to newsfeed module.

If set, the displayed title is a link to the article which is useful when running in a browser. If you now want to read the displayed article you can just click on it and it will open in a new browser tab/window.

There are no optical changes, except when `showTitleAsUrl` is set a mouse curser is displayed when moving over the title so you can click.

Will update documentation if this is accepted and merged here.


khassel:
btw I love this Codecov Report, its claiming about a decrease of coverage in `js/server.js` but this file is not changed in this PR ...

rejas (approved):

PR #2810, opened 2022-02-09
fix parsing of feels like for weathergov

philnagel:
weather.gov API does not seem to consistently provide a feels like value. The [API specs](https://weather-gov.github.io/api/gridpoints) mention an `apparentTemperature` value, which seems to not be provided most of the time in responses. There is a consistent value of `windChill` and `heatIndex`, which according to docs, only has a value when the value falls within a certain threshold.

Before this fix, feels like would only be read from `heatIndex` and it would be coerced to 0 if the value was `null`, thus showing a consistent feels like temp of 0 celsius/ 32 fahrenheit.

This pull request uses either `windChill` or `heatIndex` for feels like temp if one of them is available. If none are available, it will fall back to the current actual temperature.

### DIFF
--- a/modules/default/weather/providers/weathergov.js
+++ b/modules/default/weather/providers/weathergov.js
@@ -156,8 +156,13 @@ WeatherProvider.register("weathergov", {
 		currentWeather.rain = null;
 		currentWeather.snow = null;
 		currentWeather.precipitation = this.convertLength(currentWeatherData.precipitationLastHour.value);
-		currentWeather.feelsLikeTemp = this.convertTemp(currentWeatherData.heatIndex.value);
-
+		if (currentWeatherData.heatIndex.value !== null) {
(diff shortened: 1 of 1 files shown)


rejas:
Thanks for the PR.

Might it be usefull to check first for apparentTemperature so that IF it is provided it is used for the feelsLike value? And if not, then the other values are taken into account?

philnagel:
I think I was actually mistaken with the link I included for the reference. That one only refers to the `gridpoints` endpoint, but in this case for the current conditions, we're using the `observation` endpoint. The schema for that is found [here ](https://www.weather.gov/documentation/services-web-api#/default/station_observation_latest)and does not reference `apparentTemperature` at all, only `heatIndex` and `windChill`. 
I think based on that, the logic should be sufficient as is. Let me know your thoughts.

MichMich:
Thanks for your contribution! My apologies for my slow response. Life got in the way.

PR #2814, opened 2022-02-20
Fix log output

KristjanESPERANTO:
Console output before:
```
alert - Load translationfalse: translations/de.json
alert - Load translation fallback: translations/bg.json
```

Console output after:
```
alert - Load translation: translations/de.json
alert - Load translation fallback: translations/bg.json
```


rejas (approved):
Good catch!

PR #2818, opened 2022-03-03
Don't adjust startDate for full day events if endDate is in the past

kolbyjack:
When including past events in the event list for broadcast, the adjustment to make multi-day events appear as happening today had the side effect of promoting single-day full-day events from yesterday to today.  Adding a check to ensure that the event hasn't ended before updating the start date fixes my use case, however, I think a proper fix needs to perform the change in calendar.js when building the dom, as it is, multi-day events will still have their start date updated until after they've ended, which could be confusing for CALENDAR_EVENTS consumers.  Alternatively, a second field (`originalStartDate`?) could be added for broadcasting, but that seems like a hack.


MichMich:
Thanks for your contribution! My apologies for my slow response. Life got in the way.

PR #2822, opened 2022-03-16
add default `User-Agent=MagicMirror` when running as electron

khassel:
add default `User-Agent=MagicMirror` when running as electron because without a proper header some url's are not accepted. The default can be overridden in `config.js`. This solves problems with openstreetmap tiles urls as e.g. [discussed here](https://forum.magicmirror.builders/topic/12808/mmm-rain-map-new-version-2-x/153) and with @jalibu on discord.

With this change everyone can set his own header to whatever needed in `config.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ _This release is scheduled to be released on 2022-04-01._
 - Added new config option `showTitleAsUrl` to newsfeed module. If set, the diplayed title is a link to the article which is useful when running in a browser and you want to read this article.
 - Added internal cors proxy to get weather providers working without public proxies (fixes #2714). The new url `http(s)://address:port/cors?url=https://whatever-to-proxy` can be used in other modules too.
 - Added a WeatherProvider for Weatherflow
+- add default `User-Agent=MagicMirror` when running as electron because without a proper header some url's are not accepted. The default can be overridden in `config.js`. This solves problems with openstreetmap tiles urls as e.g. [discussed here](https://forum.magicmirror.builders/topic/12808/mmm-rain-map-new-version-2-x/153).
 
 ### Updated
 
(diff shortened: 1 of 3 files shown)


rejas:
Didnt there used to be a User-Agent already set which got removed?

khassel:
> Didnt there used to be a User-Agent already set which got removed?

yes, but I don't know exactly the content, found e.g. [this](https://user-agents.net/applications/electron). When there is a need to keep the old, we need a different approach.

sdetweil:
well, this assumes that the User-Agent value is valid for ALL module servers.. 

otherwise it should be one of the top 4. Chrome, FireFox, Safari or Edge.

khassel:
the current default User-Agent is `Mozilla/5.0 (X11; Linux armv7l) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.109 Electron/17.1.2 Safari/537.36`.

The problematically part concerning openstreetmap is `(X11; Linux armv7l)`, will convert this PR to draft until we found a solid solution.

MichMich:
I remember vaguely that it was MagicMirror in the past, but that gave some issues. So we switched to a default (chrome) to prevent issues.

khassel:
@jalibu we cannot merge this so my question is if we close this unmerged or without the default property which means every user must actively set the User-Agent in his config.js if needed (e.g. for the openstreetmap.org problem).

jalibu:
@khassel what's your question to me? :-)

khassel:
@jalibu I created this PR to solve the problem with the openstreetmap.org tile url. But setting the User-Agent to a new default will not work (see above comments). So we have 2 options:

- close this unmerged and use another osm tile server (e.g. the `.de`)
- don't set a default header in `default.js` and only set `mainWindow.webContents.userAgent = config.electronUserAgentHeader` if provided in `config.js`

With the second option the openstreetmap.org tile url would work only if every user sets a working `electronUserAgentHeader` in `config.js` so I tend to option one. Meanwhile I think the openstreetmap.org tile server is misconfigured because it blocks valid headers.

PR #2827, opened 2022-03-25
toggle pointer-events with suspension

shin10:
Hi,

I added two classes for the hidden/visible modules to toggle the pointer-events with the new `.hidden` class.
Further I removed the one with the universal selector, which made hidden modules and just anything in them stealing pointer events, if they were placed in any fullscreen area.


rejas (approved):
Thanks, looks ok although I cannot test it myself. You will need to add a CHANGELOG entry though before it gets merged.

shin10:
Sure, thanks.
By the way. I'm currently creating some modules myself. Have to do some refactoring today but I'd like to contribute it to the 3rd party list. Whom might I contact for reviewing them?

MichMich:
You can just add them to the wiki. It's open for everybody to edit.

Issue #2831, opened 2022-03-28
electron app.disableHardwareAcceleration()


khassel:
Electron is throwing gpu errors under some circumstances e.g. running in a docker container as mentioned [here](https://github.com/MichMich/MagicMirror/issues/2656#issuecomment-981160715) and there were already some threads in the forum about this.

To disable these errors you can add `app.disableHardwareAcceleration()` in `electron.js` or `--disable-gpu` in `package.json`.

Both variants are ugly because for automating this you can only use `sed` or similar tools.

So I would like to add an environment variable `ELECTRON_DISABLE_GPU` and add `app.disableHardwareAcceleration()` in `electron.js` only when set.

More info's to disableHardwareAcceleration:
- https://www.electronjs.org/docs/latest/tutorial/offscreen-rendering
- https://www.electronjs.org/docs/latest/api/app#appdisablehardwareacceleration

May it is also an option to run always with `app.disableHardwareAcceleration()`?

shin10:
Have you considered to use
```js
electronSwitches: ["disable-gpu"],
```
in your `config.js`?

I feel your `sed` pain, since I have several hardware components attached to GPIO but am developing in WSL mostly. So I have to en/disable the components depending on the platform. To do this I'm running a make recipe which substitutes via `envsubst` variables sourced from an `.env` file before starting the app. That's ok-ish I guess. At least I can handle all my secrets that way, too.
It's work in progress but you can have a look at my conf [here](https://github.com/shin10/MagicMirror-config).

Anyway, I'm super new to MM so if this is a stupid idea, just ignore it. I'm open to suggestions as well.

khassel:
> electronSwitches: ["disable-gpu"],

yes, tested this with no success, `--disable-gpu` is missing in the [list of supported command line switches](https://www.electronjs.org/de/docs/latest/api/command-line-switches).

I'm already using the `envsubst` construction in the [docker entrypoint script](https://gitlab.com/khassel/magicmirror/-/blob/master/build/entrypoint.sh#L39) but thats no solution for this problem.

shin10:
Ah, I see. Well, it's been a while, but I once used software rendering for tests of pages with three.js in docker. Back then I used Chrome with OSMesa. There seems to be a similar approach for electron with SwiftShader https://medium.com/@mohamedmansour/how-to-run-electron-js-in-software-rendering-headless-49601b87961e
Maybe that's an alternative to disabling the GPU.

PR #2832, opened 2022-03-29
Add env var to disable gpu under electron when set

khassel:
see #2831

### DIFF
--- a/js/electron.js
+++ b/js/electron.js
@@ -8,6 +8,8 @@ const Log = require("logger");
 let config = process.env.config ? JSON.parse(process.env.config) : {};
 // Module to control application life.
 const app = electron.app;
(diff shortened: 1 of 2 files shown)


js/electron.js
```diff
@@ -8,6 +8,10 @@ const Log = require("logger");
 let config = process.env.config ? JSON.parse(process.env.config) : {};
 // Module to control application life.
 const app = electron.app;
+if (process.env.ELECTRON_DISABLE_GPU !== undefined) {
```
rejas, line 11: maybe add a little comment above this?

rejas (approved):

js/electron.js
khassel: done.

PR #2836, opened 2022-04-05
Fix problems with non latin fonds caused by updating to fontsource

### LINKED ISSUE #2835: Bad font for non latin users
## Fonts bug in MagicMirror

**Platform**: MagicMirrorOS on Raspberry Pi 2 .
**MagicMirror² Version**: 2.19.0

**Description**:

In this update fonts package was changed from `roboto-fontface-bower` to `fontsource` (2dcf55ea89f752390554d655ee819148c07e4fbc).
CSS file `roboto.css` uses only `latin` subset for loading woff files. It is not fair = fonts must consist all world chars.

**Way to fix**:

Change `roboto-latin-100-normal.woff2` to `roboto-all-100-normal.woff2` and etc.

**Expected Results**: Fonts are still great with non US characters.
**Actual Results**: Fonts are ugly. Need to add custom `@font-face` from Google.

khassel:
fixes #2835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ _This release is scheduled to be released on 2022-07-01._
 
 ### Fixed
 
+- Fix problems with non latin fonds caused by updating to fontsource (fixes #2835).
+
 ## [2.19.0] - 2022-04-01
 
 Special thanks to the following contributors: @10bias, @CFenner, @JHWelch, @k1rd3rf, @khassel, @kolbyjack, @krekos, @KristjanESPERANTO, @Nerfzooka, @oraclesean, @oscarb, @philnagel, @rejas, @sdetweil, @shin10, @SiderealArt and @Tom-Hirschberger.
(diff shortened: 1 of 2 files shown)


khassel:
Need feedback to this one. I had to substitue `-latin-` with `-all-` for fixing the non latin fonds.

But there are no `woff2` files for `-all-` so I removed the links to `woff2`. Reading about this `woff2` seems to be the preferred format. Any better ideas?

MichMich:
As long as it's supported by electron it seems fine by me.

khassel:
tested this with electron, but must be tested anyway by non-latin users ...

rejas:
During standard web-developmet, woff2 would indeed be preferred due to size improvements. But I guess here we can ignore that :-)

rejas (approved):

nikow13 (approved):

MichMich:
I merged this to see if this will cause issues. We can always revert if someone runs into issues.

PR #2842, opened 2022-04-26
Added outgoing notification of the weather module

eouia:
The current `weather` module seems no notification emission.

Added `WEATHER_UPDATED` notification when new weather information is provided.

When this PR is accepted, I'll modify wiki page of MM manual.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2022-07-01._
 
 ### Added
 
+- Added the notification emitting from the weather module on infromation updated.
+
 ### Updated
 
 ### Fixed
--- a/modules/default/weather/weather.js
+++ b/modules/default/weather/weather.js
@@ -156,11 +156,11 @@ Module.register("weather", {
 		}
 
 		let notificationPayload = {
-			current: this.weatherProvider?.currentWeatherObject?.simpleClone() ?? null,
-			forecast: this.weatherProvider?.weatherForecastArray?.map((ar) => ar.simpleClone()) ?? [],
-			hourly: this.weatherProvider?.weatherHourlyArray?.map((ar) => ar.simpleClone()) ?? [],
-			location: this.weatherProvider?.fetchedLocationName,
-			provider: this.weatherProvider.providerName
+			currentWeather: this.weatherProvider?.currentWeatherObject?.simpleClone() ?? null,
+			forecastArray: this.weatherProvider?.weatherForecastArray?.map((ar) => ar.simpleClone()) ?? [],
+			hourlyArray: this.weatherProvider?.weatherHourlyArray?.map((ar) => ar.simpleClone()) ?? [],
+			locationName: this.weatherProvider?.fetchedLocationName,
+			providerName: this.weatherProvider.providerName
 		};
 		this.sendNotification("WEATHER_UPDATED", notificationPayload);
 	},
--- a/modules/default/weather/weather.js
+++ b/modules/default/weather/weather.js
@@ -154,6 +154,15 @@ Module.register("weather", {
 		if (this.weatherProvider.currentWeather()) {
 			this.sendNotification("CURRENTWEATHER_TYPE", { type: this.weatherProvider.currentWeather().weatherType.replace("-", "_") });
 		}
+
+		let notificationPayload = {
+			current: this.weatherProvider?.currentWeatherObject?.simpleClone() ?? null,
+			forecast: this.weatherProvider?.weatherForecastArray?.map((ar) => ar.simpleClone()) ?? [],
+			hourly: this.weatherProvider?.weatherHourlyArray?.map((ar) => ar.simpleClone()) ?? [],
+			location: this.weatherProvider?.fetchedLocationName,
+			provider: this.weatherProvider.providerName
+		};
+		this.sendNotification("WEATHER_UPDATED", notificationPayload);
 	},
 
 	scheduleUpdate: function (delay = null) {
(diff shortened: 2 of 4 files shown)


eouia:
Hmmm... The `optional chaining(?.)` made the test fail.
Optional changing has been adopted as standards since NodeJS 14 and ECMA 2020 (Chrome 80 and all the current stable browsers.)

Is this not acceptable?

modules/default/weather/weather.js
```diff
@@ -154,6 +154,15 @@ Module.register("weather", {
 		if (this.weatherProvider.currentWeather()) {
 			this.sendNotification("CURRENTWEATHER_TYPE", { type: this.weatherProvider.currentWeather().weatherType.replace("-", "_") });
 		}
+
+		let notificationPayload = {
+			current: this.weatherProvider?.currentWeatherObject?.simpleClone() ?? null,
+			forecast: this.weatherProvider?.weatherForecastArray?.map((ar) => ar.simpleClone()) ?? [],
+			hourly: this.weatherProvider?.weatherHourlyArray?.map((ar) => ar.simpleClone()) ?? [],
+			location: this.weatherProvider?.fetchedLocationName,
+			provider: this.weatherProvider.providerName
```
rejas, line 163: maybe also call it "providerName" to avoid misunderstanding

eouia: good point. I'll modify it.

rejas:
> Hmmm... The `optional chaining(?.)` made the test fail. Optional changing has been adopted as standards since NodeJS 14 and ECMA 2020 (Chrome 80 and all the current stable browsers.)
> 
> Is this not acceptable?

It snot node but eslint that fails due to the optional chaining. As can be read here: 
https://eslint.org/blog/2020/07/eslint-v7.5.0-released
you need to use

        "ecmaVersion": 2020

eouia:
> > Hmmm... The `optional chaining(?.)` made the test fail. Optional changing has been adopted as standards since NodeJS 14 and ECMA 2020 (Chrome 80 and all the current stable browsers.)
> > Is this not acceptable?
> 
> It snot node but eslint that fails due to the optional chaining. As can be read here: https://eslint.org/blog/2020/07/eslint-v7.5.0-released you need to use
> 
> ```
>     "ecmaVersion": 2020
> ```

Than ks for hints.

modules/default/weather/weather.js
```diff
@@ -154,6 +154,15 @@ Module.register("weather", {
 		if (this.weatherProvider.currentWeather()) {
 			this.sendNotification("CURRENTWEATHER_TYPE", { type: this.weatherProvider.currentWeather().weatherType.replace("-", "_") });
 		}
+
+		let notificationPayload = {
```
rejas, line 158: can be a const

modules/default/weather/weatherobject.js
```diff
@@ -146,6 +146,20 @@ class WeatherObject {
 		this.sunrise = moment(times.sunrise, "X");
 		this.sunset = moment(times.sunset, "X");
 	}
+
+	/**
+	 * Clone to simple object to prevent mutating and deprecated legacy library.
+	 *
+	 * @returns {object} simple object cloned.
+	 */
+	simpleClone() {
```
rejas, line 155: since there are so many js features used in this method (which might confuse beginners) would be nice to have an example of what this does in the jsdoc.

eouia: I'm not sure a good example could be possible because this feature is only for making a safe-immutable payload of notification.
Anyway, I added some description more.

sdetweil: @eouia @rejas   all the default modules send notifications of their data so that others might compose new apps.. 
BUT since the new weather module combined current weather and forecast, it (the new module) does NOT broadcast the forecast info when running in forecast mode.

eouia: By the way, all the payload of original default modules is mutable and not safe from modifying by recipient modules I think that could be a risk in some cases.

rejas: > By the way, all the payload of original default modules is mutable and not safe from modifying by recipient modules I think that could be a risk in some cases.

so thats why you added this method?

rejas: Also: I am wondering if there isnt a lodash method that would do the same thing?

rejas: > @eouia @rejas all the default modules send notifications of their data so that others might compose new apps.. BUT since the new weather module combined current weather and forecast, it (the new module) does NOT broadcast the forecast info when running in forecast mode.

yes, it only published the CURRENTWEATHER_TYPE, so that the compliments module can react to it ( I think I added that back in the days) Maybe its a good idea to remove that call and broadcast only the whole weather object (instead of just the CURRENTWEATHER_TYPE)?

eouia: moment object is not immutable and unsafe.

For example; In case of ‘weatherObject.sunrise’ is handed to other module as payload directly, the 3rd party module can manipulate the object;

```js
/* In other module’s receivedNotification function */
let afterSunRise = payload.sunrise.add(30, ‘minute’)
```
But by this execution, the original `this.weatherProvider.currentWeatherObject.sunrise` Will be changed. Usually this is a bad side-effect unintentional.

rejas: I dont mind 3rd party libraries like lodash, sicne they are well tested and we already use it in here. But your code your choice :-)

rejas (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2022-07-01._
 
 ### Added
 
+- Added the notification emitting from the weather module on infromation updated.
```
KristjanESPERANTO, line 14: Typo: information

PR #2844, opened 2022-05-02
Update github actions

rejas:
Use latest versions of the github actions
Use node 18 when running tests


.github/workflows/automated-tests.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         node-version: [14.x, 16.x, 17.x]
```
khassel, line 18: still v17?

rejas: fixed, thx!

khassel:
btw I moved my `develop` images from node v17 to v18 an hour before, works.

khassel (approved):

PR #2857, opened 2022-05-28
fix newsfeedfetcher using correct stream version

khassel:
correction for https://github.com/MichMich/MagicMirror/pull/2855

The stream format has changed in the node-internal fetch method which breaks the newsfeed module (when using node 18).


rejas (approved):

khassel:
@MichMich can we get this merged? Because the `develop` branch is broken in the current version (newsfeed module with node v18) ...

PR #2884, opened 2022-07-21
respect rss ttl provided by newsfeed

### LINKED ISSUE #2883: The default NewsFeed module does not respect the rss TTL value

Browser: Firefox
Hardware: Linux, NixOS
Node version: v16.15.0
MagicMirror: v2.20.0

Description:
The default NewsFeed module does not respect the rss TTL (time-to-live) value. Providers can set this to request that feed readers check every N minutes, which can ease strain on their services. (See [xeiaso.net](https://xeiaso.net/blog.rss) for an example of a provider that sets this ttl value).
It would be helpful if MagicMirror supported this.

Also, the default reloadInterval is too high - a more reasonable default would be every 30ish minutes.

khassel:
Fixes #2883

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ _This release is scheduled to be released on 2022-10-01._
 ## Fixed
 
 - Broadcast all calendar events while still honoring global and per-calendar maximumEntries.
+- Respect rss ttl provided by newsfeed (#2883).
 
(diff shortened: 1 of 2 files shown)


rejas (approved):

MichMich:
Thanks! Sorry for the slow merge. Was enjoying a sunny holiday. :)

PR #2889, opened 2022-07-25
English doesn't have a "day after tomorrow" word

MikeBishop:
English doesn't have a word for the day after the day after today.  "In 2 days" doesn't align with the behavior around `relative` formatting for non-full-day events -- that uses "today," "tomorrow," and then goes to days of the week.

Having played around with different formats, I've concluded that "In 2 days" is an odd-ball that shows up here only(?) for English.  Many languages have a word for that; obviously the code should continue to accommodate those languages.

The simplest "fix" then is simply to drop this term from `en.json`, as the code already handles the `DAYAFTERTOMORROW` field not being defined for a given language.  So let's just let English be such a language.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ _This release is scheduled to be released on 2022-10-01._
 
 ## Updated
 
+- Removed DAYAFTERTOMORROW from English
+
 ## Fixed
 
 - Broadcast all calendar events while still honoring global and per-calendar maximumEntries.
--- a/tests/e2e/translations_spec.js
+++ b/tests/e2e/translations_spec.js
@@ -164,7 +164,7 @@ describe("Translations", function () {
 			dom.window.onload = function () {
 				const { Translator } = dom.window;
 
-				Translator.load(mmm, translations.en, false, function () {
+				Translator.load(mmm, translations.de, false, function () {
 					base = Object.keys(Translator.translations[mmm.name]).sort();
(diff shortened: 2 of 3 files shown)


MikeBishop:
Hm.  So I see that the tests enforce that all translated items must be in both the base language and all other languages.  While that's admirable for coverage, in this instance the code explicitly checks whether the term exists and moves on properly if it's absent.

Is there a way to explicitly allow this translation not to exist, perhaps by having it present but with a null value?

khassel:
looking into the translation test: `en` is used as the "complete" language (with all keys assigned).

So if you delete one key, we have to use another language. Tested with `de`:

change in [this line](https://github.com/MichMich/MagicMirror/blob/master/tests/e2e/translations_spec.js#L167) `translations.en` to `translations.de` and

change in [this line](https://github.com/MichMich/MagicMirror/blob/master/tests/e2e/translations_spec.js#L179) `"en"` to `"de"` 

and the test will work again. May add some comments why this was changed ...

MikeBishop:
That appears to have worked -- thank you!

rejas (approved):

MichMich:
Thanks! Sorry for the slow merge. Was enjoying a sunny holiday. :)

MikeBishop:
No worries at all!  (For anyone's general edification, English does technically have such a word -- "overmorrow" -- but it hasn't been in common use since Middle English.)

PR #2890, opened 2022-07-25
Show time today

MikeBishop:
When I was trying to move to the "relative" timeFormat, I found that while relative was useful for events that were approaching -- the baseball game is on Saturday, etc. -- present-day events in relative format were at best annoying.

This adds an additional option that displays non-full-day events with an absolute time on the current day, while leaving the relative behavior alone for other days.  (Full-day events continue to display "Today".)

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -37,6 +37,7 @@ Module.register("calendar", {
 		hidePrivate: false,
 		hideOngoing: false,
 		hideTime: false,
+		showTimeToday: false,
 		colored: false,
 		coloredSymbolOnly: false,
 		customEvents: [], // Array of {keyword: "", symbol: "", color: ""} where Keyword is a regexp and symbol/color are to be applied for matched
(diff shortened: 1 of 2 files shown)


modules/default/calendar/calendar.js
```diff
@@ -368,7 +369,7 @@ Module.register("calendar", {
 						} else {
 							timeWrapper.innerHTML = this.capFirst(
 								moment(event.startDate, "x").calendar(null, {
-									sameDay: "[" + this.translate("TODAY") + "]",
+									sameDay: this.config.showTimeToday ? "LT" : "[" + this.translate("TODAY") + "]",
```
rejas, line 372: So if this is set, it displays "LT"? How does that string show up on the MM?

MikeBishop: It passes LT as the format string to Moment (compare the default value of `dateEndFormat` or other explicit uses [here](https://github.com/MichMich/MagicMirror/blob/c3339b47bb6ccfc8bdf927f9582d40217332a5ad/modules/default/calendar/calendar.js#L308) and [here](https://github.com/MichMich/MagicMirror/blob/c3339b47bb6ccfc8bdf927f9582d40217332a5ad/modules/default/calendar/calendar.js#L312).  LT renders time according to the current locale, e.g. "2:53 PM", "14:53", etc.

rejas (approved):

PR #2891, opened 2022-07-25
new script `install-mm` for simplifying mm installation

khassel:
Inspired by a comment of @sdetweil in the forums. Many users run `npm audit fix` because this is displayed when installing mm.
The current installation command in the docs is `npm install --only=prod --omit=dev`, not very handy.

So I thought of simplifying installation and

- added new scripts for the mm installation: `npm run install-mm` (and `npm run install-mm:dev` for install with devDependencies)
- added the params `--no-audit --no-fund --no-update-notifier` for less noise (also to vender and fonts install)

With this the audit and npm update messages are gone. The new command is faster (without audit) and this will (hopefully) produce less support because running the audit fix command (by the user) often destroyed a running installation.

Other installation processes as @sdetweil's install scripts or the docker image could also use the new scripts.

I'm open for renaming the new scripts or better ideas.

If this is accepted/merged I will add a PR to the documentation repo for updating the installation command.


khassel:
Codecov upload error seems unrelated to the changes in this PR.

sdetweil:
cool, will add those parms to my scripts.

package.json
```diff
@@ -7,9 +7,11 @@
 		"start": "DISPLAY=\"${DISPLAY:=:0}\" ./node_modules/.bin/electron js/electron.js",
 		"start:dev": "DISPLAY=\"${DISPLAY:=:0}\" ./node_modules/.bin/electron js/electron.js dev",
 		"server": "node ./serveronly",
-		"install": "echo \"Installing vendor files ...\n\" && cd vendor && npm install --loglevel=error",
```
rejas, line 10: so what happens now when someone executes "npm install" in the directory?

khassel: same as before. Normal `npm install` is executed (with audit ...) and then fonts and vendor installs are done.

rejas (approved):

Issue #2685, opened 2021-10-14
Calendar: Multiday events always starts at current day


buxxi:
In my calendar I have an event that spans from this tuesday to sunday, now on thursday it displays it as (1/4) instead of (3/6).
It also broadcasts to another module with the startDate being midnight for the current day. So I would guess some display logic is modifying the original data?

My config:
```
		{
			module: "calendar",
			position: "top_left",
			config: {
				displaySymbol: false,
				maximumEntries: 100,
				hideOngoing: false,
				broadcastPastEvents: true,
				sliceMultiDayEvents: true,
				calendars: [
					{
						color: "#55acee",
						url: "https://calendar.google.com/...REDACTED"
					}
				]
			}
		},
```

It's in a private calendar but here's the relevant event:
```
BEGIN:VEVENT
DTSTART;VALUE=DATE:20211012
DTEND;VALUE=DATE:20211018
DTSTAMP:20211014T154936Z
UID:REDACTED@google.com
CREATED:20211001T101108Z
DESCRIPTION:
LAST-MODIFIED:20211001T101109Z
LOCATION:
SEQUENCE:0
STATUS:CONFIRMED
SUMMARY:International 10
TRANSP:TRANSPARENT
BEGIN:VALARM
ACTION:DISPLAY
DESCRIPTION:This is an event reminder
TRIGGER:-P0DT3H0M0S
END:VALARM
END:VEVENT
```

jupadin:
This is the intended behavior.
@sdetweil commented on this in https://github.com/MichMich/MagicMirror/issues/2629#issuecomment-901983877.
For a multi day event, the current day instead of the start of the event is used (and we then don't know when the event has started - could be yesterday or the day before yesterday - but we use the current day as the beginning of the event).
Hope the link and the short answer helps...

PR #2895, opened 2022-08-02
Make the calendar show the whole multi day event with how far it has progressed

buxxi:
Made an attempt to debug issue https://github.com/MichMich/MagicMirror/issues/2685 and couldn't really find a reason to why multi day events had their start date changed to the current date. So I tried to remove it and made comparisons with and without it, using sliceMultiDayEvents=true/false.

|| sliceMultiDayEvents=false | sliceMultiDayEvents=true |
|-|------------------------------------|-----------------------------------|
|**startDate=today**| Vacation - Today | Vacation (1/13) - Today |
|**Don't modify startDate**| Vacation - Ends in 12 days | Vacation (25/37) - Today |

It feels like it's wrong with sliceMultiDayEvents=true without this pull request, but since it changes the other one from "Today" to "Ends in 12 days" there might be something more that is needed to be changed? Both looks fine for me at least.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ _This release is scheduled to be released on 2022-10-01._
 
 - Broadcast all calendar events while still honoring global and per-calendar maximumEntries.
 - Respect rss ttl provided by newsfeed (#2883).
+- Fix multi day calendar events always presented as "(1/X)" instead of the amount of days the event has progressed.
 
 ## [2.20.0] - 2022-07-02
 
--- a/modules/default/calendar/calendarutils.js
+++ b/modules/default/calendar/calendarutils.js
@@ -469,10 +469,6 @@ const CalendarUtils = {
 						return;
 					}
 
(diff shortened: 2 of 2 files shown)


MichMich:
I think it's a good idea to find out why the code was there in the first place. @rejas Do you have any idea? The blame points to you. :)

rejas:
Well, i didnt write the code, I merely moved it from the fetcher class into this new file... @oemel09 wrote the original code back in 2019 (6a5f0225fe87f7cd0a2db1ad71167c75a1e93be6) and @kolbyjack added a fix in 2022...

oemel09:
[As stated in my PR where I added this code](https://github.com/MichMich/MagicMirror/pull/1848) back then your example would have been displayed as `Vacataion - 25 days ago`. If that is no longer the case and it is instead being displayed as `Vacation - Ends in 12 days`, like you describe, feel free to remove my changes and do not modify the events start date 👍

MichMich:
Awesome! Thanks for your feedback @oemel09

Much appreciated!

Issue #2885, opened 2022-07-23
Calendar Repeating Events one day off 


DAHines57:
Hardware: RPi 3, Stretch
Node version: v16.14.2
MagicMirror: v2.20.0

Description: 
I have repeating google calendar event set for Thursdays at 8pm to 11pm, and this displays correctly there, but in the calendar module this event is shown to be on Friday from 8 to 11pm.  I have also noticed if I set the event to be 6pm to 8pm, the event displays correctly in the module.

Timezone: America/Chicago

iCal:
```
BEGIN:VEVENT
DTSTART;TZID=America/Chicago:20220728T200000
DTEND;TZID=America/Chicago:20220728T230000
RRULE:FREQ=DAILY;COUNT=30;INTERVAL=7
DTSTAMP:20220723T162624Z
```

Calendar Config:
```{
			module: "calendar",
			header: "Calendar",
			position: "bottom_left",
			disabled: false,
			config: {
				showEnd: false,
				dateFormat: "MMM DD - hh:mm A",
				maximumEntries: 6,
				calendars: [
```

Screenshot:
![image](https://user-images.githubusercontent.com/16733851/180614012-ba7d5954-6ede-4ab9-95f4-86d8eb2ed8b4.png)

Thanks!

sdetweil:
@DAHines57 what did u do to fix this?

DAHines57:
@sdetweil I had noticed that my repository was acting strange, so I re-cloned the repo and tried again. However I did notice that when I create the recurring event from my iPhone the date is still off. But if I create the event directly in google calendar the dates show correctly.

sdetweil:
ok, thanks. I will look at this.  is the ve ent above a good example?

DAHines57:
Here's a more updated example, the first was created directly in gCal ('Test Gcal'), the second using the iphone default calendar app ('Test') .

```
BEGIN:VEVENT
DTSTART;TZID=America/Chicago:20220721T200000
DTEND;TZID=America/Chicago:20220721T230000
RRULE:FREQ=WEEKLY;BYDAY=TH
DTSTAMP:20220725T015743Z
UID:
CREATED:20220725T005851Z

BEGIN:VEVENT
DTSTART;TZID=America/Chicago:20220728T210000
DTEND;TZID=America/Chicago:20220728T220000
RRULE:FREQ=WEEKLY;BYDAY=TH
DTSTAMP:20220725T015743Z
UID:
CREATED:20220725T014034Z
[...]

PR #2905, opened 2022-09-05
Fixcal

sdetweil:
rrule and luxon miscalculate event day when byday is specified, but NOT without. 

correct adjustment in calendarutils.js to adjust only when byday is specified

fixes: #2885

### DIFF
--- a/modules/default/calendar/calendarutils.js
+++ b/modules/default/calendar/calendarutils.js
@@ -333,9 +333,9 @@ const CalendarUtils = {
 							// If the offset is negative (east of GMT), where the problem is
 							if (dateoffset < 0) {
 								if (dh < Math.abs(dateoffset / 60)) {
-									// if the rrule byweekday WAS explicitly set , correct it									
+									// if the rrule byweekday WAS explicitly set , correct it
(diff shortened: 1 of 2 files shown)


modules/default/calendar/calendarutils.js
```diff
@@ -346,8 +349,13 @@ const CalendarUtils = {
 								//if (event.start.tz === moment.tz.guess()) {
 								// if the date hour is less than the offset
 								if (24 - dh <= Math.abs(dateoffset / 60)) {
-									// apply the correction to the date/time back to right day
-									date = new Date(date.getTime() + Math.abs(24 * 60) * 60000);
+									// if the rrule byweekday WAS explicitly set , correct it
+									if (curEvent.rrule.origOptions.byweekday !== undefined) {
+										if (curEvent.rrule.origOptions.byweekday !== undefined) {
```
rejas, line 354: looks like a duplicate line?

sdetweil: good catch, sent update

rejas (approved):

PR #2908, opened 2022-09-08
css changes

BKeyport:

Not that I've found

Adjusts font tree for consistent behavior - uses rem sizing throughout, uses variables throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ _This release is scheduled to be released on 2022-10-01._
 - Removed DAYAFTERTOMORROW from English
 - Update dependencies
 - Updated jsdoc
-- Updated font tree to use variables consistantly 
+- Updated font tree to use variables consistantly
 
 ## Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ _This release is scheduled to be released on 2022-10-01._
 - Removed DAYAFTERTOMORROW from English
 - Update dependencies
 - Updated jsdoc
+- Updated font tree to use variables consistantly 
 
 ## Fixed
 
--- a/css/main.css
+++ b/css/main.css
@@ -7,8 +7,12 @@
   --font-primary: "Roboto Condensed";
   --font-secondary: "Roboto";
 
-  --font-size: 20px;
-  --font-size-small: 0.75rem;
+ 	--font-size: 20px;
+	--font-size-xsmall: 0.5rem;
+	--font-size-small: 0.75rem;
+	--font-size-medium: 1rem;
+	--font-size-large: 1.25rem;
+	--font-size-xlarge: 1.5rem;
 
   --gap-body-top: 60px;
   --gap-body-right: 60px;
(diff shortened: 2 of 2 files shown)


rejas (changes requested):
I do like adding more variables, but why change the actual sizes (for example .medium shrinks from 1.5rem to 1rem)? That would mean that my mirror suddenly displays everything smaller, wouldnt it?

BKeyport:
Yeah, you are correct. I put in the pull that the variables can be adjusted
to fit. I just copy and pasted a sensible pattern - putting medium as the
core value as it should be, rather than small, which was the core.

rejas:
I am not sure if most users would read the changelog and immeditaly know what to do (which would lead to a lot of questions on the forum and here on github). 

Yes, sometimes stuff breaks on releases but there must be a more valid reason than just a naming issue...

In the end its up to @MichMich to decide...

MichMich:
I like the idea, but just as @rejas I'm not a big fan of breaking changes if it doesn't add any impactful value. What might be an idea is to put this new size tree onder a specific class and let users opt in for using the new size tree (although that might complicate things without any real added value)...

BKeyport:
Set font sizes to original, now only change is addition of variables to make custom CSS easier.

rejas:
This is better but unfortunatly will still break existing layouts: Those that use the --font-size-small variable will now get a different rem (1rem instead of 0.75rem)

Yes, I know its a bad naming at the moment (css selector "xsmall" uses "small" variable) but thats somethign we have to live with :-(

BKeyport:
I don't think we're going to get a full compatibility with this and the logic you're using, then. 

The simple fact is using "small" in the extra small layout point is an error and should be treated as such. 

This corrects that, as well as adding the rest of the variables. On my screen, the change is VERY minor in look, while still maintaining everything, even if I rename my custom.css to default everything back to the new main.css file. 

I will leave it up to Michael to approve or deny. 

To sum up changes as it's now written: 
1) the use of --font-size-small in .xsmall is an error. This fixes that. 
2) addition of variables for the remainder of the sizes for consistency.

rejas:
I understand your reasoning, like you say: in the end @MichMich will have the final say :-) Thx for the PR in any case!

MichMich:
Being a bit busy (lazy🤫), could you post a screenshot of the difference?

BKeyport:
As you can see, there's no visual difference between new and old, IMO. 

image

If you wanna check it for yourself, here's my fiddle playing with it. I couldn't get a working module together to do this for whatever reason, so I just copy and pasted the css over. 
https://jsfiddle.net/bkey1970/vdboL7jy/

KristjanESPERANTO (approved):
The current status looks good to me :slightly_smiling_face:

rejas:
On my current setup the sizes of the header changes: Bildschirmfoto 2022-09-17 um 08 31 32

BKeyport:
> On my current setup the sizes of the header changes:

Not any longer. missed that the header used the --font-size-small - now uses --font-size-xsmall to match intent.

rejas (approved):
Since you are a quick fixer I approve :-)

MichMich:
Unfortunately the tests are failing.

rejas:
maybe just a hiccup during npm install? Can you trigger the tests again?

khassel:
from the failed tests:

```bash
Checking formatting...
[warn] CHANGELOG.md
[warn] css/main.css
[warn] Code style issues found in 2 files. Forgot to run Prettier?
```

BKeyport:
code style issues in both files are outside my field of work.

rejas:
just run "npm run lint:prettier" and commit the changes :-)

KristjanESPERANTO (approved):

PR #2914, opened 2022-09-21
Remove old weather modules

khassel:
The modules `currentweather` and `weatherforecast` were already only dummies displaying they are deprecated:

"This module is deprecated since release v2.15 and removed with v2.19."

So removal was announced with v2.19, time to get rid of them ...


rejas (approved):

rejas:
I think the README of the new weather module could be adjusted too

PR #2919, opened 2022-09-30
Ongoing multi-day events

MikeBishop:
This adds a `multiDay` flag to events, so that multi-day events can be treated slightly differently.  In the current code, ongoing multi-day events always occur "Today".  This code excludes multi-day events from that code block, causing them to fall down into the ongoing event code.  That renders them as "Ends in X days" until the last day of the multi-day event, then renders them as "Ends in X hours".

My preference would have been to reuse the code at line ~402 to get "Ends Today", "Ends Tomorrow", "Ends Thursday", etc. but the string in the translations file in "Ends in {time}" and I wasn't able to massage Moment to produce something that looks right without getting into multi-language world. Maybe an area for future work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ _This release is scheduled to be released on 2022-10-01._
 - Fix weatherbit provider to use type config value instead of endpoint
 - Fix calendar events which DO NOT specify rrule byday adjusted incorrectly #2885
 - Fix e2e tests not failing on errors (#2911)
+- Ongoing multi-day events display time until end instead of "Today"
 
 ## [2.20.0] - 2022-07-02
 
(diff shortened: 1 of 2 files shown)


rejas:
Thx for your PR. But isnt it displayed already in the format you describe? My Mirror shows the current ongoing event as ending in two days...

MikeBishop:
Hm, you're right, it should be doing that....  I just switched back to develop and confirmed that the event that made me investigate is no longer perpetually showing "Today", because `event.today` is now false.  Displaying "Today" on the first day, then "Ends in...." on subsequent days seems perfectly reasonable behavior.

What I'm failing to understand then is why `event.today` was true when I started investigating this yesterday, two days into the multi-day event. But I suppose that will have to be something to debug if/when I have another repro.

For now, this isn't the right fix.  Thanks for the catch!

MikeBishop:
Ah, I found a little more detail.  On `master` prior to the most recent release, `event.startDate` gets set to today's date, at least for this particular event. That doesn't occur in `develop` or in the newly released version.  It appears that this was another manifestation of #2685; with that fixed, no further change is necessary!

Issue #2487, opened 2021-03-17
Async node_helper start()


MikeBishop:
If a node_helper uses an async method for start(), it appears that MagicMirror declares the system ready as soon as the function returns the initial Promise.  It would be nice if MM would collect the promises from the various modules, then `await Promise.all(...)` of them.

**Steps to Reproduce**: Use the async/await pattern within a node_helper's start() method

**Expected Results**: Initialization completes when all node_helpers are done initializing

**Actual Results**: MagicMirror reports ready while node_helper is still printing loading status to console

rejas:
Is there a module you can point to that uses async start?

MikeBishop:
I've been trying to use it in my own module, MMM-Powerwall.  I borrowed logic from MMM-RemoteControl that reads config files off the disk, and I'd already used the async file access methods elsewhere.

sdetweil:
but u only need to use async functions if u intend to use await. which  is similar to using sync calls.
there are a few different approaches to using async and await without doing it with start. 
you can start a timer, or use the notification  call from module to node helper.

rejas:
Hi @MikeBishop finally came around to look at some old issues, and whipped up a PR for your suggestion. Seems to work on my local setup but I want to get a little feedback first before I proceed.

rejas:
Fixed with #2928

MikeBishop:
So there's apparently an issue with waiting for modules that take a long time.  It looks like if initialization of the modules gets pushed out past a certain point in electron's loading process, electron will display a blank screen locally.  Accessing with a browser works fine, because everything initialized by then.

I think I have a clean repro of the issue in https://github.com/MikeBishop/MMM-asyncdeath – the synchronous start function produces this in the logs and loads normally:

```
[21.01.2023 14:55.02.692] [LOG]   Synchronous start-up -- all is well
[21.01.2023 14:55.02.693] [LOG]   Sockets connected & modules started ...
[21.01.2023 14:55.03.235] [LOG]   Launching application.
```

The async start function produces this in the logs and fails:

```
[21.01.2023 14:52.54.143] [LOG]   About to break by using await
[21.01.2023 14:52.54.884] [LOG]   Launching application.
[21.01.2023 14:52.56.512] [LOG]   Sockets connected & modules started ...
```

Note the reversed order of “Launching application” and “modules started”. Seems feasible this could also happen with a synchronous module startup that takes a long time, but I haven’t tested that.

[...]

PR #2928, opened 2022-10-03
Wait till all node_helper are started before finishing startup

rejas:
In response to #2487 this implements a Promise.all for the node_helper start calls. Just a draft becuase I want to get feedback first....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Special thanks to: @rejas, @sdetweil
 ### Updated
 
 - Cleaned up test directory
+- Wait for all modules to start before declaring the system ready (#2487)
 - Updated e2e tests (moved `done()` in helper functions) and use es6 syntax in all tests
 - Updated da translation
 - Rework weather module
--- a/js/app.js
+++ b/js/app.js
@@ -226,17 +226,21 @@ function App() {
 				httpServer = new Server(config, function (app, io) {
 					Log.log("Server started ...");
 
+					const nodePromises = [];
+
 					for (let nodeHelper of nodeHelpers) {
 						nodeHelper.setExpressApp(app);
 						nodeHelper.setSocketIO(io);
-						nodeHelper.start();
+						nodePromises.push(nodeHelper.start());
 					}
 
 					Log.log("Sockets connected & modules started ...");
 
-					if (typeof callback === "function") {
-						callback(config);
-					}
+					Promise.all(nodePromises).then(() => {
+						if (typeof callback === "function") {
+							callback(config);
+						}
+					});
 				});
 			});
 		});
--- a/js/app.js
+++ b/js/app.js
@@ -234,9 +234,9 @@ function App() {
 						nodePromises.push(nodeHelper.start());
 					}
 
-					Log.log("Sockets connected & modules started ...");
+					Promise.allSettled(nodePromises).then(() => {
+						Log.log("Sockets connected & modules started ...");
 
-					Promise.all(nodePromises).then(() => {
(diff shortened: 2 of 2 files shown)


MikeBishop (commented):
Two pieces of feedback:

- It might make sense to move the log line into the completion block, so that the message indicates when all modules _have_ started rather than when all modules have been told to start. Or log in both places, to give some visibility into the time it takes for that to happen.
- Contrary to my original suggestion, `Promise.allSettled()` (Node 12.9+) might be more appropriate so that an error doesn't cause it to complete prematurely.

rejas:
Thanks for the feedback @MikeBishop Now ready for real review ;-)

sdetweil (commented):
what about those that fail?   are we expecting the module to alert the user somehow, and we say nothing?  just thinking of debug

rejas:
not sure how many module developers use async here at all ....

sdetweil:
well, none in the start function.. cause it doesn't exist..  i use both promises and async where appropriate

sdetweil (approved):

rejas:
a module developer could use async anyway, even if it is not used in the prototype method (just like the OP does). might be interesting to mention it in the docs for developers...

khassel (approved):

MikeBishop (approved):

MikeBishop:
FWIW, I wound up shipping mine using async anyway, and just putting the things that might take longer last.  I haven't encountered any ill effects from my start-up routine completing a bit after the parent process thinks it did.

But @sdetweil makes an interesting point.  In the current design, if a module throws an exception during start-up, does the server start-up fail entirely?  If that's the desired behavior, then `Promises.all()` was the correct method to start with, and having it throw the first exception encountered and kill the program would provide a similar result.

rejas:
Im currently on a trip, so no real way for me to check the current behaviour. 
I would hope though that in case of an error the other modules start too and a meaningful error is shown somewhere (console or notification)...

rejas:
> In the current design, if a module throws an exception during start-up, does the server start-up fail entirely? If that's the desired behavior, then `Promises.all()` was the correct method to start with, and having it throw the first exception encountered and kill the program would provide a similar result.

Currently if a node_helper of a module throws an exception in the start method, all other node_helper after that dont get started due to:

"
[12.10.2022 22:49.47.556] [ERROR] Error: listen EADDRINUSE: address already in use 127.0.0.1:8080
"

As an example, use the default config and throw an exception in the newsfeed helper module. Calendar module wont start and show nothing...

With the changes here that will still happen. Why? Because the start method doesnt return a promise, and we have no try/catch anywhere in the startup sequence.

Do I like this behvior? No, since it confuses a user. If a module errors, only that module should error and not any other module that has nothing to do with it. 

Two ways to fix it:
1) Add a try catch somewhere around where the server starts and handle it gracefully
2) Make the start method async (since it will then return a promise that can be handled with Promise.all...) and hope module developers change their method signature to use async too

Obviously 2) is a wishful thinking and would probably result in old modules that still work not being resilient. So I think 1) is the way to go. But since that doesnt have to happen here in this PR it can be done seperatly after this gets merged.

Comments?

sdetweil:
yes, 2 is wonderful idea, but we have so many modules that are no longer maintained, this will never happen

so it must be option 1. 

I personally hate that we have a fatal config error but continue to run with something else

khassel:
from my side this can be merged as it doesn't change the current state.

For a future PR: option 1 ...

rejas:
Created a ticket for the error handling improvement: https://github.com/MichMich/MagicMirror/issues/2944

PR #2902, opened 2022-08-28
feat(weather/smhi): Add hourly forecasts, apparent temperature & custom location name

SkySails:
This PR does not close any issues. It does, however, aim to improve the SMHI weather provider with support for hourly forecasts and apparent ("Feels like") temperature, along with the ability to customize the name of the location.

PR #2931, opened 2022-10-04
fix(weather/smhi) Correctly reference apparent temp method

SkySails:
This PR addresses [this comment](https://github.com/MichMich/MagicMirror/commit/48756e877403d914c64c8da06a4f7e5bfbace07c#commitcomment-85772193), which points out an issue with #2902.

Looks like the apparent temp calculation method was incorrectly referenced 😅


rejas (approved):
Good catch @majorfrog

Issue #2840, opened 2022-04-22
Impossible to display full article


josselynL:
Hello everyone,

I want to see details of current news displayed with notification ARTICLE_MORE_DETAIL or ARTICLE_TOGGLE_FULL notification, but this is not functionnal (just display black screen)

Logs
```
0 | mm
(node:26811) electron: Failed to load URL: https://XXX  with error:ERR_BLOCKED_BY_RESPONSE
```

I think this is because the website block iframe to avoid click-jacking (https://stackoverflow.com/questions/48668767/error-when-loading-an-external-url-in-an-iframe-with-electron), so there is no possibilities in this case ?

Thank you for your help !

khassel:
This should be possible since https://github.com/MichMich/MagicMirror/pull/2850

Can this be closed? @josselynL

josselynL:
> This should be possible since #2850
> 
> Can this be closed? @josselynL

Thank you for this information.
I will try and come back to you soon with the result

josselynL:
@khassel 
MM freshly updated, but problem is still the same with same error.
Does I need to do something ?

Thank you.

khassel:
> Does I need to do something ?

you have to add the line `httpHeaders` into your `config.js` as provided below:

```js
let config = {
	address: "localhost",
	port: 8080,
	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false, frameguard: false },
	...
}
```

josselynL:
> ```js
> 	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false, frameguard: false },
> ```

Thank you for your replay.
But still the same issue.

`0|mm  | (node:27973) electron: Failed to load URL: https://actu.fr/occitanie/toulouse_31555/***********.html with error: ERR_BLOCKED_BY_RESPONSE
`

khassel:
can you provide your `config.js` with the related module config so we can reproduce this error?

Please remove all private information as e.g. apikeys before ...

josselynL:
Thank you.
Here is. I've delete all modules configuration, I think you don't need it :)

`js
/* Magic Mirror Config Sample
 *
 * By Michael Teeuw https://michaelteeuw.nl
[...]

PR #2940, opened 2022-10-08
fix cors problems with newsfeed articles

khassel:
solves #2840 as far as possible. There could still be errors on the embedded iframe when the owner of the site has set `X-Frame-Options` or `Access-Control-Allow-Origin` headers (as already mentioned in the docs).

### DIFF
--- a/modules/default/newsfeed/newsfeed.js
+++ b/modules/default/newsfeed/newsfeed.js
@@ -42,7 +42,13 @@ Module.register("newsfeed", {
 		dangerouslyDisableAutoEscaping: false
 	},
 
-	urlPrefix: location.protocol + "//" + location.host + "/cors?url=",
+	getUrlPrefix: function (item) {
+		if (item.useCorsProxy) {
+			return location.protocol + "//" + location.host + "/cors?url=";
+		} else {
+			return "";
+		}
+	},
 
 	// Define required scripts.
 	getScripts: function () {
@@ -144,7 +150,7 @@ Module.register("newsfeed", {
 			sourceTitle: item.sourceTitle,
 			publishDate: moment(new Date(item.pubdate)).fromNow(),
 			title: item.title,
-			url: this.urlPrefix + item.url,
+			url: this.getUrlPrefix(item) + item.url,
 			description: item.description,
 			items: items
 		};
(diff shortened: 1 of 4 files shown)


rejas:
Seems to look good, but I'd like to wait for my approval until Tuesday when I return from my trip

modules/default/newsfeed/newsfeed.js
```diff
@@ -42,6 +42,8 @@ Module.register("newsfeed", {
 		dangerouslyDisableAutoEscaping: false
 	},
 
+	urlPrefix: location.protocol + "//" + location.host + "/cors?url=",
```
rejas, line 45: so this will always be used even if cors is not necessary?

khassel: yes, I thought about a config option but as it works so too I decided against.

rejas: Does that have any downsides? I dont have much knowledge into cors, so I am afraif sites might not be loading if this cors hack is used all of the time?

khassel: I found no downsides.

I tested with the urls provided in #2840 and with mine, no one did work without the cors prefix. These are not the newsfeed url's but the url's of the articles and they are only used if you want read the full article (which I think is not very often).

rejas: I would feel better if we had a config option to disable cors so that we could point users that run into an issue to it when the next release is out.

khassel: so on what level do you see this config option? One option for whole newsfeed module or option for every feed?

And which should be the default? From my side I would enable the cors proxy by default.

rejas: Yes, I would enable the cors proxy by default too. 
And have the config option per newsfeed, not for the whole module.

modules/default/newsfeed/newsfeedfetcher.js
```diff
@@ -20,7 +20,7 @@ const stream = require("stream");
  * @param {boolean} logFeedWarnings If true log warnings when there is an error parsing a news article.
  * @class
```
rejas, line 21: Missing JSDoc @param "useCorsProxy" declaration

rejas (changes requested):
just one small jsdoc addition needed :-)

modules/default/newsfeed/newsfeedfetcher.js
khassel: fixed ... but should these warning not treated as errors? Or will you check every test manually in the future?

rejas: na, wont check manually since the warning shows up here in the changes ... erroring it might trigger the old ones, I fear...

rejas (approved):

Issue #2944, opened 2022-10-13
Node_helper errors during startup prevent other module helpers from starting


rejas:
As discussed in https://github.com/MichMich/MagicMirror/pull/2928#issuecomment-1276729550 the current behaviour in the startup of the node_helpers is not optimal:

once a nodehelper throws an exception, the whole chains stops due to some internal error handling of the server. 

that leads to modules not showing up fully although they are fine. 

therefore some error handling has to happen so that one bad module doesnt spoil the whole bunch :-)

PR #2849, opened 2022-05-10
Use metric units internally in all weatherproviders

rejas:
So finally I think this refactorin is ready to be reviewed :-)

DONE:
- [x] Removed all conversion functions for wind and temperature from specific weatherproviders
- [x] Use internally only metric units: celsius for temperature, meters per seconds for wind
- [x] Convert temp and wind into the configured units when displaying data on the UI
- [x] look how beaufort calculation uses metrics, added knots as new windunit
- [x] add more e2e tests 

Checked providers:
- [x] Darksky
- [x] EnvCanada
- [x] OpenWeatherMap
- [x] SMHI provider 
- [x] UK Met Office
- [x] UK Met Office DataHub
- [x] WeatherBit
- [x] WeatherFlow
- [x] WeatherGov

TODO in different tickets:
- check weatherproviders for usage of weatherEndpoint (as seen in https://github.com/MichMich/MagicMirror-Documentation/issues/131) -> see #2926 
- cleanup precipations -> #2953

PR #2945, opened 2022-10-13
Add error handling to node_helper startup sequence

rejas:
Fixes https://github.com/MichMich/MagicMirror/issues/2944


js/app.js
```diff
@@ -231,7 +231,13 @@ function App() {
 				for (let nodeHelper of nodeHelpers) {
 					nodeHelper.setExpressApp(app);
 					nodeHelper.setSocketIO(io);
-					nodePromises.push(nodeHelper.start());
+
+					try {
+						nodePromises.push(nodeHelper.start());
+					} catch (error) {
```
buxxi, line 237: This wouldn't catch errors happening if `nodeHelper.start()` returns a promise which fails. I think you need to make sure the result from `start()` is converted to a promise and then use the .catch() on it.

rejas: The promise would be catched (and can be accessed in the allSettled results value. Added logging for that.

buxxi: Yeah, I meant it wouldn't be caught in that try-catch and logged. Looks good now.

js/server.js
```diff
@@ -19,117 +18,119 @@ const Utils = require("./utils.js");
...
-		},
-		allowEIO3: true
-	});
-
-	server.on("connection", (socket) => {
-		serverSockets.add(socket);
-		socket.on("close", () => {
-			serverSockets.delete(socket);
+
+	this.open = function () {
```
buxxi, line 29: I think this should have the callback. I'll need it in https://github.com/MichMich/MagicMirror/pull/2952 and should probably wait for `on("listening")` before calling back (I'll fix that in that PR if I can get the tests to work).

khassel (approved):

js/server.js
rejas: so, you approve of this PR anyway?

buxxi (approved):

js/server.js
buxxi: yeah :)

I'll add it back if I need it.

khassel:
"never merge your own PR" ...

rejas:
wups :-(

rejas:
will be more patient with my weather refactor branch, I promise!

khassel:
tell me if you want to wait for more approvals there or if I should merge ...

rejas:
Since that (#2849) is quite large, the more eyes the better (looking at @sdetweil @MichMich and maybe @buxxi too for reviews of it)

PR #2946, opened 2022-10-14
Support 3rd party modules using both ESM and CommonJS

buxxi:
The problem we had before with node-fetch migrating to an ESM got me thinking about what it would require to make 3rd party modules ESM. So I made this to test having the node_helper being either CommonJS or ESM.

If naming the file `node_helper.mjs` instead of `node_helper.js` it should load it as ESM instead.

This is just a draft, there are a few things left:
- the module alias for node_helper isn't working, the module needs to specify the file with `import NodeHelper from "../../js/node_helper.js";` which is annoying
- I haven't done anything about the client side loading, but should be possible with `<script type="module".../>`, probably a bit harder to have the dual support of .mjs and .js without any configuration...

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -102,9 +102,9 @@ function App() {
 	 * Loads a specific module.
 	 *
 	 * @param {string} module The name of the module (including subpath).
-	 * @param {Function} callback Function to be called after loading
+	 * @returns {Promise} A promise that resolves as soon as the module is loaded.
 	 */
-	function loadModule(module, callback) {
+	async function loadModule(module) {
 		const elements = module.split("/");
 		const moduleName = elements[elements.length - 1];
 		let moduleFolder = `${__dirname}/../modules/${module}`;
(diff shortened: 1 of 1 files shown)


rejas:
Thanks for the brainstorming first of all. 

But reading "the module alias for node_helper isn't working, the module needs to specify the file with import NodeHelper from "../../js/node_helper.js"; which is annoying" is a little understatement :-) Its quite impossible to get modules to use a new syntax without leaving a lot of old ones behind (and users possibly angry)

Using a more modern syntax is probably the way to go but maybe we need to make small steps (like your async work which looks like some stuff I try out too in my branches).

buxxi:
Could move most of the async stuff out to a separate PR to make the original intent smaller and possibly ignore my second point for now and only make this about the node_helpers. `require()` is sync which can only load CommonJS, `import()` is async which can load both, that's why I started with it and just felt the flow of continuing :)

This wasn't intended to break any of the existing modules (CommonJS modules should still load fine) and doing ESM modules will always be a bit different (no `require()`, `import()`/`export()`, no `_dirname` etc). I can't see node dropping CommonJS support without going the way of Python 2 & 3 and splitting up the community. The point is mainly about new modules and not the existing ones (sure I will migrate my own 3rd party modules). A JS developer coming from the frontend is probably already using ESM modules, Node documentation gives examples with `import()` and not `require()`.

And I agree that the relative path is not an ok solution, just what I had to do for my test to actually work (as this is only a draft).

Will have to figure out why the tests segmentation faults too :)

buxxi:
Since I failed to solve the prerequisite PR https://github.com/MichMich/MagicMirror/pull/2962 I will close this for now.

Issue #2942, opened 2022-10-11
Improve mocking dates/times in e2e tests


rejas:
Some module e2e tests need to fake a certain date/time to be run, especially the calendar module.

There isnt (yet) a general way to mock the current date/time in the e2e tests. Only hackish solutions exist, like for the compliments module for displaying a compliment on a specific date:

https://github.com/MichMich/MagicMirror/blob/a328ce537faf8d9b90d7eb2c09237fe83ae3a827/tests/e2e/modules/compliments_spec.js#L83

Here a mockDate is in the test config which gets read in the mm-js-logic. Using test logic in normal code is of course a no-no!

There are quite a few solutions out there, like sinon, mockDate or adding a jest.mock. Alas, the author hasnt gotten any of them to run for faking a test :-(

Adding the functionality to mock a certain date/time would open up a lot of new test posibilities, for example [this new feature](https://github.com/MichMich/MagicMirror/pull/2939) could be tested very easily.

rejas:
Looks like what we'd need, did you try to replace the mockDate construct in the mentioned compliments-test?

> Need we more?

One gotta look through the code where it makes sense to test and fake dates/time to be sure :-)

khassel:
sounds like I should assign this to myself ;)

khassel:
are these your points in compliments tests?

- `mockDate` is used in `compliments.js`, very ugly, so this should be removed
- some tests run only within specific hours, should run every time and hour set as mock before

khassel:
the above mentioned `jest.setSystemTime` works for normal tests (`unit`) but not in our `e2e` test setup because the date/time of the browser has to be mocked. Found no solution for this.

Other possibility is to move such tests back to `electron` section because `playwright` can fake browser date/time.

Here a first (dirty) working example:

```js
const { _electron: electron } = require("playwright");

let electronApp = null;
process.env.MM_CONFIG_FILE = "tests/configs/modules/compliments/compliments_parts_day.js";
process.env.TZ = "GMT";
jest.retryTimes(3);

describe("Compliments", () => {
	beforeEach(async () => {
		electronApp = await electron.launch({ args: ["js/electron.js"] });
	});

	afterEach(async () => {
		await electronApp.close();
	});

	it("Morning compliments for that part of day", async () => {
		expect(await electronApp.windows().length).toBe(1);
		const page = await electronApp.firstWindow();
		expect(await page.title()).toBe("MagicMirror²");
[...]

PR #2947, opened 2022-10-15
added new electron tests supporting date mocking

khassel:
first PR for #2942 

- added new electron tests for calendar which test new css classes from https://github.com/MichMich/MagicMirror/pull/2939
- moved some compliments tests from `e2e` to `electron` because of date mocking
- removed mock stuff from compliments module


tests/electron/modules/compliments_spec.js
```diff
@@ -0,0 +1,68 @@
...
+	describe("Feature date in compliments module", () => {
+		describe("Set date and empty compliments for anytime, morning, evening and afternoon", () => {
+			it("Show happy new year compliment on new years day", async () => {
+				await helpers.startApplication("tests/configs/modules/compliments/compliments_date.js", "01 Jan 2022 10:00:00 GMT");
+				await doTest(["Happy new year!"]);
+			});
+		});
+	});
+
+	describe("remoteFile option", () => {
```
rejas, line 62: Do all tests have to be moved to the electron page? Seems unnecessary....

rejas (changes requested):

tests/electron/modules/compliments_spec.js
khassel: thats more a general question if you want split the small test into 2 or having all together.

We have here a range from
- move only tests to `electron` which needs date/time mocking

to
- move all `e2e` tests back to `electron` section

rejas: From my understanding (and please correct me if I am wrong) it was that electron had only the tests that opened a real electron window, while e2e didnt had to open a electron window. I personally like that distinction and would like to keep the electron tests at a minimum, since running the tests in e2e doesnt disrupt as often and seems faster...

rejas: Maybe @sdetweil can give an opinion on how he likes the tests sorted :-)

khassel: I'm asking because I don't want to change this several times, can live with the decision to minimize electron tests.

rejas (approved):

PR #2935, opened 2022-10-06
Use fetch instead of XMLHttpRequest in weatherprovider

rejas:
small update to the fetchData method to use the fetch helper instead of the old XCMLHttpRequest. 
Also fixes some typos :-)

PR #2948, opened 2022-10-16
New weather provider: Yr.no

MagMar94:
# Added Yr.no as a weather provider

Yr.no is a free Norwegian weather service. The configuration is quite simple:
```js
{
    weatherProvider: "yr",
    lat: 59.9171,
    lon: 10.7276,
    altitude: 30
}
```
The latitude and longitude cannot have more than 4 decimals, but that should be plenty. To quote yr: "There is no need to ask for weather forecasts with nanometer precision!". The altitude should be meters above sea level and defaults to 0. If `type` is set to `current` the symbol can display the next 1, 6 or 12 hours by setting `currentForecastHours` (default is 1).

It states in [Getting started-guide](https://developer.yr.no/doc/GettingStarted/) that users of the API should cache the results and use the `Expires`-header to know when to ask for new data. By using the `If-Modified-Since`-header we can avoid downloading the same data over and over again. I chose not to override the `User-Agent`-header set in [`server.js`](https://github.com/MichMich/MagicMirror/blob/a328ce5/js/server.js#L97) even though it does not comply with [the terms of service](https://developer.yr.no/doc/TermsOfService/). It currently works with the default header, and by searching the web for MagicMirror the GitHub-repo should be easy to find without an explicit link.

I also had to make some minor changes to `server.js` and `weatherprovider.js` to be able to send and return HTTP headers. To handle the HTTP 304 response without body I chose to return `undefined` so we easily can use the response as a condition: `if (response) ...`.

The documentation for the API is available here:
- [API Reference overview](https://api.met.no/weatherapi/)
   - [Locationforecast](https://api.met.no/weatherapi/locationforecast/2.0/) - Used to get the weather forecast
   - [Sunrise](https://api.met.no/weatherapi/sunrise/2.0/documentation) - used to find sunrise and sunset times


modules/default/weather/weatherprovider.js
```diff
@@ -126,19 +158,31 @@ const WeatherProvider = Class.extend({
 	 * @param {string} type what contenttype to expect in the response, can be "json" or "xml"
 	 * @returns {Promise} resolved when the fetch is done
 	 */
-	fetchData: async function (url, type = "json") {
-		url = this.getCorsUrl() + url;
+	fetchData: async function (url, method = "GET", type = "json", requestHeaders = undefined, expectedResponseHeaders = undefined) {
```
khassel, line 161: fetchData is used in all providers so changing the params order here will break at least the `envcanada.js` provider and AFAIS `method` is not used inside the `fetchData` function.

MagMar94: Ah, you're right, nicely spotted. When merging my changes into the development-branch I didn't catch that the method-parameter was removed in https://github.com/MichMich/MagicMirror/pull/2935. I'll fix it 😊

CHANGELOG.md
```diff
@@ -26,9 +27,11 @@ Special thanks to: @rejas, @sdetweil
 - Wait for all modules to start before declaring the system ready (#2487)
 - Updated e2e tests (moved `done()` in helper functions) and use es6 syntax in all tests
 - Updated da translation
+- The `cors`-method in `server.js` now supports sending and recieving HTTP headers.
```
rejas, line 30: could we split up the yr provider from the rest of the changes? so we can take a better look at each before merging?
havin two or more features in a PR makes things sometimes easy to miss

rejas (changes requested):

CHANGELOG.md
MagMar94: Sure 👍  Looking at the code coverage it looks like I have to learn Jest as well. I'll put them up as separate PRs, but probably not until next week 😊

MagMar94: Added the support for HTTP-headers in https://github.com/MichMich/MagicMirror/pull/2957. Will change this PR to reflect the changes.

MagMar94:
When testing the yr weather provider I got a sense that the synchronization inspired by [this article](https://medium.com/swlh/semaphores-in-javascript-e415b0d684bc) wasn't working, so I added a few log-statements like this:
```js
getWeatherDataSynchrounous() {
		console.log("Getting weather data. Waiting: " + this.waitingForWeatherData + ". Queue: " + this.weatherDataQueue.length);
	...
	if (!this.waitingForWeatherData) {
		this.waitingForWeatherData = true;
		console.log("Locked weather data.");
		...
		if (this.weatherDataIsValid(weatherData)) {
			console.log("Unocked weather data.");
			this.waitingForWeatherData = false;
			this.getWeatherDataSynchrounous();
			Log.debug("Weather data found in cache.");
			...
		} else {
			this.getWeatherDataFromYr(weatherData?.downloadedAt)
				.then((weatherData) => {
					Log.debug("Got weather data from yr.");
					...
				})
				.catch((err) => {
					...
				})
				.finally(() => {
					console.log("Unocked weather data.");
					  this.waitingForWeatherData = false;
					  this.getWeatherDataSynchrounous();
				});
		}
	}
},
```
The output proved me right (disregard the error, I changed the `server.js` to temporarily avoid calling fetch):<br/>
logs

I'm not that experienced with JavaScript, but I figure this is because the two threads have different copies of the variables. I tried using `SharedArrayBuffer`, but it seems not to be defined. Does anyone have any ideas on how to avoid duplicate calls to the API?

A possible work-around is to trust all users to set `initialLoadDelay` if they use several instances of the provider, but I don't really want to do that.

modules/default/weather/providers/yr.js
```diff
@@ -0,0 +1,631 @@
...
+				if ((forecastDiffToEight === undefined || closestTime < forecastDiffToEight) && timeseries.data.next_12_hours) {
+					forecastDiffToEight = closestTime;
+					forecast = timeseries;
+				}
+			}
+			const forecastXHours = forecast.data.next_12_hours ?? forecast.data.next_6_hours ?? forecast.data.next_1_hours;
+			if (forecastXHours) {
+				forecast.symbol = forecastXHours.summary?.symbol_code;
+				forecast.precipitation = forecastXHours.details?.precipitation_amount;
+				forecast.minTemperature = precipitation_amount_min;
```
rejas, line 608: The temperature is read from a precipation field?

modules/default/weather/providers/yr.js
```diff
@@ -0,0 +1,631 @@
...
+			const comparison = currentTime.diff(moment(forecastTime.time));
+			if (0 < comparison && comparison < closestTimeInPast) {
+				closestTimeInPast = comparison;
+				forecast = forecastTime;
+			}
+		}
+		const forecastXHours = this.getForecastForXHoursFrom(forecast.data);
+		forecast.weatherType = this.convertWeatherType(forecastXHours.summary.symbol_code, forecast.time);
+		forecast.precipitation = forecastXHours.details?.precipitation_amount;
+		forecast.minTemperature = forecastXHours.details?.precipitation_amount_min;
```
rejas, line 69: The temperature is read from a precipation-amount field?

modules/default/weather/weatherprovider.js
```diff
@@ -111,36 +111,132 @@ const WeatherProvider = Class.extend({
...
+				if (!dataResponse.headers) {
+					dataResponse.headers = this.getHeadersFromResponse(expectedResponseHeaders, response);
+				}
+				return dataResponse;
+			}
+		}
+	},
+
+	/**
+	 * Gets a URL that will be used when calling the CORS-method on the server.
```
rejas, line 148: is this whole stuff necessary after you rmerge of the Support HTTP headers with CORS-method (#2957) ?

rejas (changes requested):

modules/default/weather/weatherprovider.js
rejas: Is this something that can be generalized so that other modules could use it too on the client side?

MagMar94: Yes, I moved it to a `utils`-file now. This should make it easy to utilize for other modules.

modules/default/utils.js
```diff
@@ -0,0 +1,146 @@
...
+		for (const header of expectedResponseHeaders) {
+			const headerValue = response.headers.get(header);
+			responseHeaders.push({ name: header, value: headerValue });
+		}
+	}
+
+	return responseHeaders;
+};
+
+module.exports = {
```
rejas, line 144: This isnt recognized by the brwoser and leads to 
"Uncaught ReferenceError: module is not defined
    at utils.js:144:1
    "

modules/default/weather/providers/yr.js
```diff
@@ -0,0 +1,622 @@
...
+		if (!this.config.lon) {
+			Log.error("Longitude not provided.");
+			throw new Error("Longitude not provided.");
+		}
+
+		const lat = this.config.lat.toString();
+		const lon = this.config.lon.toString();
+		const altitude = this.config.altitude ?? 0;
+
+		if (lat.includes(".") && lat.split(".")[1].length > 4) {
```
rejas, line 184: I think it would be more user friendly if the logic would gracefully cut the decimals to max 4 and just issue a warning instead of not starting at all

rejas (changes requested):

modules/default/utils.js
MagMar94: strange, "works on my machine"... I'll fix it 👍

MagMar94: How about now? I used the `if (typeof module !== "undefined")`-check used in several other parts of the code.

tests/unit/modules/default/utils_spec.js
```diff
@@ -0,0 +1,111 @@
...
+
+					const response = await performWebRequest(url, "json", false, undefined, ["header1"]);
+
+					expect(response.headers.length).toBe(1);
+					expect(response.headers[0].name).toBe("header1");
+					expect(response.headers[0].value).toBe("value1");
+				});
+			});
+		} else {
+			test("Always ok, need one test", () => {});
```
rejas, line 108: so why exactly dont we need a test for node < 18 ?

rejas (approved):
looks good I'd say. any more would be nitpicking (and/or can be done later once it is in the wild with people testing it :-)

just one question I have abou the tests

tests/unit/modules/default/utils_spec.js
MagMar94: Because they don't run 👀  It boils down to the fetch-method (in much the same way as I struggled with fetch in https://github.com/MichMich/MagicMirror/pull/2957#issuecomment-1290958862). When I tested with node `16.17.1` I got the following error:

Skjermbilde 2022-11-07 kl  22 14 32

It seems the fetch-related types are not defined. According to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/fetch#browser_compatibility) fetch was introduced in Node 18, so it makes sense. Fetch was introduced to the weather module in https://github.com/MichMich/MagicMirror/pull/2935. I figured since fetch was already implemented I wouldn't bother with reimplementing another kind of fetch.

rejas:
Thx for your work!!

PR #2952, opened 2022-10-18
Make the e2e tests wait for the app to start and close before running next test

buxxi:
When trying to debug why the tests broke for https://github.com/MichMich/MagicMirror/pull/2946 I found that the tests does not wait for the app to start and close. So if the startup isn't blocking that would fail.

So I added a callback for `close()` too and converted them to promises for the `startApplication()` and `stopApplication()` and updated all the e2e tests to await both. Will try to refactor all these callbacks to promises in a later PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Special thanks to: @rejas, @sdetweil
 - Ensure updatenotification module isn't shown when local is _ahead_ of remote
 - Possibility to change FontAwesome class in calendar, so icons like `fab fa-facebook-square` works.
 - Fix cors problems with newsfeed articles (as far as possible), allow disabling cors per feed with option `useCorsProxy: false` (#2840)
+- Tests not waiting for the application to start and stop before starting the next test
 
 ## [2.21.0] - 2022-10-01
 
--- a/js/app.js
+++ b/js/app.js
@@ -252,13 +252,13 @@ function App() {
 	 *
 	 * Added to fix #1056
 	 */
-	this.stop = function () {
+	this.stop = function (callback) {
 		for (const nodeHelper of nodeHelpers) {
 			if (typeof nodeHelper.stop === "function") {
 				nodeHelper.stop();
 			}
 		}
-		httpServer.close();
+		httpServer.close(callback);
 	};
 
 	/**
--- a/js/app.js
+++ b/js/app.js
@@ -222,9 +222,9 @@ function App() {
 				}
 			}
 
-			loadModules(modules, function () {
+			loadModules(modules, async function () {
 				httpServer = new Server(config);
-				const { app, io } = httpServer.open();
+				const { app, io } = await httpServer.open();
 				Log.log("Server started ...");
 
 				const nodePromises = [];
@@ -262,6 +262,8 @@ function App() {
 	 * exists.
 	 *
 	 * Added to fix #1056
+	 *
+	 * @param callback
 	 */
 	this.stop = function (callback) {
 		for (const nodeHelper of nodeHelpers) {
@@ -269,7 +271,7 @@ function App() {
 				nodeHelper.stop();
 			}
 		}
-		httpServer.close(callback);
+		httpServer.close().then(callback);
 	};
 
 	/**
--- a/js/app.js
+++ b/js/app.js
@@ -263,7 +263,7 @@ function App() {
 	 *
 	 * Added to fix #1056
 	 *
-	 * @param callback
+	 * @param {Function} callback Function to be called after the app has stopped
 	 */
 	this.stop = function (callback) {
 		for (const nodeHelper of nodeHelpers) {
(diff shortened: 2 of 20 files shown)


buxxi:
Hmm, it's so strange that it only fails on 18.x, seems to be when it uses the built in fetch instead of node-fetch. Changing that and it works...

khassel:
> Hmm, it's so strange that it only fails on 18.x, seems to be when it uses the built in fetch instead of node-fetch. Changing that and it works...

running the tests locally with `npm run test:e2e` the one ipWhitelist test fails as in the automated tests here.

But running the single test with `npx jest tests/e2e/ipWhitelist_spec.js` works ...

rejas:
Maybe something isnt shutdown correctly before the one test?

khassel:
> Maybe something isnt shutdown correctly before the one test?

yes, but it not depends on which test is running before, had different tests as predecessor ...

khassel:
if I put back the ugly "wait" in `stopApplication` it works ...

```js
exports.stopApplication = async () => {
        if (global.app) {
                return new Promise((resolve, reject) => {
                        global.app.stop(resolve);
                });
        }
        await new Promise((resolve) => setTimeout(resolve, 100));
        return Promise.resolve();
};
```

buxxi:
The fetch fails and there's no `catch()` in the wrapper in `global-setup.js` and that causes the test to wait forever for the resolve to happen (which it doesn't).

Using the following it fails more in a proper way:
```js
exports.fetch = (url) => {
	return corefetch(url);
};
```

But haven't yet found out why the fetch fails with native in node 18 and why it works with node-fetch...

buxxi:
Changed the tests to use `node-fetch` instead of the built-in fetch for node >= 18 because the built-in sets the header `Connection: keep-alive` between all requests which messes with the server starting up and shutting down between all the tests. Tried setting the header manually and setting a separate agent without any success. And it always logs that fetch is still an experimental feature, so I think we should treat it as that...

rejas (changes requested):
Looks fine to me. I would just like you to fill the missing jsdoc entries.

rejas:
Not sure if I requests @khassel to review, otherwise I would merge it

khassel:
with this we are using internal node fetch command (for node >=18) in the default modules but the tests are running with `node-fetch` instead. Means for me these are no tests for the modules (for node >=18).

buxxi:
- If the tests call production code on node >= 18 the internal fetch will still be used, it's just the fetches that is used only in the test code that would use `node-fetch`
- I don't believe the internal `fetch` is production ready and should probably not be used at all (yet) because:
	- Running the following:
	```javascript
	node -e "global.fetch('http://en.wikipedia.org').then((res) => console.log(res.status))"
	```
	Produces:
	```
	(node:71384) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
	(Use `node --trace-warnings ...` to show where the warning was created)
	200
	```
	- There's no documentation for it on https://nodejs.org/dist/latest-v18.x/docs/api/
	- The issue it creates with lingering connections like in this test is probably not wanted in the production code either
- Having tests depend on a randomly selected timeout and sometimes fail is really annoying, had that happen to me while debugging this with the old 200ms timeout.

khassel:
> I don't believe the internal fetch is production ready and should probably not be used at all (yet)

that was my point, I don't like it when we migrate the tests back but the rest not. If the internal fetch is now seen as not production ready it should be removed completly (and may evaluated later again, it is still experimental in node v19)

khassel:
but from my side we can merge this and remove internal fetch in another PR

rejas:
I am not so deep into the whole node/fetch thing but maybe what is done in the other PR (https://github.com/MichMich/MagicMirror/commit/1c8ea72e1e4e1cd6873a74eaae78a43c237d84cc#)
would be a solution? 

```
let nodefetch;
		if (nodeVersion < 18) {
			jest.mock("node-fetch");
			nodefetch = require("node-fetch");
		}
		let fetchMock;

```

But that is just my "gefährliches halbwissen" (dagnerously half-knowing :-)

buxxi:
Mocking the requests would probably make most of the tests useless as they verify the response from the server. Don't think that solution can be applied here.

I agree that it's better to do it in another PR.

Guess we have to wait a bit longer to get rid of that extra dependency :)

rejas:
I agree. It would be nice to get rid of a dependency but we should play it safe. 
So, whats the plan now? First a PR for getting rid of internal fetch, or first merge your PR?

khassel:
you can merge.

The other change would be [here](https://github.com/MichMich/MagicMirror/blob/7058fc5fd81674ef4496d7ff071db742c2d04c39/js/fetch.js#L10). I would suggest to comment out the `if/else` (but leave it in) and return always the `node-fetch` part. So we have to change nothing else and can insert the internal or other fetch functions there when needed.

can do this PR the next days

rejas (approved):

rejas:
> can do this PR the next days

Or I do it on a saturday evening? https://github.com/MichMich/MagicMirror/pull/2961 ;-)

PR #2957, opened 2022-10-25
Support HTTP headers with CORS-method

MagMar94:
Adds support for sending and receiving HTTP-headers when using the CORS-method. 

This change is required for the Yr weather-provider introduced in https://github.com/MichMich/MagicMirror/pull/2948.

To make it easier to add unit tests I moved the server-functions into a separate file.


MagMar94:
The tests fail because GitHub uses a different node-version. GitHub uses 16.17.1, whilst I used 19.0.0 locally. Will try to implement the tests to run on both.

package.json
```diff
@@ -82,6 +82,7 @@
 		"feedme": "^2.0.2",
 		"helmet": "^6.0.0",
 		"iconv-lite": "^0.6.3",
+		"isomorphic-fetch": "^3.0.0",
```
KristjanESPERANTO, line 85: Isn't that a dev dependency?

tests/unit/functions/server_functions_spec.js
```diff
@@ -0,0 +1,136 @@
+require("isomorphic-fetch");
```
KristjanESPERANTO, line 1: Maybe it's obvious, but I don't see it. What can `isomorphic-fetch` do that the current implementation of `fetch` cannot?

MagMar94: It won't be in the final version 👍 

It was an attempt to fix an error that occurred when I pushed to GitHub. Suddenly I got an [error](https://github.com/MichMich/MagicMirror/actions/runs/3322851239) saying the `Response` is not defined. I got [the idea to test it from StackOverflow](https://stackoverflow.com/questions/52420318/referenceerror-response-is-not-defined). The real reason turns out to be different node versions on GitHub and what I use locally, so [GitHub uses `node-fetch` instead of `global.fetch`](https://github.com/MichMich/MagicMirror/blob/9e0293047f8a8c0ced7d9fd8de1b839c06baf9cb/js/fetch.js).

package.json
MagMar94: Related to the [problem explained below](https://github.com/MichMich/MagicMirror/pull/2957#discussion_r1004858313) 😊 This won't be in the final version of the PR.

khassel:
> The tests fail because GitHub uses a different node-version. GitHub uses 16.17.1, whilst I used 19.0.0 locally. Will try to implement the tests to run on both.

we are running the tests here on 3 node version, 14+16+18. Our fetch function uses the internal fetch of Node if Node version >=18 and `node-fetch` for older versions. May related to your problem with 14+16.

MagMar94:
> > The tests fail because GitHub uses a different node-version. GitHub uses 16.17.1, whilst I used 19.0.0 locally. Will try to implement the tests to run on both.
> 
> we are running the tests here on 3 node version, 14+16+18. Our fetch function uses the internal fetch of Node if Node version >=18 and `node-fetch` for older versions. May related to your problem with 14+16.

Yep, working now 👍

tests/unit/functions/server_functions_spec.js
MagMar94: Removed all references to `isomorphic-fetch` now.

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+/**
+ * Gets headers and values to attatch to the web request.
+ *
+ * @param {string} url - The url containing the headers and values to send.
+ * @returns {object} An object specifying name and value of the headers.
+ */
+function GetHeadersToSend(url) {
```
rejas, line 64: Capitalized function names? I guess to mark them somehow as private? Or what is the reasoning here?

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/**
+ * Gets the headers expected from the response.
+ *
+ * @param {string} url - The url containing the expected headers from the response.
+ * @returns {string[]} headers - The name of the expected headers.
+ */
+function GeExpectedRecievedHeaders(url) {
+	let expectedRecievedHeaders = ["Content-Type"];
```
rejas, line 87: Can be a const, right?

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+			url = "invalid url: " + req.url;
+			Log.error(url);
+			res.send(url);
+		} else {
+			url = match[1];
+
+			const headersToSend = GetHeadersToSend(req.url);
+			const expectedRecievedHeaders = GeExpectedRecievedHeaders(req.url);
+
+			Log.log("cors url: " + url);
```
rejas, line 42: Do we really need to log this? Or just make it a debug maybe?

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param {Request} req - the request
+ * @param {Response} res - the result
+ */
+async function cors(req, res) {
+	try {
+		const urlRegEx = "url=(.+?)$";
+		let url = "";
+
+		let match = new RegExp(urlRegEx, "g").exec(req.url);
```
rejas, line 31: const?

rejas (changes requested):
some small things only

js/server_functions.js
khassel: I think this was there before and yes this should be logged for supporting stuff, its helpful as hint for supporters when reviewing log files of others e.g. in the forum.

js/server_functions.js
MagMar94: Whoops, had been writing C# all day. I'll fix it together with the changes requested below 👍

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+function getHeadersToSend(url) {
+	const headersToSend = { "User-Agent": "Mozilla/5.0 MagicMirror/" + global.version };
+	const headersToSendMatch = new RegExp("sendheaders=(.+?)(&|$)", "g").exec(url);
+	if (headersToSendMatch) {
+		const headers = headersToSendMatch[1].split(",");
+		for (const header of headers) {
+			const keyValue = header.split(":");
+			if (keyValue.length !== 2) {
+				throw new Error(`$Invalid format for header ${header}`);
```
rejas, line 72: $Invalid ? is that some C# notation too? ;-)

js/server_functions.js
```diff
@@ -0,0 +1,127 @@
...
+	if (typeof global.configuration_file !== "undefined") {
+		configFile = global.configuration_file;
+	}
+	html = html.replace("#CONFIG_FILE#", configFile);
+
+	res.send(html);
+}
+
+/**
+ * Gets the MacigMirror version.
```
rejas, line 118: small typo :-)

tests/unit/functions/server_functions_spec.js
```diff
@@ -0,0 +1,162 @@
...
+describe("server_functions tests", () => {
+	describe("The cors method", () => {
+		let fetchResponse;
+		let fetchResponseHeadersGet;
+		let fetchResponseHeadersText;
+		let corsResponse;
+		let request;
+
+		let nodefetch;
+		if (nodeVersion < 18) {
```
rejas, line 13: Now that we reverted the native fetch implementation (https://github.com/MichMich/MagicMirror/pull/2961) do we still need this?

rejas (changes requested):
Unfortunately some merge conflicts now after yesterays merges. See https://github.com/MichMich/MagicMirror/pull/2961 and https://github.com/MichMich/MagicMirror/pull/2952 for node18 and fetch disuccsion.

tests/unit/functions/server_functions_spec.js
MagMar94: No, implemented everything for `node-fetch` now. Almost sad after finally getting the tests to pass for all node versions 😅

PR #2959, opened 2022-10-27
Cleanup jest config

rejas:
Some small cleanups with regards to jest
- call jest directly (nyc is integrated in jest these days)
- move jest config into seperate file so we dont clutter up the package.json
- remove empty test file for newsletter-unit-tests
- update dependencies that touch jest
- try out v8 as coverageProvider


package.json
```diff
@@ -50,16 +50,16 @@
...
+		"eslint-plugin-jest": "^27.1.3",
 		"eslint-plugin-jsdoc": "^39.3.6",
 		"eslint-plugin-prettier": "^4.2.1",
 		"express-basic-auth": "^1.2.1",
 		"husky": "^8.0.1",
-		"jest": "^29.1.2",
+		"jest": "^29.2.2",
 		"jsdom": "^20.0.1",
 		"lodash": "^4.17.21",
 		"nyc": "^15.1.0",
```
khassel, line 61: is `nyc` still needed here?

rejas: good point, removed it now

khassel (approved):

rejas:
FYI: I think the coverage goes down so drastically because now we really check all js files. Prior to this PR the list of files was a lot less: 
https://github.com/MichMich/MagicMirror/actions/runs/3282686049/jobs/5406418456

khassel:
this test did never produce any logical output so I'm ignoring this (and therefore didn't see the great difference) ...

rejas:
:-D Lets hope they stay now a little more consistent.

PR #2962, opened 2022-10-30
Make the app start using async/await instead of callbacks

buxxi:
- Modernizing the startup using async/await instead of callbacks.
- Since it's not blocking now at startup I had to move the creation of the app/server into the `ready`-event for electron, otherwise the screen was black at startup.
- Moved the repeated code for timeouts when calling `app.close()` as an argument to that method instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Special thanks to: @rejas, @sdetweil
   - Reworked how weatherproviders handle units (#2849)
   - Use unix() method for parsing times, fix suntimes on the way (#2950)
   - Refactor conversion functions into utils class (#2958)
+- Use async/await for startup of the application instead of callbacks
 
 ### Fixed
 
--- a/js/app.js
+++ b/js/app.js
@@ -258,9 +258,10 @@ function App() {
 	 *
 	 * Added to fix #1056
 	 *
+	 * @param {number} timeout the amount of milliseconds before the returned promise should be automatically resolved
 	 * @returns {Promise} A promise that is resolved when all node_helpers and the http server has been closed
 	 */
-	this.stop = async function () {
+	this.stop = async function (timeout) {
 		for (const nodeHelper of nodeHelpers) {
 			if (typeof nodeHelper.stop === "function") {
 				nodeHelper.stop();
@@ -272,7 +273,18 @@ function App() {
 			return Promise.resolve();
 		}
 
-		return httpServer.close();
+		let serverClosePromise = httpServer.close();
+
+		// If a timeout is set, resolve when the server is closed or the timeout has been reached
+		if (timeout) {
+			let timeoutPromise = new Promise((resolve) => {
+				setTimeout(resolve, timeout);
+			});
+
+			return Promise.race([serverClosePromise, timeoutPromise]);
+		} else {
+			return serverClosePromise;
+		}
 	};
 
 	/**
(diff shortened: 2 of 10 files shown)


buxxi:
Hmm, works on my machine with node `16.17.1` and `14.20.1` which these checks run...

rejas:
Yeah, the tests are still somehow fragile.... Sometimes re-running helps, sometimes not...

buxxi:
And now 2 other fails instead, great...

`global.page` seems to be null sometimes here: https://github.com/MichMich/MagicMirror/blob/4d47c0837ff5bb853664fac599dd182ef1628c37/tests/electron/helpers/global-setup.js#L39

I could see that happen if this if-statement fails: https://github.com/MichMich/MagicMirror/blob/4d47c0837ff5bb853664fac599dd182ef1628c37/tests/electron/helpers/global-setup.js#L15

According to https://playwright.dev/docs/api/class-electronapplication#electron-application-windows `windows()` doesn't return a promise, so awaiting for it doesn't do anything. I wonder if this if-statement is even needed?

khassel:
yeah, tests are really fun, you make changes and they all run locally and then they began to fail on github, but not always ...

> I wonder if this if-statement is even needed?

try it without, the next statement is trying to get first window so we will see whats happens ...

rejas:
Trying your branch out, I also got (only once!) some errors in the electron tests where the date is used (like compliments or weather sunset)... But now two runs without errors... SOme weird race conditions?

buxxi:
Most likely since the whole startup was blocking the thread before and NodeJS is single threaded. So the server startup probably blocked electron from receiving any events before that was done.

rejas:
> According to https://playwright.dev/docs/api/class-electronapplication#electron-application-windows `windows()` doesn't return a promise, so awaiting for it doesn't do anything. I wonder if this if-statement is even needed?

There are two more "await window()" calls in the tests, one in electron/env_spec the other in electron/helpers/global-setup could you get rid of them too?

buxxi:
Think I understand what it was trying to check, when I removed this if-statement it failed once because the test that opens the dev console had the dev console as the first window and not the magicmirror. So probably would need to wait for each window until the correct one appears.

That doesn't explain the error on github actions. will have to look into it more tomorrow.

buxxi:
Finally...

Please retry the checks a few times in case this was a fluke :)

rejas:
Bad news, the tests/electron/modules/calendar_spec.js tests dont fail if change the date (the weather and compliments seem to be fine and fail if I change the dates) :-( Could you look into it?

buxxi:
It's the same in `develop`-branch, so it's nothing that I broke :)

buxxi:
The tests will now wait until timeout if the css-classes are missing, not optimal...

buxxi:
Been running the electron-tests in a loop for over an hour now without any of them failing locally... This is a bit annoying

khassel:
I spent hours with the tests, running locally means nothing ... every time I thought now they are o.k. I did a run on Github and they failed there ... The timing seems totally different, the stuff on Github is slower and sometimes tests there fails because of the 20 sec. timeout which never happens locally. You can use your `develop` branch in your fork and push it to your fork on Github to run the tests (codecov will fail but the rest runs) so you can check Github without using this repo (and showing everyone the fails). From the `actions` tab you can rerun tests without pushing again, thats whats I'm doing after changing tests ...

buxxi:
Well, I'm totally out of ideas so I guess it's time to give up on this PR...

buxxi:
I wont have any time in in the nearby months to continue this so closing the PR.

PR #2963, opened 2022-11-02
remove "x-frame-options" and "content-security-policy" response headers

Tom-Hirschberger:
Many users like me do have the problem that they want to embed other sites to their mirror by "iframe".
As some developers set the "x-frame-options" and "content-security-policy" for security reasons these sites can not be embedded.
Electron provides the "webview" element additionally to "iframe" which allows to embed these sites although. The main difference is that a new process is started which handles the "webview" element.
BUT: As the "webview" process needs to be started and is isolated "webview" is slower and the elements can not be accessed from the embedding website.

As an alternative i implemented a small callback function in electron.js which removes the response headers that forbid the embedding.

The removing can be controlled with the new config options:
* ignoreXOriginHeader
* ignoreContentSecurityPolicy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Special thanks to: @rejas, @sdetweil, @MagMar94
 - Added css class names "today" and "tomorrow" for default calendar
 - Added Collaboration.md
 - Added new github action for dependency review (#2862)
+- Added config options "ignoreXOriginHeader" and "ignoreContentSecurityPolicy"
 
 ### Removed
 
--- a/js/electron.js
+++ b/js/electron.js
@@ -103,6 +103,20 @@ function createWindow() {
 			}, 1000);
 		});
 	}
+
+	//remove response headers that prevent sites of being embedded into iframes if configured
+	mainWindow.webContents.session.webRequest.onHeadersReceived((details, callback) => {
+		let curHeaders = details.responseHeaders;
+		if (config["ignoreXOriginHeader"] || false) {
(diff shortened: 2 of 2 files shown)


sdetweil:
interesting. but doesn't that violate the source's content policies?

Tom-Hirschberger:
It might violate but it it does not necessarily.
Most time the header is set to prevent of XSS attacks and not to avoid embedding. But a XSS attack at the electron instance providing the mirrors page is very unlikely.

khassel:
if I remember this correctly we had already such issues and we inserted a config option `httpHeaders` for this.

Have you tested iframe embedding with these

```js
let config = {
	address: "localhost",
	port: 8080,
	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false, frameguard: false },
	...
}
```

`httpHeaders`?

Tom-Hirschberger:
Did a quick check in server.js and the Changelog.
The httpHeaders control can be used to control if the MagicMirror page can be embedded with a Iframe.
But they can not be used to control the headers of pages that are embedded to the mirror page.

khassel:
thanks for clarifying, had the older iframe change in mind and wanted to avoid doing something which is already implemented.

khassel (approved):

rejas (approved):

Tom-Hirschberger:
Thank you very much.

PR #2964, opened 2022-11-03
Added a WeatherProvider for Open-Meteo

angeldeejay:
## Added Weather Provider for Open-Meteo.

I've found a completely free weather REST API (event with option of self-hosting) after having problems with API keys from all MagicMirror weather providers currently implemented (the remote services, not the providers themselves).

This API doesn't return information about reverse geocode from latitude and longitude options like others. I solved that issue using another free API.

### APIs used
- [Open-Meteo Weather Forecast API](https://open-meteo.com/en/docs)
- [BigDataCloud’s Free Client-Side Reverse Geocoding API](https://www.bigdatacloud.com/docs/api/free-reverse-geocode-to-city-api)

### Considerations
- This provider is config reliable so, be free to use the same config you can found in the official MagicMirror Weather module documentation. 
- This module config skips the `apiKey` parameter. It's not used at all. Only `latitude` and `longitude` are required.

#### Config examples:
```
modules: [
  {
    module: "weather",
    position: "top_right",
    header: "Weather Forecast",
    config: {
      updateInterval: <number here>, 
      weatherProvider: "openmeteo",
      type: "current",
      lat: <number here>,
      lon: <number here>,
      showHumidity: true,
      showWindDirectionAsArrow: true,
      showWindDirection: true,
      degreeLabel: true,
    }
  },
  {
    module: "weather",
    position: "top_right",
    header: "Weather Forecast",
    config: {
      updateInterval: <number here>,
      weatherProvider: "openmeteo",
      type: "daily",
      lat: <number here>,
      lon: <number here>
      colored: true,
      maxNumberOfDays: <number here>,
      showPrecipitationAmount: true,
      appendLocationNameToHeader: true
    }
  },
  {
    module: "weather",
    position: "top_right",
    header: "Weather Forecast",
    config: {
      updateInterval: <number here>,
      weatherProvider: "openmeteo",
      type: "hourly",
      lat: <number here>,
      lon: <number here>,
      maxEntries: <number here>,
      showPrecipitationAmount: true,
      degreeLabel: true,
      appendLocationNameToHeader: true
    }
  },
]
```


angeldeejay:
@rejas can you please take a look? Thanks!

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,549 @@
...
+	 * @param {object} config The configuration object
+	 */
+	setConfig(config) {
+		this.config = {
+			lang: config.lang ?? "en",
+			...this.defaults,
+			...config
+		};
+
+		Log.log(this.config);
```
rejas, line 220: No real need to log this here

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,549 @@
...
+			})
+			.catch((request) => {
+				Log.error("Could not load data ... ", request);
+			});
+	},
+
+	// Implement WeatherDay generator.
+	generateWeatherDayFromCurrentWeather(weather) {
+		const now = moment();
+		const i = 0;
```
rejas, line 376: why have that here? seems redundant

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,549 @@
...
+	defaults: {
+		apiBase: "https://api.open-meteo.com/v1",
+		lat: 0,
+		lon: 0,
+		past_days: 0,
+		type: "current",
+		timezone: null
+	},
+
+	units: {
```
rejas, line 31: We recentyl changed how the weatherproviders handle units:
Please use internally ony metric units (celsius for temperature, meters per second for wind). So you have to make sure that either a) the api returns metric units or b) cif it returns imperial, convert them when you get them

The weather module will then convert the units from the weatherpovider once they are about to be displayed into what the user has configured. So nothing for the provider to do here

rejas (changes requested):

modules/default/weather/providers/openmeteo.js
angeldeejay: To be clear:
- Units will be "stored" in weather object only in metric values for all providers?

angeldeejay:
@rejas Changes suggested marked as solved

rejas:
Trying out your PR I get an error with this config:

```
{
	module: "weather",
	//disabled: true,
	position: "top_right",
	config: {
		weatherProvider: 'openmeteo',
		type: "current",
		lat: 52.520008,
		lon: 13.404954,
	}
},
```

Error is:
```
Uncaught TypeError: moment.tz is undefined [openmeteo.js:218:5](http://localhost:8080/modules/default/weather/providers/openmeteo.js)
```

It seems it tries to guess the timezone? Is that really necessary for a current weather call?

angeldeejay:
I'll take a look. Thx

angeldeejay:
@rejas I've removed support for moment-timezone in favor of API Documentation's param `timezone` with 'auto' value supported now. Timezone is provided automatically by the OpenMeteo API using coordinates given in payload.

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,541 @@
...
+	 * @param {object} config The configuration object
+	 */
+	setConfig(config) {
+		this.config = {
+			lang: config.lang ?? "en",
+			...this.defaults,
+			...config
+		};
+
+		this.config.ignoreToday = (this.config.ignoreToday ?? false) && ["daily", "forecast"].includes(this.config.type);
```
rejas, line 206: IgnoreToday is not needed in the providers, its used when displaying the data

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,541 @@
...
+			past_days: this.config.past_days ?? 0,
+			daily: this.dailyParams,
+			hourly: this.hourlyParams,
+			// Fixed units as metric
+			temperature_unit: "celsius",
+			windspeed_unit: "kmh",
+			precipitation_unit: "mm"
+		};
+
+		const startDateOffset = this.config.ignoreToday ? 1 : 0;
```
rejas, line 240: Dont need to offset the startDate when ignoreToday is set, the display-tempalte does that for you already

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,541 @@
...
+			latitude: this.config.lat,
+			longitude: this.config.lon,
+			timeformat: "unixtime",
+			timezone: "auto",
+			past_days: this.config.past_days ?? 0,
+			daily: this.dailyParams,
+			hourly: this.hourlyParams,
+			// Fixed units as metric
+			temperature_unit: "celsius",
+			windspeed_unit: "kmh",
```
rejas, line 236: The wind units need to be stored in meters per second in the weatherobject, not kmh. Please convert them if the api doesnt provide that

modules/default/weather/providers/openmeteo.js
```diff
@@ -0,0 +1,541 @@
...
+		const today = moment().startOf("day");
+		const startDate = moment(today).add(startDateOffset, "days");
+
+		switch (this.config.type) {
+			case "hourly":
+				params["start_date"] = startDate.format("YYYY-MM-DD");
+				params["end_date"] = moment(startDate).add(this.config.maxNumberOfDays, "days").format("YYYY-MM-DD");
+				break;
+			case "daily":
+			case "forecast":
```
rejas, line 250: I am trying this in my config:
```
		{
			module: "weather",
			position: "top_right",
			config: {
				weatherProvider: 'openmeteo',
				ignoreToday: true,
				type: "forecast",
				lat: 52.520008,
				lon: 13.404954,
			}
		},
```
but dont get any data displayed. Using 
```
				type: "forecast",
```
works though, Bug or Feature?

rejas (changes requested):

modules/default/weather/providers/openmeteo.js
angeldeejay: A bug. Already fixed on line 303

modules/default/weather/providers/openmeteo.js
rejas: This isnt fixed yet

rejas:
Since the year is rapidly nearing its end (and therefore a new release is coming up soon) I would propose to merge this even though it still has a bug (or two depending if I caught them all).
The reasoning behind it is:
- this PR doesnt touch any other code, therefor wont break anything
- it adds a new feature/provider which works for most use cases
- releasing it now will get more reach for testing so more bugs can be found

Any comments @angeldeejay @khassel @sdetweil @MichMich ?

rejas:
Merging this in the holiday spirit :-)

@angeldeejay would be nice if you could add some documentation to the doc repo for this provider (https://github.com/MichMich/MagicMirror-Documentation) and in the new year see if you can fix the remaining bugs. Happy holidays everyone!

angeldeejay:
> hi @angeldeejay I wanted to check in if you are still interested in updating the documentation or if you need support for that?

Off course!, I had some personal issues the last month, but I'll be updating documentation for this provider. Sorry

angeldeejay:
@rejas Please refer to:

- [Documentation PR](https://github.com/MichMich/MagicMirror-Documentation/pull/158)
- [Technical Debt from this PR (wind speed units)](https://github.com/MichMich/MagicMirror/pull/3054)

PR #2973, opened 2022-11-26
Replace `&hellip;` with `…`

KristjanESPERANTO:
I think it is clearer if we don't use the HTML entity.


rejas:
Maybe we can remove the safe markers for nunjucks now? see https://github.com/MichMich/MagicMirror/pull/2427

khassel:
> Maybe we can remove the safe markers for nunjucks now? see https://github.com/MichMich/MagicMirror/pull/2427

so should this be done with this PR oder in another one?

rejas:
I'd prefere a second PR if one of you could do it. Dont think it should go into the next release since it might need some feedback if it breaks something.

khassel (approved):

PR #2991, opened 2023-01-05
Fix badge status

novialriptide:


rejas (approved):
Probably best to merge it into master like the PR wants to. Any objections @MichMich ?

khassel:
should be available on `master` asap.

We could merge this on `develop` as usual and then cherry pick this commit on `master`.

rejas:
Sounds like a plan. Do you want to?

khassel:
> Sounds like a plan. Do you want to?

o.k.

khassel (approved):

khassel:
did it the other way (merge on `master`, cherry pick on `develop`) because I had 162 changes in the web ui after changing base to `develop` ...

MichMich:
Sorry for my slow response. My bad. But for future PR's I think we should refrain from any merge into master if it's not solving a blocking issue. 

This PR into master caused an update notification for all users which seem to have caused a lot of confusion (especially since it's not a tagged commit). 

I really try to limit any update to master to the four releases per year to prevent update-fatigue. :)

rejas:
shall we add the "never ever merge to master" rule to our Collaboration.md @sdetweil @khassel ?

khassel:
and "only MichMich does releases" ...

I think I have a solution for "never ever merge to master". I updated the `updatenotification` module (still draft)

- if repo = MagicMirror repo
- and branch = `master`
- and behind head of `master`
- check if there is a tag on the commits between current commit and head of `master`
- and only send a notification if so

This would allow us to update things on `master` without user getting notified.

sdetweil:
i think never commit to master unless it is a hotfix and changes the version number and is tagged

certainly no reason to update the readme

rejas:
> and "only MichMich does releases" ...
> 
> I think I have a solution for "never ever merge to master". I updated the `updatenotification` module (still draft)
> 
>     * if repo = MagicMirror repo
> 
>     * and branch = `master`
> 
>     * and behind head of `master`
> 
>     * check if there is a tag on the commits between current commit and head of `master`
> 
>     * and only send a notification if so
> 
> 
> This would allow us to update things on `master` without user getting notified.

Nice idea, but this has got to be very foolproof and well tested. Imagine the update notifications being never shown again and therefore no bugfixes will get pulled...

So maybe just go the "never merge to master unless a hotfix with a new version tag is necessary" route?

khassel:
> So maybe just go the "never merge to master unless a hotfix with a new version tag is necessary" route?

thats the current situation (at least until next release) ...

> but this has got to be very foolproof and well tested

agree, did already some testing. Will do a draft PR (not yet, depends on another upcoming PR) and the you can take a look ...

novialriptide:
> This PR into master caused an update notification for all users which seem to have caused a lot of confusion (especially since it's not a tagged commit).

I'm confused, does this mean everyone using MagicMirror2 has received a notification about this pull request?

sdetweil:
yes. the updatenotification module was checking commits, and this one was a+1, so it popped up a change notice on every mirror running the notification module.  

and my update script failed  cause the version didn't change.

novialriptide:
> yes. the updatenotification module was checking commits, and this one was a+1, so it popped up a change notice on every mirror running the notification module.  
> 
> and my update script failed  cause the version didn't change. 

I feel famous

khassel:
> I feel famous

me too, because I did the push 😚

PR #3002, opened 2023-01-11
added possibility to ignore MagicMirror repo in updatenotification

khassel:
was [requested in the forum](https://forum.magicmirror.builders/topic/17519/updatenotification).

- added possibility to exclude MagicMirror Repo and renamed it from `default` to `MagicMirror`
- improved getting `behind` in case a hard `git fetch` was already done
- removed test "excludes repo if refs don't match regex" because of above improvement this case is obsolete
- improved `git fetch --dry-run` with `-n` option to exclude tags (noise reduction)

### DIFF
--- a/tests/unit/functions/updatenotification_spec.js
+++ b/tests/unit/functions/updatenotification_spec.js
@@ -57,7 +57,7 @@ describe("Updatenotification", () => {
 				return { stdout: gitRevParseOut, stderr: gitRevParseErr };
 			} else if (command.includes("git status -sb")) {
 				return { stdout: gitStatusOut, stderr: gitStatusErr };
-			} else if (command.includes("git fetch --dry-run")) {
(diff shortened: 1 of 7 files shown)


tests/unit/functions/updatenotification_spec.js
```diff
@@ -108,13 +108,6 @@ describe("Updatenotification", () => {
 			const { error } = require("logger");
 			expect(error).toHaveBeenCalledWith(`Failed to retrieve repo info for ${moduleName}: Failed to retrieve status`);
 		});
-
-		it("excludes repo if refs don't match regex", async () => {
```
rejas, line 112: why was that removed?

khassel: see description

the case "no regex match" is now handled different, before an empty object was returned, now another call is done and it is always an object returned. So this test makes no sense anymore.

PR #3004, opened 2023-01-14
updatenotification: only notify mm-repo for master if tag present

khassel:
see discussion here: https://github.com/MichMich/MagicMirror/pull/2991#issuecomment-1376372720

I still see a need for updating `master` in special cases (e.g. correct errors in README.md which would otherwise be present up to 3 month until next release) so with this PR **only for MagicMirror repo** and **only for `master` branch** updatenotifications are only triggered if at least one of the new commits has a tag.

May @MichMich must decide if this is wanted.


MichMich:
Looks good to me! 👍🏻

rejas (approved):
If it looks good for @MichMich then it looks good for me too :-)

PR #3008, opened 2023-01-15
Update weather tests

rejas:
Added a few tests for sunset/sunrise and feelsLike 
Lets see if they run through first...


rejas:
Tests failing, therefor back to the draft...

rejas:
Somhow the evaluate fucntion for mocking the Date here: https://github.com/MichMich/MagicMirror/blob/2fec314ff51667be54da59ccaf38991ad805d3f6/tests/electron/helpers/global-setup.js#L22
is never called @khassel therefore the time is never faked... Any idea why?

khassel:
> is never called @khassel therefore the time is never faked... Any idea why?

no, I think it is called, otherwise weather and compliment tests would fail.

You have this error in the electron tests:

```bash
    TypeError: Cannot read properties of null (reading 'locator')

      40 | exports.getElement = async (selector) => {
      41 | 	expect(global.page);
    > 42 | 	let elem = global.page.locator(selector);
```

After digging around I found that the line `jest.retryTimes(3);` causes running the tests several times.
So please remove this line (can you do this in this PR? Otherwise I have to make another one ...) and run the tests again.

Now you are getting 

```bash
 FAIL   electron  tests/electron/modules/weather_spec.js
  Weather module
    Current weather with sunrise
      ✕ should render sunrise (369 ms)
    Current weather with sunset
      ✓ should render sunset (234 ms)

  ● Weather module › Current weather with sunrise › should render sunrise

    expect(received).toBe(expected) // Object.is equality

    Expected: "7:00 am"
    Received: "3:45 pm"
```

which is may caused by your changes in this PR (tests are o.k. on `develop`).

rejas:
Found the "issue" with tests. Introduced it myself in the first commit :-) Had to work around to get electron and playwright tests both work out. But now its ready for review :-)

khassel (approved):

PR #3009, opened 2023-01-15
Add (in Alert module) templateName parameter

rajniszp:
This simple change allows to use your own templates (under "templates" directory). The parameter `templateName` is optional (ignored on falsy value - undefined, null, empty string, etc.) and independent of `type`.


sdetweil:
@rajniszp could you change the title to add, 

in alert module

rajniszp:
Sure, changed

PR #3016, opened 2023-01-19
Refresh the Calendar DOM every minute

psieg:
This keeps relative dates accurate when the calendar's fetch frequency is much larger than a minute.
As fetching incurs network traffic and load on servers and most calendars don't update that often, simply refreshing locally is enough.

When using relative for today's events, dates will show as "in X minutes" or "ends in X minutes" for events within an hour and this goes out of date quickly. It's weird to see that the time is, say, 16:30 and an event that you know ends at 16:45 is shown to "ends in 23 minutes" because that's when the last fetch happened.

Please forgive me if there's style issue, I don't have npm set up on my machine to run the formatter.

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -86,6 +86,9 @@ Module.register("calendar", {
 
 	// Override start method.
 	start: function () {
+		const ONE_SECOND = 1000; // 1,000 milliseconds
+		const ONE_MINUTE = ONE_SECOND * 60;
+
 		Log.info("Starting module: " + this.name);
 
 		// Set locale.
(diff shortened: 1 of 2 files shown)


modules/default/calendar/calendar.js
```diff
@@ -131,6 +134,14 @@ Module.register("calendar", {
 			// fetcher till cycle
 			this.addCalendar(calendar.url, calendar.auth, calendarConfig);
 		});
+
+		// Refresh the DOM every minute if needed: When using relative date format for events that start
+		// or end in less than an hour, the date shows minute granularity and we want to keep that accurate.
+		setTimeout(() => {
```
rejas, line 140: why is that setTimeout wrapped around setInterval?

modules/default/calendar/calendar.js
```diff
@@ -86,6 +86,9 @@ Module.register("calendar", {
 
 	// Override start method.
 	start: function () {
+		const ONE_SECOND = 1000; // 1,000 milliseconds
+		const ONE_MINUTE = ONE_SECOND * 60;
```
rejas, line 90: const ONE_MINUTE = 60000;

is clear enough :-)

rejas:
Good idea and PR. Just some minor clean/clear ups :-)

modules/default/calendar/calendar.js
psieg: The idea is to trigger the calendar refresh on the minute change such that the time is always accurate. If you just use an interval, you will off by half a minute on average :)

modules/default/calendar/calendar.js
psieg: This was in line with the rest of the file, I don't mind.

rejas (approved):

Issue #3019, opened 2023-01-21
Weather module: Wind arrow is inverted


MagMar94:
If using the `showWindDirectionAsArrow`-setting, the arrow points at where the wind is blowing from. I feel it is more appropriate to point it in the direction of the wind. This is how wind arrows usually work in weather forecasts.

MagMar94:
I am going to fix this, but should I hide it behind a setting and leave the current arrow as the default? That way no one will get the change forced on them. However, I think most people want the inverted arrow.

rejas:
I also think that we should have the arrow pointed in the direction the wind is going. That makes most sense. Thanks for your report (and hopefully your fix ;-)

rejas:
Thx for your PR!

PR #3020, opened 2023-01-21
Fix yr weather direction

MagMar94:
Fixes so the Yr weather direction is not inverted. This error was [reported in the community](https://forum.magicmirror.builders/topic/17561/wind-direction-180-degrees-twisted-at-yr).

I misunderstood when developing the module because of the wind direction arrow was pointing opposite to the arrow displayed on Yr.no. I renamed the variable to help other developers in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Fix weathergov provider hourly forecast (#3008)
 - Fix message display with HTML code into alert module (#2828)
 - Fix typo into french translation
+- Yr wind direction is no longer inverted
 
 ## [2.22.0] - 2023-01-01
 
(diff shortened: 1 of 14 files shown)


rejas:
Will have a look at this later when home (and if there are tests for this) Have you checked if every provider handles wind direction the same?

MagMar94:
> Have you checked if every provider handles wind direction the same?

No, but I only changed the variable assignment in the Yr-provider.

I renamed the variable for everyone to make It easier to understand what value the weather module expects. If a provider returns the wrong value it will result in wrong cardinal directions.

I missed this when I implemented the provider, because I matched the arrow (`showWindDirectionAsArrow`) to what was displayed on Yr.no with the mirror. On Yr (and other weather services) arrows points in the direction the wind is blowing, but the current MM-implementation points to where the wind is blowing from. I will fix this in another PR referencing https://github.com/MichMich/MagicMirror/issues/3019. I created the issue to get some feedback before I fix it.

rejas (approved):

PR #3023, opened 2023-01-22
Tidy up precipitation

### LINKED ISSUE #2953: Cleanup precipation usage in weather module
Preipation seems to be used in different ways in the weatherproviders:

some are using it as amount of precipation, some are using it as a percentage (risk of rain)

Whoever takes this ticket should look if the usage and display of these values can be optimized:
- split up in two config options?
- update diplsay layout (german word for riskofrain is waaaaaaaays to long, making me think nobody really uses this feature)
- whatever they else find

MagMar94:
Fixes #2953

This is an attempt to fix the issue with precipitation amount and percentage mixup. I have created a separate `precipitationPercentage`-variable where the probability of rain can be stored.

The config options now has the old `showPrecipitationAmount` in addition to a new setting: `showPrecipitationProbability` (shows the likelihood of rain).

<details>
  <summary>Examples</summary>
  
  ### Yr

I tested the Yr weather provider for a Norwegian city Bergen that has a lot of rain. I have removed properties that are irrelevant for this demo from the config-samples below.

Config:
  ```js
{
	module: "weather",
	config: {
			weatherProvider: "yr",
			type: "current",
			showPrecipitationAmount: true,
			showPrecipitationProbability: true
	}
},
{
	module: "weather",
	config: {
			weatherProvider: "yr",
			type: "hourly",
			showPrecipitationAmount: true,
			showPrecipitationProbability: true
	}
},
{
	module: "weather",
	config: {
			weatherProvider: "yr",
			type: "daily",
			showPrecipitationAmount: true,
			showPrecipitationProbability: true
	}
}
  ```

Result:<br/>
screenshot

</details>


rejas:
Thanks for the PR, looks good from the outside so far. 

Two favors, if I might ask:
- Can you check the other providers what they actually provide in the fields used? IIRC some do amount some percentage...
- Can you add a test for displaying the amount / percentage?

MagMar94:
I checked the various docs and tried some configs, and this is what I found:

| Provider       | Amount support  | Probability support    |
| ------------ | ----------------- | -------------- |
| envcanada   | ✅                        | ✅                    |
| openmeteo  | ✅                        | ❌                    |
| openweathermap | ✅               | ✅                   |
| smhi             | ✅                         | ❌                    |
| ukmetoffice | ❌                         | ✅                   |
| ukmetofficedatahub | ❌          | ✅                    |
| weatherbit   | ✅                         | ✅                   |
| weatherflow | ❌ (possible in hourly, but method not implemented) | ✅ |
| weathergov | ❌                          | ✅                    |
| Yr                  | ✅                        | ✅                    | 

I have implemented this in all providers, but I don't have all the API keys. It shouldn't be breaking though, worst case the new field does not get assigned correctly.

Another question is what to to with the `WeatherObject.rain` and `WeatherObject.snow`-properties. These are not used in the weather-module, but some providers assign to the properties and sum them together in the `precipitationAmount`-field. This does not require the variables to be defined in the global `WeatherObject`, so I just removed them.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ _This release is scheduled to be released on 2023-04-01._
 
 - Use develop as target branch for dependabot
 - Update issue template and contributing doc
+- The weather modules clearly separates precipation amount and probability (risk of rain/snow)
+  - This requires all providers that only supports probability to change the config from `showPrecipitationAmount` to `showPrecipitation` or `showPrecipitationProbability`.
```
rejas, line 31: Those config options need to be added to the default config: https://github.com/MichMich/MagicMirror/blob/42d42ef4526be1f74ef9570a81e7027592e072f5/modules/default/weather/weather.js#L11

modules/default/weather/providers/openweathermap.js
```diff
@@ -282,7 +282,8 @@ WeatherProvider.register("openweathermap", {
 				}
 			}
 
-			weather.precipitation = weather.rain + weather.snow;
+			weather.precipitationAmount = weather.rain + weather.snow;
+			weather.precipitationProbability = weather.pop * 100;
```
rejas, line 286: weather.pop is never defined, so this is always empty...

translations/de.json
```diff
@@ -26,7 +26,9 @@
 	"NNW": "NNW",
 
 	"FEELS": "Gefühlt {DEGREE}",
-	"PRECIP": "Niederschlagswahrscheinlichkeit",
+	"PRECIP": "Precipitation",
```
rejas, line 29: "Niederschlag"

translations/de.json
```diff
@@ -26,7 +26,9 @@
...
+	"PRECIP_POP": "Niederschlagswahrscheinlichkeit",
+	"PRECIP_AMOUNT": "Precipitation",
```
rejas, line 31: "Niederschlagsmenge" 

(we need shorter words for the german stuff OR need to do a new layout where precipation is displayed, currently its in one line with feelslike an d takes too much space there) but thats another issue ;-)

rejas (changes requested):
Getting there, just a few small things! Great work so far!

translations/de.json
MagMar94: I agree, we have the same problem in Norwegian. As a possible workaround I added a `<span>`-element around it so it can be easily hidden with CSS.

modules/default/weather/providers/openweathermap.js
MagMar94: Oops, I was supposed to target the `forecast` ([documentation](https://openweathermap.org/forecast16)).<br/>
Screenshot of documentation

modules/default/weather/providers/envcanada.js
```diff
@@ -508,17 +501,17 @@ WeatherProvider.register("envcanada", {
...
 		if (foreGroup[today].querySelector("precipitation accumulation")) {
-			weather.precipitation = foreGroup[today].querySelector("precipitation accumulation amount").textContent * 1.0;
+			weather.precipitationAmount = foreGroup[today].querySelector("precipitation accumulation amount").textContent * 1.0;
 
 			weather.precipitationUnits = " " + foreGroup[today].querySelector("precipitation accumulation amount").getAttribute("units");
 
 			if (this.config.units === "imperial") {
 				if (weather.precipitationUnits === " cm") {
-					weather.precipitation = (weather.precipitation * 0.394).toFixed(2);
+					weather.precipitationAmount = (weather.precipitationAmount * 0.394).toFixed(2);
```
rejas, line 510: as a rule of thumb we dont want the weather providers to check for config.units and do conversions stuff based on that value.

instead all providers should store the data in metric units in the object (and the conversion based on the config happens only on the "frontend part" (aka in the njk file and when it gets the values from the weatherobject).

could you change the logic here and in other providers too?

rejas (changes requested):

modules/default/weather/current.njk
```diff
@@ -54,16 +54,32 @@
             </div>
         {% endif %}
     </div>
-    {% if (config.showFeelsLike or config.showPrecipitationAmount) and not config.onlyTemp %}
+    {% if (config.showFeelsLike or config.showPrecipitation or config.showPrecipitationAmount or config.showPrecipitationProbability) and not config.onlyTemp %}
```
rejas, line 57: not happy with these if constructs...

can we get rid of the the old "config.showPrecipitationAmount" here by setting the other values correctly in weatherjs when starting up? something similar like we did for useKmh / useBeaufort https://github.com/MichMich/MagicMirror/blob/684c727430586d88a48202f9276f63ed0ae3decc/modules/default/weather/weather.js#L79

rejas (changes requested):

modules/default/weather/providers/envcanada.js
MagMar94: Great idea 👍 We now support mm (default) and cm as input units. This fixes the conversion for all providers that haven't bothered with the calculation in the first place.

modules/default/weather/current.njk
MagMar94: I agree, but I'm not sure removing `config.showPrecipitationAmount`. Personally I consider using `config.showPrecipitationAmount` or `config.showPrecipitationProbability`. I find the `config.showPrecipitation` to take too much horizontal space. Have a look at the examples in the description. Maybe we don't need to offer displaying both values with `config.showPrecipitation`?

On the other hand, the probability can be hidden with CSS (`.precipitationProbability { display: none; }`). Maybe this is enough? If it is we can consider removing `config.showPrecipitationProbability` as well? By adding another `<span>` around the parenthesis it can be hidden together with the amount, resulting in a clean display of the probability.

rejas: I mixed something up, didnt realize that `showPrecipitationAmount` was the old config option. 
So my reasoning now is: 
since precipation isnt really use dmuch (otherwise more users would have complained :-) I 'd just add the `showPrecipitationProbability` config option and let the user decide if they want to show one, both or none (default). no deprecation warning necessary then and the code is a little smaller.

rejas (changes requested):

modules/default/weather/current.njk
MagMar94: Agreed :blush:

modules/default/weather/providers/weatherflow.js
```diff
@@ -55,6 +55,7 @@ WeatherProvider.register("weatherflow", {
 					weather.date = moment.unix(forecast.day_start_local);
 					weather.minTemperature = forecast.air_temp_low;
 					weather.maxTemperature = forecast.air_temp_high;
+					forecast.precipitationProbability = forecast.precip_probability;
```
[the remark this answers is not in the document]
MagMar94: Ah, the old copy paste...

rejas (approved):

rejas:
Approved! Thanks a thousand times for your work and patience! 

Lets get this out into the develop-wilderness and see if we get any bug reports :-)

MagMar94:
Haha, thank you so much for the excellent reviews! FYI, I opened https://github.com/MichMich/MagicMirror-Documentation/pull/149 to handle the documentation.

PR #3028, opened 2023-02-06
Create Thai Language file

tomzt:
Create Thai Language file. 
Translation from English to Thai.

Hello and thank you for wanting to contribute to the MagicMirror² project

**Please make sure that you have followed these 4 rules before submitting your Pull Request:**

> 1. Base your pull requests against the `develop` branch.
>
> 2. Include these infos in the description:
>
> - Does the pull request solve a **related** issue?
> - If so, can you reference the issue like this `Fixes #<issue_number>`?
> - What does the pull request accomplish? Use a list if needed.
> - If it includes major visual changes please add screenshots.
>
> 3. Please run `npm run lint:prettier` before submitting so that
>    style issues are fixed.
>
> 4. Don't forget to add an entry about your changes to
>    the CHANGELOG.md file.

**Note**: Sometimes the development moves very fast. It is highly
recommended that you update your branch of `develop` before creating a
pull request to send us your changes. This makes everyone's lives
easier (including yours) and helps us out on the development team.

Thanks again and have a nice day!

PR #3031, opened 2023-02-07
Enforce PRs to be based against develop branch

rejas:
"Inspired" by my mistake in https://github.com/MichMich/MagicMirror/pull/3028 this PR will add a worfklow check for the branch a PR is based against.

Open question is if this prevents @MichMich from preparing a release?


khassel:
> Open question is if this prevents @MichMich from preparing a release?

looking into how he does releases, yes, he uses a PR.

May we could use something like "skipLabels" as provided in the changelog action - or look for a special string in the commit message (which would be not so safe)

khassel:
may we could use e.g. https://github.com/marketplace/actions/pr-labels to get the PR labels and not exit if a specific label exists.

We need this not only for releases but also for some edge cases as updating README (which is possible after next release without update notifications).

rejas:
> may we could use e.g. https://github.com/marketplace/actions/pr-labels to get the PR labels and not exit if a specific label exists.

I assume we maintainers are the only ones who can add labels to issues @MichMich ? Then adding a label like you say (maybe "hotfix") would be enough I think to prevent future master merges and allow stuff to be merged for such cases

sdetweil:
I think there should never be a direct fix to master.  always promote from some other branch. 
pr to hotfix-x to master, 
pr to develop to master

rejas:
So this PR is ready I think. It will complain if the PR is based upon master and does not have the new "mastermerge" label. Maintainers can test it on this PR, just add / remove the label and trigger the tests: https://github.com/MichMich/MagicMirror/pull/3036

Only thing that doesnt get caught is IF the PR is developed against Develop, but then submitted against Master (thats why this PR here wont fail if I change the base to master). Any ideas on how to fix that or is that a usecase we can ignore?

khassel:
> Any ideas on how to fix that or is that a usecase we can ignore?

I think we can ignore it. If someone has an idea for a fix we can add it later ...

khassel (approved):

PR #3032, opened 2023-02-08
introduce envsubst for config.js, update deps

khassel:
This is the implemenation of envsubst discussed in #1756 

Documentation update will follow after merge.


js/app.js
```diff
@@ -56,13 +57,62 @@ function App() {
 	 *
 	 * @param {Function} callback Function to be called after loading the config
 	 */
-	function loadConfig(callback) {
+	async function loadConfig(callback) {
```
rejas, line 60: I would like to see the loadConfig being asynced and also being freed from the callback. Have you tried that?

rejas (changes requested):
Can you add an example(maybe just with a comment in the sample.config so that people know how to use it?
Would it be possible to add a test for this?
And of course the async question if we convert the whole loadCOnfig mthod to async without callback?

.eslintignore
```diff
@@ -1 +1,2 @@
 modules/default/calendar/vendor/*
+tests/configs/port_variable.js
```
rejas, line 2: why are these two ignores necessary? locally nothing complains about them....

khassel: you are right, will remove them. Had first `port_variable.js` (forgot `.template` at the end) and with `.js` the tests failed.

rejas (approved):

PR #3038, opened 2023-02-16
Update coverage and jest settings

rejas:


rejas:
> I'm not happy with the jest changes because when I now debug one test I have to add `--verbose` to it.

Thats why I added the verbose flag to the specialized test scripts (test:e2e etc) so that one can use them and get the verbose stuff (they are not used anywhere else as far as I can see=. Would that be okay?

khassel:
As already mentioned before please don't change `verbose: true,` but instead use `"test:coverage": "NODE_ENV=test jest --coverage -i --forceExit --verbose false",` for the one test where you don't want to see the verbose stuff.

In all other tests we want to have the verbose stuff and also when testing locally, e.g. running one test with `npx jest tests/unit/modules/default/utils_spec.js` I don't want to add the `--verbose` flag.

rejas:
Undid the changes :-)

khassel (approved):

khassel:
Thanks!

PR #3042, opened 2023-02-18
stylelint

KristjanESPERANTO:
1. Update `stylelint` dependencies
  - As of stylelint v15, we do not need `stylelint-config-prettier` anymore: https://github.com/prettier/stylelint-config-prettier/releases/tag/v9.0.5
2. Switch to `stylelint-config-standard`: `stylelint-prettier/recommended` has not been updated for a long time and still needs the old `stylelint-config-prettier` 
3. Handle new `stylelint` issues


modules/default/alert/styles/notificationFx.css
```diff
@@ -1,7 +1,9 @@
+/* stylelint-disable number-max-precision */
```
rejas, line 1: can we get rid of that ignore?

modules/default/clock/clock_styles.css
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable selector-class-pattern */
```
rejas, line 1: can we get rid of that ignore?

rejas (changes requested):

modules/default/alert/styles/notificationFx.css
KristjanESPERANTO: Gladly. I see two approaches:

1. Reduce the values by one digit in the animations. I don't think it has any visible impact.
2. Increase the value for number-max-precision by one in the .stylelintrc.json.

What do you think?

modules/default/clock/clock_styles.css
KristjanESPERANTO: Done :slightly_smiling_face:

modules/default/alert/styles/notificationFx.css
rejas: I would go for 1) since 
- like you said it wouldnt probably impact the looks
- we would stay closer to the mainline stylelint rules

KristjanESPERANTO: I also think this is the better choice. I re-enabled the rule and performed the rounding

rejas (approved):

rejas (approved):

PR #3050, opened 2023-02-22
Convert load callbacks to async/await

rejas:
Another one of these PRs....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Update dates in Calendar widgets every minute
 - Cleanup jest coverage for patches
 - Update `stylelint` dependencies, switch to `stylelint-config-standard` and handle `stylelint` issues
+- Convert load callbacks to async/await
 - Convert module start to async/await
 - Convert translator callbacks to async/await
 - Convert app-start/-stop callbacks to async/awaits
--- a/js/loader.js
+++ b/js/loader.js
@@ -190,13 +190,13 @@ const Loader = (function () {
 	 * @param {string} fileName Path of the file we want to load.
 	 * @returns {Promise} resolved when the file is loaded
 	 */
-	const loadFile = function (fileName) {
+	const loadFile = async function (fileName) {
 		const extension = fileName.slice((Math.max(0, fileName.lastIndexOf(".")) || Infinity) + 1);
 		let script, stylesheet;
 
 		switch (extension.toLowerCase()) {
 			case "js":
-				return new Promise((resolve, reject) => {
+				return new Promise((resolve) => {
 					Log.log("Load script: " + fileName);
 					script = document.createElement("script");
 					script.type = "text/javascript";
@@ -211,7 +211,7 @@ const Loader = (function () {
 					document.getElementsByTagName("body")[0].appendChild(script);
 				});
 			case "css":
-				return new Promise((resolve, reject) => {
+				return new Promise((resolve) => {
 					Log.log("Load stylesheet: " + fileName);
 
 					stylesheet = document.createElement("link");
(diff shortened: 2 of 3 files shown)


rejas:
Merge conflict fixed

khassel:
I think the tests are not stable anymore. I know that the tests on github are very sensitive and that fails there are normal but I'm running own tests on gitlab after building new develop images and after this PR its the first time since months that this test fails.

I repeated the test run 3 times (all failed) and you can find the results here:
- https://gitlab.com/khassel/magicmirror/-/jobs/3818204533
- https://gitlab.com/khassel/magicmirror/-/jobs/3818339454
- https://gitlab.com/khassel/magicmirror/-/jobs/3818480669

Running locally I had similar errors but also successful runs, so it seems a transient problem.

rejas:
Not good :-( And this happened after this PR? If so, we should undo that for now...

khassel:
not sure but I ran also a pipeline before this PR which was successful

rejas:
Can you check if my cleanup branch / PR improves the situation?

khassel:
I tested now locally against your branch `misc/cleanup` and got 5 successful runs, no error.

Running this in ci is to complicated, we will see after merge ...

rejas:
The https://github.com/MichMich/MagicMirror/pull/3051 looks a little more stable indeed on my pc. Shall we give it a try?

khassel:
merged but not better, see https://gitlab.com/khassel/magicmirror/-/jobs/3836164426

We should be able to revert this before next release if there is no solution ...

khassel:
there are simliar error in the node18 tests: https://github.com/MichMich/MagicMirror/actions/runs/4276168598/jobs/7444059604#step:4:1625

Another word to my setup, on my develop images (where above tests are running) I'm on node v19.

khassel:
this becomes strange ... 

I changed my pipeline to do the tests with node v16/18/19, you can see th results [here](https://gitlab.com/khassel/magicmirror/-/pipelines/790578549).

The v16 pipeline succeeded, but the newer node versions fails.

rejas:
The first error I see is something new for me:
"
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From tests/e2e/modules/calendar_spec.js.
"

Any idea on this? I am on vacation this week so I am a little slow...

rejas:
So, what shall we do now with this?
I am probably a little busy this month and the next few too due to family expansion (delivery is expected around easter).
Shall we:
- keep it as is
- revert async commits

Opinions @sdetweil @khassel @MichMich ?

MichMich:
As long as the tests don't break I'm all for it.

MichMich:
More importantly: congrats with the planned family expansion! 😁

rejas:
> As long as the tests don't break I'm all for it.

well, they dont break a 100% of the time, the tests are just sometimes breaking. so we have to restart the checks untik they run through.

khassel:
> and the next few too due to family expansion (delivery is expected around easter).

cool 👍

I have very limited time the whole week before next release ...

I'm unhappy with the current tests. I think we have 2 problems:
1. after the async changes the tests are failing for node versions >= 18 locally or in my gitlab tests (not on github because the jobs there are very slow)
2. some electron tests are failing only on github (restart mostly works)

AFAIS are both timing problems.

I try to find some time to test Nr. 1 by reverting the async commits, for Nr. 2 we can try to increase jest timeout from 20 to 30 sec.

The current situation must be solved, no one is looking deeper into tests which are failing with every run.

khassel:
> after the async changes the tests are failing for node versions >= 18 locally or in my gitlab tests (not on github because the jobs there are very slow)

this happens also on github, see https://github.com/MichMich/MagicMirror/actions/runs/4276168598/jobs/7444059604

Looking at all the fails this happens only in the 2 weather tests `weather_forecast_spec.js` and `weather_hourly_spec.js` in the `for (const [index, xxx]` construction.

The weather tests have a own helper functions and a special construction to use mock data, may some of these things are causing these errors.

PR #3060, opened 2023-03-07
Undo async patches

rejas:
As discussed in #3050 this PR undoes all async patches until we get the timing issues (or whatever causes the tests to fail) sorted out.


KristjanESPERANTO:
It's a shame about the work you put in. Maybe it would be a good idea after the new release to slowly convert one part after the other to async? Maybe not all cause the problem and we can find out which parts are critical.

khassel:
yes, but its not reverted yet and if we have to its not lost, we have to find the real issue which seems to be difficult, see also my comment [here](https://github.com/MichMich/MagicMirror/pull/3050#issuecomment-1462581181), if you have an idea let us know ...

KristjanESPERANTO:
@khassel Is that (#3061) any help?

PR #3061, opened 2023-03-11
Try to fix the test problems after the async changes

KristjanESPERANTO:
To fix the test problems after the async changes (#3050), I tried to find out the cause with help of the Jest option `--detectOpenHandles` on the e2e tests. I got some error messages. And since I haven't had much experience with Jest yet, I threw some async and await keywords around until the error messages were gone. Probably not all changes make sense, but maybe this approach will help us.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,10 +22,10 @@
 		}
 	},
 	"rules": {
-		"prettier/prettier": "error",
 		"eqeqeq": "error",
 		"no-prototype-builtins": "off",
 		"no-unused-vars": "off",
-		"no-useless-return": "error"
+		"no-useless-return": "error",
+		"prefer-template": "error"
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Fix wrong vertical alignment of calendar title column when wrapEvents is true (#3053)
 - Fix empty news feed stopping the reload forever
 - Fix e2e tests (failed after async changes) by running calendar and newsfeed tests last
+- Lint: Use template literals instead of string concatenation
 - Fix default alert module to render HTML for title and message
 
 ## [2.22.0] - 2023-01-01
--- a/js/app.js
+++ b/js/app.js
@@ -18,7 +18,7 @@ const envsub = require("envsub");
 
 // Get version number.
 global.version = require(`${__dirname}/../package.json`).version;
-Log.log("Starting MagicMirror: v" + global.version);
+Log.log(`Starting MagicMirror: v${global.version}`);
 
 // global absolute root path
 global.root_path = path.resolve(`${__dirname}/../`);
@@ -64,7 +64,7 @@ function App() {
 		// For this check proposed to TestSuite
 		// https://forum.magicmirror.builders/topic/1456/test-suite-for-magicmirror/8
 		const configFilename = path.resolve(global.configuration_file || `${global.root_path}/config/config.js`);
-		let templateFile = configFilename + ".template";
+		let templateFile = `${configFilename}.template`;
 
 		// check if templateFile exists
 		try {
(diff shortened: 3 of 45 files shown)


rejas:
Using your PR I still get these messages:
"Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?"

Also, the tests still fail sometimes locally for me :-( Any luck on your side maybe?

KristjanESPERANTO:
Oh, I just looked at the e2e test and ignored the others. I'll take a look at the others.

KristjanESPERANTO:
> "Force exiting Jest: Have you considered using --detectOpenHandles to detect async operations that kept running after all tests finished?"

Is that an indication of the error? I get the same message in the master branch.

At the beginning I also had locally an error message three times during the electron tests. After the last commit this didn't happen anymore. Since the error only occurs very rarely, I'm not sure if it really makes a difference.

khassel:
> At the beginning I also had locally an error message three times during the electron tests. After the last commit this didn't happen anymore. Since the error only occurs very rarely, I'm not sure if it really makes a difference.

this is hopefully a fix for the electron tests :)

> Is that an indication of the error? I get the same message in the master branch.

no, this is no indication of the error and this is printed after every test run (we could add --detectOpenHandles which would suppress this message)

We have to use `--forceExit` in the e2e tests, may this is also an indication that something isn't correct. Running everey e2e test as single test I got:

```bash
tests hanging, needing --forceExit (all these tests are using `waitForElement`):
- tests/e2e/modules_empty_spec.js
- tests/e2e/env_spec.js
- tests/e2e/modules/weather_current_spec.js
- tests/e2e/modules/weather_forecast_spec.js
- tests/e2e/modules/newsfeed_spec.js
- tests/e2e/modules/clock_spec.js
- tests/e2e/modules/calendar_spec.js
- tests/e2e/modules/compliments_spec.js
- tests/e2e/modules/alert_spec.js
- tests/e2e/modules/weather_hourly_spec.js
- tests/e2e/modules/clock_es_spec.js
tests ending correctly not using `waitForElement`:
+ tests/e2e/template_spec.js
+ tests/e2e/port_spec.js
+ tests/e2e/ipWhitelist_spec.js
+ tests/e2e/fonts_spec.js
+ tests/e2e/vendor_spec.js
+ tests/e2e/translations_spec.js
tests ending correctly using `waitForElement`:
+ tests/e2e/modules_position_spec.js
+ tests/e2e/modules_display_spec.js
+ tests/e2e/modules/helloworld_spec.js
```

The e2e tests are still failing as before with this PR :(

Because of these errors

```bash
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From tests/e2e/modules/calendar_spec.js.

      22 |      //      return nodefetch(url, options);
      23 |      // }
    > 24 |      const nodefetch = await require("node-fetch");
         |                              ^
      25 |      return nodefetch(url, options);
      26 | }
      27 |

      at require (js/fetch.js:24:26)
      at fetchCalendar (modules/default/calendar/calendarfetcher.js:63:17)
      at Timeout.<anonymous> (modules/default/calendar/calendarfetcher.js:98:7)

ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From tests/e2e/modules/newsfeed_spec.js.

      22 |      //      return nodefetch(url, options);
      23 |      // }
    > 24 |      const nodefetch = await require("node-fetch");
         |                              ^
      25 |      return nodefetch(url, options);
      26 | }
      27 |

      at require (js/fetch.js:24:26)
      at fetchNews (modules/default/newsfeed/newsfeedfetcher.js:92:5)
      at Timeout.<anonymous> (modules/default/newsfeed/newsfeedfetcher.js:114:7)
```

I deleted the calendar and newsfeed tests and without them all errors are gone, also the weather tests are working now (so I think the weather tests were affected because they ran after the above tests).

So I will now dig into the calendar and newsfeed tests ...

KristjanESPERANTO:
> we could add --detectOpenHandles which would suppress this message

According to the [jest documentation](https://jestjs.io/docs/cli#--detectopenhandles), this would have performance disadvantages. But my tests with this option didn't seem that long to me.

```sh
    > 24 |      const nodefetch = await require("node-fetch");
         |                              ^
```

Is this is a sign that we should finally get rid of node-fetch 2? \*justjoking\* 😅

I've just tried a few more things, but haven't really made any progress. By the way, I'm impressed with all the tests you've created. Respect! 👏

rejas:
So what about this PR @khassel , will this be obsolete soon too?

khassel:
> So what about this PR @khassel , will this be obsolete soon too?

yes, but may we need the fix for the electron test ...

PR #3063, opened 2023-03-11
fix e2e tests (failed after async changes)

khassel:
by running calendar and newsfeed tests last.

Additional change: allow unit tests to run parallel

This is no fix of the real issue of calendar and newsfeed tests but I moved them to the end of the tests so other tests are not failing anymore.

There are coming follow up PR's for the real fixes (when I find the culprits).

With these change we can stay with the async changes done by @rejas and https://github.com/MichMich/MagicMirror/pull/3060 is obsolete.


KristjanESPERANTO (approved):
Nice! Looks good to me.

rejas (approved):

PR #3066, opened 2023-03-14
Use template literals instead of string concatenation

KristjanESPERANTO:
We have used it inconsistently till now. Template literals are more modern and easier to maintain in my opinion.

Because that's a large amount of changes, here's a way to reproduce it: I added the rule `"prefer-template": "error"` to the `.eslintrc.json` and did an autofix. Since this caused a new problem in line 409 of `newsfeed.js`, I reversed it in that line and also removed the rule from the eslint config file.

The rule is described here: https://eslint.org/docs/latest/rules/prefer-template

Note: I've played around with some other linter rules as well, and some seem to point to some specific, non-cosmetic, issues. But before I dive even deeper and then introduce even bigger and hardly understandable changes at once, I thought I'd start with this simple cosmetic rule.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -26,6 +26,7 @@
 		"eqeqeq": "error",
 		"no-prototype-builtins": "off",
 		"no-unused-vars": "off",
-		"no-useless-return": "error"
+		"no-useless-return": "error",
+		"prefer-template": "error"
 	}
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,6 @@
 		}
 	},
 	"rules": {
-		"prettier/prettier": "error",
 		"eqeqeq": "error",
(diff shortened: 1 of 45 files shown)


khassel (approved):

rejas:
Good PR. I would vote for adding the rule to the eslint config and add one "// disable" exception to the line in the newsfeed module. That way we stay "modern" and can fix the line maybe later :-)

rejas (changes requested):
add rule to eslintrc and add exception to the one line in newsfeed module

KristjanESPERANTO:
I just did the following:
- fixed the problematic line
- added the rule (`"prefer-template": "error"`)
- removed a redundant rule (`"prettier/prettier": "error"`) - This change has nothing to do with the other changes, but is too minor for an extra PR.

PR #3067, opened 2023-03-19
update dependencies and main.css to match new stylelint rules

khassel:
On `develop` my css test failed because I was already on newest dependencies.

- update deps
- update `main.css` because of stylelint issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Changed updatenotification module for MagicMirror repo only: Send only notifications for `master` if there is a tag on a newer commit
 - Update dates in Calendar widgets every minute
 - Cleanup jest coverage for patches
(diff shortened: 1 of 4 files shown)


KristjanESPERANTO (approved):
Interesting, the css test works for me without an adjustment of the css file (with new dependencies). So I don't see any new rule.

But switching to inset is fine anyway, I think.

khassel:
strange, got this:

```bash
> stylelint 'css/main.css' 'fonts/*.css' 'modules/default/**/*.css' 'vendor/*.css' --config .stylelintrc.json
css/main.css
 177:3  ✖  Expected shorthand property "inset"  declaration-block-no-redundant-longhand-properties
1 problem (1 error, 0 warnings)
```

KristjanESPERANTO:
I should think once more before I write. I must have forgotten `npm i` 🤦. Now I get the message too.

PR #3069, opened 2023-03-20
Fixed clock module sunrise/sunset timezone bug

JakeBinney:
PR to fix sunrise/sunset timezone rendering bug. This PR converts existing system timezone to user-requested local timezone

Hello and thank you for wanting to contribute to the MagicMirror² project

**Please make sure that you have followed these 4 rules before submitting your Pull Request:**

> 1. Base your pull requests against the `develop` branch.
>
> 2. Include these infos in the description:
>
> - Does the pull request solve a **related** issue?
> - If so, can you reference the issue like this `Fixes #<issue_number>`?
> - What does the pull request accomplish? Use a list if needed.
> - If it includes major visual changes please add screenshots.
>
> 3. Please run `npm run lint:prettier` before submitting so that
>    style issues are fixed.
>
> 4. Don't forget to add an entry about your changes to
>    the CHANGELOG.md file.

**Note**: Sometimes the development moves very fast. It is highly
recommended that you update your branch of `develop` before creating a
pull request to send us your changes. This makes everyone's lives
easier (including yours) and helps us out on the development team.

Thanks again and have a nice day!

PR #3070, opened 2023-03-20
Fixed clock module sunrise/sunset timezone bug

JakeBinney:
Updated sunrise/sunset times to display user-requested timezone rather than system timezone.

Note: rebase of  #3069 against develop rather than master

Hello and thank you for wanting to contribute to the MagicMirror² project

**Please make sure that you have followed these 4 rules before submitting your Pull Request:**

>
>
>
>    style issues are fixed.
>
> 4. Don't forget to add an entry about your changes to
>    the CHANGELOG.md file.


modules/default/clock/clock.js
```diff
@@ -196,6 +196,12 @@ Module.register("clock", {
 			}
 			const untilNextEvent = moment.duration(moment(nextEvent).diff(now));
 			const untilNextEventString = `${untilNextEvent.hours()}h ${untilNextEvent.minutes()}m`;
+
+			// Convert sunrise/sunset time to local representation of requested timezone
+			if (this.config.timezone) {
+				sunTimes.sunrise = sunTimes.sunrise.toLocaleString("en-us", { timeZone: this.config.timezone });
```
rejas, line 202: probably needs to respect the "locale" config value instead of the hardcoded "en-us"

rejas (changes requested):
small change necessary. but thanks for the PR, do you feel experienced enough for writing a test about this too?

rejas (approved):
so, any luck with a test?

rejas (changes requested):
sorry, tested it and doesnt work.
Locale has to be done somthign like 

const locale = config.locale??'en-US';
	
but there seem to be other issues with toLocaleString:

Adding locale: "de-DE" to my config leads then to an "invalid date" on the ui...

JakeBinney:
I'll make that change to fall back to en-US but also start to troubleshoot other locales to figure out what's going on there.

As for a test, is there an existing test module i should add to or is the intention for me to write a new test for the clock module?

JakeBinney:
Confirming that locale `de-DE` produces same "invalid date" error on my end. 

The error is not necessarily within `toLocaleString`, but rather traces back to the existing `formatTime` function as the moment call in that function is expecting a datetime in the `en-US` locale. The `de-DE` locale gives a date format delineator of `.` whereas moment is expecting `:` delineator. 

Will need to research better internationalization support for this function

JakeBinney:
Found a fix for localization of time. Don't have time to implement it tonight but it is possible to call `toLocaleString` with options such as:

`date.toLocaleString(this.config.locale, {hour: '2-digit', minute: '2-digit', hour12: false}` to honor locale.

rejas:
> Found a fix for localization of time. Don't have time to implement it tonight but it is possible to call `toLocaleString` with options such as:
> 
> `date.toLocaleString(this.config.locale, {hour: '2-digit', minute: '2-digit', hour12: false}` to honor locale.

Do we need the locale at all? For formatting the time we already have showPeriod / showPeriodUpper in the module, so what else would the locale help here in terms of foramtting time?

Take alook at my PR here: https://github.com/MichMich/MagicMirror/pull/3073

Maybe you could just use that new util method for formatting the sunrise/sunset time (once it is merged of course)?

JakeBinney:
Using locale was initially intended to be a lightweight fix just for converting timezones, but using moment.tz in the common util works just as well and introduces less disruption to the existing code base. If you're already refactoring formatTime then I like your solution for converting to timezone in that function. I'll update this PR considering that solution. Thanks!

rejas:
Hi @JakeBinney with #3073 merged, can you check if the latest develop branch fixes your bug too?

JakeBinney:
Bug is still present in latest develop branch. Sunrise/sunset times are still showing in system time rather than converted to local time of the timezone. Can anyone else confirm that I'm not crazy?

JakeBinney:
Updated utils.js. Was converting current time instead of converting time passed by SunCalc

rejas:
Good catch :-) Adjusted my faulty tests so that they would fail if your PR wasnt there.

khassel:
@rejas looking in the changed files of this PR is a mess, there are 18 files changed but in the resulting merge commit only 2 changed.

So in the future I would appreciate to rebase (or merge latest `develop`) before merging such PR's ...

rejas:
I agree @khassel. Weird why github does this changed-files mess though...

JakeBinney:
I was wondering about that yesterday. Was that something i did? I'm still trying to get better at git hygiene.

khassel:
> Was that something i did? I'm still trying to get better at git hygiene.

not actively ...

your PR was very old and I think you never get the changes from `develop` since you created your branch. With this when your branch is compared to `develop` it will show your changes **and** all changes done on `develop`.

To avoid this you should rebase your branch on current `develop` (or merge current `develop` in your branch).

My point was not to merge such a PR without rebase/merge to current `develop`. We should ask the author to do this or do it ourself if the author is not experienced enough with git.

This all has no impact to the new `develop` branch after this PR was merged (would be identical if rebase/merge was done before). But if someone looks into this PR and wants to know what has changed, it would be better to see only the 2 files affected instead of 18. So this all is only to get better diffs for reviewing such a PR and for Github PR history.

JakeBinney:
That all makes plenty of sense. I appreciate you taking the time to write all that out!

PR #3071, opened 2023-03-20
Revise require imports

KristjanESPERANTO:
- order (external first)
- remove superfluous file extensions
- new line after imports
- deconstruct (only one time (in `check_config.js`))
- fix path (only one time (in `global-setup.js`))

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
 	"extends": ["eslint:recommended", "plugin:prettier/recommended", "plugin:jsdoc/recommended"],
-	"plugins": ["prettier", "jsdoc", "jest"],
+	"plugins": ["prettier", "import", "jsdoc", "jest"],
 	"env": {
 		"browser": true,
 		"es2022": true,
@@ -23,6 +23,7 @@
 	},
 	"rules": {
 		"eqeqeq": "error",
+		"import/order": "error",
 		"no-prototype-builtins": "off",
 		"no-throw-literal": "error",
 		"no-unused-vars": "off",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Update `stylelint` dependencies, switch to `stylelint-config-standard` and handle `stylelint` issues, update `main.css` matching new rules
 - Update Eslint config, add new rule and handle issue
 - Convert lots of callbacks to async/await
-- Fixed Open-Meteo wind speed units
+- Revise require imports
 
 ### Fixed
 
@@ -58,6 +58,7 @@ _This release is scheduled to be released on 2023-04-01._
 - Fix e2e tests (failed after async changes) by running calendar and newsfeed tests last
 - Lint: Use template literals instead of string concatenation
 - Fix default alert module to render HTML for title and message
+- Fixed Open-Meteo wind speed units
 
 ## [2.22.0] - 2023-01-01
 
--- a/js/app.js
+++ b/js/app.js
@@ -8,13 +8,13 @@
 // Alias modules mentioned in package.js under _moduleAliases.
 require("module-alias/register");
 
+const envsub = require("envsub");
 const fs = require("fs");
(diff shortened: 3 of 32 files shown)


khassel (approved):

KristjanESPERANTO:
@rejas Yes, I have worked with this plugin and the rule. But I'm not ready to implement them yet. There are still some problems with the `_moduleAliases` and I want to find a way to get along without exceptions.

Since I don't know yet when I could integrate the plugin, I wanted to bring in these improvements already.

Or would it make more sense to set the PR to draft for so long?

@sdetweil Things like unified code and modern coding conventions increase code quality and make projects easier to maintain. Functional changes are easier to integrate. Sometimes you also discover errors that you wouldn't have seen otherwise. 

Also, I personally learn better how the code works and can help more when problems occur.

If you want to read deeper, the topics [Code review](https://en.wikipedia.org/wiki/Code_review) and [Code refactoring](https://en.wikipedia.org/wiki/Code_refactoring) are good starting points.

sdetweil:
I understand all the principles, and the objectives, just not the process.

yes, when u change a file for a functional reason, then apply these cosmetic changes. don't change the files JUST for cosmetic purposes.   

if it's not broke, you don't apply mass changes for policy purposes. 

been doing this a long time. you don't change coding practices to match your developers. 

never once has the alphabetical order of imports been functionally useful, and some times the order matters

KristjanESPERANTO:
> don't change the files JUST for cosmetic purposes.

Some changes are cosmetic but the purpose is not cosmetic. The purpose is cleaner code which is easier to maintain and has fewer bugs.

> if it's not broke, you don't apply mass changes for policy purposes.

The sense of refactoring is exactly the opposite of that statement.

_Refactoring is a systematic process of improving code without creating new functionality that can transform a mess into clean code and simple design._ (https://refactoring.guru/refactoring)

Note: I don't want to say that MM is a mess code :smile: 

> never once has the alphabetical order of imports been functionally useful

This PR is not just about alphabetical order.

With this PR, it may not be entirely clear where the whole process is going, I'll admit that.

Are there no review or refactoring processes in your professional projects? In my understanding, this is part of best practice.

sdetweil:
review and refactoring are engaged when the error rate or new function demands it. not just because it's a good long term policy.

you'll do whatever. I disagree.  fought this battle over and over and it never turns out for the best.

KristjanESPERANTO:
Apparently we have had very contrasting experiences.

> you'll do whatever.

How do you mean that? I submitted a PR and you are free to reject it. No hard feelings :slightly_smiling_face:

rejas:
You both argue well, and I can see both sides of the argument. Still, I favor merging this PR. So, whats the third opinion here @khassel ?

rejas (approved):

khassel:
already approved this

KristjanESPERANTO:
I am about to implement the linter rule for this. Please wait a few more minutes.

KristjanESPERANTO:
Okay, the tests are running successfully locally and here at GitHub. I'm done with the PR so far :slightly_smiling_face: 

In the future I will try to include the linter rule right away.

khassel:
the change here in `serveronly.index.js` breaks running mm in serveronly mode, I think this must reverted ...

KristjanESPERANTO:
> the change here in `serveronly/index.js` breaks running mm in serveronly mode, I think this must reverted ...

I have an idea to solve this without making an exception to the new rule here. I will make a PR this tomorrow.

How did you find the issue?

khassel:
I have a serveronly setup running as website and I built a new docker image with this changes which is substituted automatically via watchtower so my mm was down (container restarting).

KristjanESPERANTO:
> I have a serveronly setup running as website and I built a new docker image with this changes which is substituted automatically via watchtower so my mm was down (container restarting).

Ah I see. `npm run server` has thrown an error.

> other point: Do we really have no test for serveronly?

I can't see a test for serveronly, but it  would be important. I can't create one at the moment. Do we open an issue to address later? Or do you have the time?

PR #3073, opened 2023-03-25
Refactor formatTime into util class

rejas:
While looking at https://github.com/MichMich/MagicMirror/pull/3070 I noticed that the weather and clock module do some formatTime stuff, so why not use a common function for that?


KristjanESPERANTO (approved):
Nice! I haven't tested it, but it looks good to me.

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ Special thanks to @khassel, @rejas and @sdetweil for taking over most (if not al
 - Update Eslint config, add new rule and handle issue
 - Convert lots of callbacks to async/await
 - Revise require imports (#3071 and #3072)
+- Refactor formatTime into common util function for default modules
```
KristjanESPERANTO, line 65: That's for 2.23.

rejas: Good catch :-)

khassel (approved):

khassel:
> Running tests with times/dates is a PITA ... Had to add an action to the github-actions for setting the timezone, otherwise the tests would fail here (while being fine locally)...

hopefully this helps ...

khassel:
now I understood that only the new tests need the timezone (because these tests are now failing on my side). Is there no way to get these test running without needing a specific timezone? Or move them to the `electron` section where we can handle such things already?

khassel:
found a jest solution, will provide a new PR ...

PR #3085, opened 2023-04-07
Drop node v14 support

KristjanESPERANTO:
node v14 will have reached end-of-life by the next release.

From April 18th we can use node v20 instead of v19 for testing.


KristjanESPERANTO:
Interesting that the tests work with v16 and 19, but not with v18. The v18 test works locally for me.

khassel:
> Interesting that the tests work with v16 and 19, but not with v18. The v18 test works locally for me.

tests are still (or again) not stable ... unrelated to your PR ... (started the v18 test again and it passed)

khassel (approved):

rejas (approved):

PR #3088, opened 2023-04-08
Refactor calendar methods into util class

rejas:
Refactored some methods in calendar module:
- move methods into own file 
- dont call shorten method from titelTransform because why? just call them after each other.
- added tests for util methods
- cleaned up other tests


khassel (approved):

khassel (approved):

rejas:
would then be ready to merge @khassel  :-)

PR #3093, opened 2023-04-21
use node v20 in github workflow (replacing v19)

khassel:
additional:

- update deps
- suppress more unwanted log errors in e2e tests
- add .gitattributes
- fix prettier/js warnings


tests/e2e/helpers/mock-console.js
```diff
@@ -1,10 +1,17 @@
 /**
  * Suppresses errors concerning web server already shut down.
- *
```
rejas, line 3: Lots of empty line changes. Probably through the eslint-jsdoc-plugin update? If its a new default, maybe do that in a seperate PR to keep the git history cleaner?

khassel: > Probably through the eslint-jsdoc-plugin update?

think so ...

will revert the second commit here and make second PR.

rejas (approved):

PR #3099, opened 2023-05-07
Add openmeteo precipitation probability

OWL4C:
Due to the size of the commit there is no need for a long explanation:
Openmeteo supports precipitation probability in api, but the weather provider .js had no support for it. After adding 4 lines it works as expected. This should have no consequences on other files but improves usability for (imo) the best weather provider.


rejas:
Thx for the PR, could you add a CHANGELOG entry too?

OWL4C:
Like this? I have no former git/pr experience...

rejas:
> Like this? I have no former git/pr experience...

Like this indeed. Now only thing failing the checks is the linter. Please run 'npm run lint:prettier' and commit the changes. Thx in advance!

PR #3104, opened 2023-05-14
Use single config backup file when config.template is used

oscarb:
Copy `config.js` to a single `config.js-old` file whenever a `config.template` is present, instead of using `Date()` as a suffix. Creating files with a timestamp suffix means that whenever Magic Mirror is restarted or recreated a new file is written into the config directory.

Benefits:
* Single backup file will avoid excessive writing of files
* Saves space and usage on SD cards.
* Makes the folder cleaner and easier to overview, compared to ending up with something like this as time goes on:
image


rejas:
Thx for the PR, I am fine with it.
What dou you think @khassel as the OC (original coder)?

PR #3108, opened 2023-05-21
Adding experimental UV Index support for weather module

OWL4C:
This pr adds (config.js toggleable) showUV_Index to the hourly and current weather modules, with right now only openmeteo configured to supply the data. Other providers could have support too by adding `uv_index` to current and hourly.

For example the current weather looks like image positioned after sunset in the top row.

The following "hacks" are included and could be fixed to make it cleaner, but the functionality is wanted and it works without problem.
- To hide entries where the UV Index is 0 i added an if statement to the `hourly.njk` which is not how precipitation is handled.
The following are minor things that might not need fixing:
- The forecast option does not have UV support. This might not be relevant since UV changes throughout the day but i tried to implement a "max_UV" to openmeteo.js, but am not confident enough in JS to accomplish that.
-  The UV Icon is wi-hot and manually added to the `.njk`'s. This could be made changeable by a config but does not seem relevant since wi-hot is not used by anything else as far as i can tell.


OWL4C:
The last batch of tests failed because "uv_index" is not Kebap Case, which is unfixable since the API provides it only as such.

sdetweil:
you can use 
/* eslint-disable */
.....
/* eslint-enable */

around the troublesome block

rejas:
> The last batch of tests failed because "uv_index" is not Kebap Case, which is unfixable since the API provides it only as such.

The error comes from modules/default/weather/weather.css where you define a class with that name (and reference in the njk template). So those places you can fix. The API param name doesnt trigger the error.

rejas:
But thanks for the PR. COuld you provide a screenshot of the whole weather look now?

OWL4C:
I am pretty sure i fixed everything and made it look nice
weather
. 
The icon is still set manually and the "ugly" UV == 0 still there but the api already supports daily UV so i added that too

modules/default/weather/hourly.njk
```diff
@@ -7,9 +7,17 @@
             <tr {% if config.colored %}class="colored"{% endif %} {% if config.fade %}style="opacity: {{ currentStep | opacity(numSteps) }};"{% endif %}>
                 <td class="day">{{ hour.date | formatTime }}</td>
                 <td class="bright weather-icon"><span class="wi weathericon wi-{{ hour.weatherType }}"></span></td>
-                <td class="align-right bright">
+                <td class="align-right bright temperature">
```
rejas, line 10: is that new class there for a purpose?

modules/default/weather/weather.js
```diff
@@ -27,6 +27,7 @@ Module.register("weather", {
 		showPeriodUpper: false,
 		showPrecipitationAmount: false,
 		showPrecipitationProbability: false,
+		showUV_Index: false,
```
rejas, line 30: Could you rename that config option to showUVIndex (so it stays consistent with the other options)?

rejas (changes requested):

modules/default/weather/weather.js
OWL4C: done

modules/default/weather/current.njk
```diff
@@ -28,6 +28,12 @@
                     {% endif %}
                 </span>
             {% endif %}
+            {% if config.showUVIndex %}
+                    <td class="align-right bright uv-index">
```
rejas, line 32: one small thing left I think: intendation seems off here.

rejas (approved):

PR #3119, opened 2023-06-07
Updates notification

bugsounet:
Hi,

Like some default modules, I propose to send an `UPDATES` notification in an array with the git information of these modules

This allows developers to create their own auto-update system (which I've been using in my case since 3 years, with automatic things)

Of course, for security reasons `MagicMirror` is excluded


khassel (approved):

modules/default/updatenotification/git_helper.js
```diff
@@ -171,21 +172,38 @@ class GitHelper {
...
-					gitResultList.push(gitInfo);
+					this.gitResultList.push(gitInfo);
 				}
 			} catch (e) {
 				Log.error(`Failed to retrieve repo info for ${repo.module}: ${e}`);
 			}
 		}
 
-		return gitResultList;
+		return this.gitResultList;
```
rejas, line 189: now that gitResultList is global we dont need to return it, or do we?

rejas (changes requested):

rejas:
Nice PR and functionality. Adding a test for this would be a nice addition (cherry on top) :-)

modules/default/updatenotification/git_helper.js
bugsounet: yes it's a global value only on the `GitHelper` class (constructor) but not available somewhere else (like this.gitRepos)
you have to query `GitHelper` to retrieve this value

I prefert save the result of `getRepos` in a global value
Why ? because, I reuse it for `checkUpdates`

it saves time (perhaps minimal) to pick up from the node helper `repos` values of performFetch function and do the processing. 
The values are already present elsewhere (gitHelper) so much to exploit them directly

bugsounet:
After, we (or I)  can create a template for an auto-update of a module for testing (it's only for modules of developer)

Note:
 * I use my own protocol for this since 3 years now (works fine ! and makes you lazy)
 * Able to update properly a module and restart MM² when update done
 * Catch any error ? (with alert)
 * Naturally, I will not apply this rules to MM² (and default modules)
 
Is this what you are looking for?

rejas:
> After, we (or I) can create a template for an auto-update of a module for testing (it's only for modules of developer)
> 
> Note:
> 
>     * I use my own protocol for this since 3 years now (works fine ! and makes you lazy)
> 
>     * Able to update properly a module and restart MM² when update done
> 
>     * Catch any error ? (with alert)
> 
>     * Naturally, I will not apply this rules to MM² (and default modules)
> 
> 
> Is this what you are looking for?

sorry, dont get what you mean. is it about testing?

will merge the PR anyway even without tests, they can be done anytime later

rejas (approved):

PR #3121, opened 2023-06-08
update roboto fonts to v5

khassel:
there are no "all" files in the v5 versions of roboto-fonts anymore so I used the content of their css files in `roboto.css`.

~~PR is set to draft because @rejas wants to wait until next release is done.~~


rejas (approved):
Lets see how it works out :-)

Issue #3122, opened 2023-06-10
showModule and force, lockString options


bugsounet:
Hi,

I just saw a strange behavior of this function

it will therefore not be possible to do

`show([speed], { force: true, lockString: [Lock-String] })`

I was seeing more behavior like "cancel all lockString" and set the new one

currently the result will be:
  * sets the lockString next
  * reset all locks
  
Is this really the desired result? It shocks me a bit

sdetweil:
the doc 
https://docs.magicmirror.builders/development/core-module-file.html#module-instance-methods
says 
force: true  
clears ALL lock strings in the list 


>Note: The locking mechanism can be overwritten by using the force tag:
```
moduleA.show(0, { force: true });
```
>This will reset the lockstring array, and will show the module.

sdetweil:
so, lockstrings is not a stack (fifo/lifo) but a collection..  

as long as there is something in the list, the module is hidden.. 

when the list is empty the module will be shown.. 

there is no 'checking' of the lockstrings (like you mod x, only mod x can remove)
if you KNOW the string, you can remove it.. the 'value' does NOT have to be a module identifier..

bugsounet:
@sdetweil 

yes yes well I know the rules

I was just surprised that you can't `reset` and put a new lock instead

sdetweil:
ah.. yes,
show never adds a lockstring..

bugsounet:
So, I will to purpose and code this

rejas:
I never looked at the whole lockstring code to be honest (so I dont know how useful or good it is), since it wasnt used really in the default modules. There's only one obscure usage [here](https://github.com/MichMich/MagicMirror/blob/abe5c08a52ffee6b37c30c1f3160e130f0e966a2/modules/default/weather/providers/openweathermap.js#L438) which is never reached anyway. 

So, are you using it @bugsounet in your modules?

sdetweil:
and MMM-Pages, and RemoteControl use it too.. (at least)

bugsounet:
@rejas: you have right, I have to use it in one of my plugin (EXT-Pages)

@sdetweil  also found another module that this option will help

Why ? it's complex to manage lockStrings on start of MM². 

I will propose a PR in this sense

khassel:
> So, in all case, I think we can close this PR

can we close the issue too?

bugsounet:
sorry, I forget to close it :)

PR #3123, opened 2023-06-11
Reset option

bugsounet:
issue #3122

add reset with lockString options

---
Reset all lockStrings and set a new one in replacement


js/main.js
```diff
@@ -224,6 +224,11 @@ const MM = (function () {
 	 * @param {object} [options] Optional settings for the hide method.
 	 */
 	const hideModule = function (module, speed, callback, options = {}) {
+		// reset lockString and prepare for set the new one
+		if (options.reset === true && options.lockString) {
```
rejas, line 228: is checking for options.lockString necessary?

js/main.js
```diff
@@ -268,6 +273,11 @@ const MM = (function () {
 	 * @param {object} [options] Optional settings for the show method.
 	 */
 	const showModule = function (module, speed, callback, options = {}) {
+		// reset lockString and prepare for set the new one
+		if (options.reset === true && options.lockString) {
```
rejas, line 277: I'd call the option "resetLockStrings" to make the usage clearer

rejas (changes requested):

js/main.js
bugsounet: yes, because lockString replacement
reset and lockString works both (security)

js/main.js
bugsounet: sure, You want to rename it or i do it ?

js/main.js
bugsounet: I don't really see the point of destroying the Array without placing a new lock
but ... I may be wrong in my way of seeing things

@rejas: What do you think ?

sdetweil:
but this is on hide, not show

explain the use case

bugsounet:
it's on show too (I don't just think about me)

Personnaly with EXT-Pages:
 * Hide all modules on start (loading and initialized time) : i use `hide()`
 * reset lockString and set new lockString (like a manager)
 * Initialize time done: show needed module and timed rotate display

MMM-Page using:
 * Display all needed module in start (it use `show()`)
   * lockString are set by modules
   * conflit for display modules
 * first rotating time: strange and buggy (modules are displayed when it should be hidden)
 * rotating second time: all is clean

rejas:
As I am not so deep into this issue, I would ask @sdetweil if he agrees with your usecase example.

bugsounet:
Finally, humm... I Don't think really needed. It's just a MM² bug

Why ?

By reading not modified main code, I just understand this:

by using the callback fonction for hide() / show()  (From another module) AND if module updateDom() at the +/- same time
Callbacks function from another module will never executed. See [there](https://github.com/MichMich/MagicMirror/blob/develop/js/main.js#L310-L315)
maybe we have to recode hide/show / updateDom part

It's complex for reproduce but it's happen sometime

After, I can check/recode it. Before, I will Finish my AnimateCSS code 

So, in all case, I think we can close this PR

PR #3135, opened 2023-06-25
Npmrc

bugsounet:
Fix engines check on npm install

it will alow to check engine requirement of `package.json`

actually:
```js
	"engines": {
		"node": ">=16"
	}
```

if requirements are not suitable, it's break installer

```sh
bugsounet@Kubuntu:~/MagicMirror-dev$ npm install
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: magicmirror@2.24.0-develop
npm ERR! notsup Not compatible with your version of node/npm: magicmirror@2.24.0-develop
npm ERR! notsup Required: {"node":">=16"}
npm ERR! notsup Actual:   {"npm":"x.y.z","node":"v14.xx.yy"}
```

actually, it's just a warn:
```sh
bugsounet@KUbuntu:~/MagicMirror-dev$ npm install
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'magicmirror@2.24.0-develop',
npm WARN EBADENGINE   required: { node: '>=16' },
npm WARN EBADENGINE   current: { node: 'v14.xx.yy', npm: 'x.y.z' }
npm WARN EBADENGINE }
...
```
and don't break installation, this can causes errors in the main core of MM²


bugsounet:
au fait, electron v25.2.0 est sorti

main core node version is v18
don't you turn engine to node v18 ?

khassel:
> by the way, electron v25.2.0 is out

meanwhile merged

> main core node version is v18 don't you turn engine to node v18 ?

we will update the docs with next release, see https://github.com/MichMich/MagicMirror-Documentation/pull/163

khassel (approved):

bugsounet:
so, in package.json:

```
	"engines": {
		"node": ">=18"
	}
```

Humm If i add it: i will break suite test for node v16

khassel:
I think we should not force all current node16 installations to use node18 with the new release.

We will update the docs to use node 18 but the engine section above should stay with `>=16` (if we use strict) until node 16 is out of maintenance in Sep. 2023 (see https://endoflife.date/nodejs). I'm open to better ideas.

bugsounet:
don't forget next release of MM² (v2.25.x) will be in october 2023.
node v16 is out of maintenance in Sep. 2023

Maybe changing now is the best way no ?

khassel:
> Maybe changing now is the best way no ?

from my side this is o.k., I'm already on v18 in my docker images (and on v20 in develop images).

If @rejas and @sdetweil agree then the node 16 test must be removed in the github workflow and there are 2  `*.md` files in the `.github` folder where `16` should be replace by `18`.

rejas:
I tend to lean on the "keep node16 until next release in october". Mostly because with only 5 days away from the next release I want to keep the changes at a bare minimum.
One month of unmaintained node16 shoudlnt be too bad, but maybe if @sdetweil wants to turn the scales ;-)

sdetweil:
I will update installer and upgrader to use 18

sdetweil:
yes agree to move to 18, 

it's less than a week of testing tho

bugsounet:
@sdetweil : If you are able to do it, why not
otherwise follow the advice of Veeck

@rejas: but be careful according to the `package-lock.json` (sub-dependencies)
the minimum node version to use is v16.13.0
`cat package-lock.json | grep 'node":' | grep 16`

Remember that we ask in package.json node >=16.0.0 for MM²

What we do ?

rejas:
Can you test this @sdetweil ? I dont have the time and I am not sure its feasible for tomorrows release....

bugsounet:
@rejas: I think it's better to take time.
Don't worry, we see this for next release ;)

rejas (approved):

bugsounet:
@rejas: we can add `audit=false` for not check vulnerabilities

why ?

Because some user like `npm audit` command and can make some package trouble

rejas:
> @rejas: we can add `audit=false` for not check vulnerabilities
> 
> why ?
> 
> Because some user like `npm audit` command and can make some package trouble

Make sense. We developers do the dependency updates anyway on a regular basis.

Will you add it to the branch?

rejas (approved):

bugsounet:
added

Generaly, bots make jobs
but we can check it manualy with `npm audit`

PR #2961, opened 2022-10-29
Switch back to third party fetch lib for all node versions

rejas:
As discussed in https://github.com/MichMich/MagicMirror/pull/2952

PR #3170, opened 2023-08-16
Minimum node version v18

KristjanESPERANTO:
When the next MagicMirror version is released, node v16 will have reached its end of life.

### DIFF
--- a/.github/workflows/automated-tests.yaml
+++ b/.github/workflows/automated-tests.yaml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
     steps:
       - name: "Checkout code"
         uses: actions/checkout@v3
--- a/js/fetch.js
+++ b/js/fetch.js
@@ -11,15 +11,8 @@
  * @class
  */
 async function fetch(url, options = {}) {
-	// const nodeVersion = process.version.match(/^v(\d+)\.*/)[1];
-	// if (nodeVersion >= 18) {
-	// 	// node version >= 18
-	// 	return global.fetch(url, options);
-	// } else {
-	// 	// node version < 18
-	// 	const nodefetch = require("node-fetch");
-	// 	return nodefetch(url, options);
-	// }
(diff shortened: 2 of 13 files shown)


js/fetch.js
```diff
@@ -11,15 +11,8 @@
  * @class
  */
 async function fetch(url, options = {}) {
-	// const nodeVersion = process.version.match(/^v(\d+)\.*/)[1];
-	// if (nodeVersion >= 18) {
-	// 	// node version >= 18
-	// 	return global.fetch(url, options);
```
rejas, line 17: we cant switch to global.fetch now?

.github/workflows/automated-tests.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
```
rejas, line 21: No new version beyond 20 to test against?

js/fetch.js
KristjanESPERANTO: I'd love to get rid of the dependency and my calendars also work with global.fetch, but I'm not sure if they all do.

@khassel, you had dealt with it more intensively. What do you think?

.github/workflows/automated-tests.yaml
KristjanESPERANTO: No, release date of 21.x is 2023-10-17 :slightly_smiling_face:

js/fetch.js
KristjanESPERANTO: I will prepare an extra PR for node-fetch removal. Then we can continue the discussion there. That would have the advantage that we could accept this PR here.

khassel: see https://github.com/MichMich/MagicMirror/pull/2961 and https://github.com/MichMich/MagicMirror/pull/2952

We could try it again and see what happens with the e2e tests, they are meanwhile unstable anyway ...

KristjanESPERANTO: To continue the fetch topic I created PR #3172.

rejas (approved):

PR #3172, opened 2023-08-16
Use internal fetch API

KristjanESPERANTO:
For now, this is a test of how the e2e tests react to the internal fetch function. See #2649, #2961, #2952 and #3170.


rejas:
I guess this is now obsolete...

PR #3176, opened 2023-09-01
npm stop command

bugsounet:
Hi,

I just writed `npm stop` command
It allows to stop MagicMirror²

MagicMirror will create `MagicMirror.pid` with pid number process
`npm stop` will allow to stop this process

It can be used with the new updateNotification module version (when approuved)


bugsounet:
I think, I can code more

 * check if MagicMirror is already launched on start (check pid file for example)
 * delete pid file when MagicMirror is stopped
 
this will avoid being able to launch MagicMirror twice without doing it on purpose

@rejas  what do you think about this ?

rejas:
Not sure if we want to add all these "infrastrcuture" things to the core. @MichMich wasnt very keen on this (and I sometimes agree with him on that). Maybe he and @sdetweil can weigh in on this?

sdetweil:
@rejas @bugsounet  yeh, I'm not sure we need this either. its a short learning experience for the user.  I've been trying to figure out a way to communicate with the user about pm2 and npm start duplication, but haven't found a good place to put that. during install is just too much data going by for the new user..   adding ANOTHER command doesn't seem useful.. as they don't know where to look anyhow.. 

and adding all this code we have to support on all the platforms (remember we document that we run on windows too now) just doesn't make sense to me.

bugsounet:
I wanted to go further in the tests (on Windows for example) but if this functionality does not interest you then let's move on ;)
We have better things to do :)

bugsounet:
can I close this PR ?

PR #3179, opened 2023-09-05
Update dependencies and actions

rejas:


khassel:
no changelog? 😱

the failing test was introduced while I was in holidays:

```bash
FAIL tests/e2e/ipWhitelist_spec.js (20.715 s)
  ● ipWhitelist directive configuration › Set ipWhitelist without access › should return 403

    FetchError: request to http://localhost:8080/ failed, reason: socket hang up

      at ClientRequest.<anonymous> (node_modules/node-fetch/lib/index.js:1501:11)

  ● ipWhitelist directive configuration › Set ipWhitelist without access › should return 403
```

This tests fails always in my setup when running with node v20. The 2 tests run if they were not started together (e.g. delete one), so sounds like a race condition.

Any ideas?

rejas:
> no changelog? 😱

forgot it / to make it draft :-)

KristjanESPERANTO:
> This test fails always in my setup when running with node v20. The 2 tests run if they were not started together (e.g. delete one), so sounds like a race condition.

I ran into the same problem when I was working with the internal fetch API. I was able to work around it by giving the two tests different ports (admittedly this is a bit hacky as it addresses the sypmtom and not the cause): https://github.com/MichMich/MagicMirror/pull/3172/commits/e0b3b01f424a5e88fbac0f07053655a100216427

khassel:
Thanks. I prefer to have running tests so we should use the hack until someone finds the root cause ...

Can you provide a PR with your commit?

KristjanESPERANTO:
> Can you provide a PR with your commit?

I'll try to provide it tomorrow 🙂

PR #3180, opened 2023-09-06
Fix ipWhiteList test: Change ports + use internal fetch API

KristjanESPERANTO:
- Port change should fix a timing issue (mentioned in #3179).
- Use the internal fetch API in preparation for not being dependent on the external fetch package in the future.


tests/e2e/ipWhitelist_spec.js
```diff
@@ -10,7 +10,7 @@ describe("ipWhitelist directive configuration", () => {
 		});
 
 		it("should return 403", async () => {
-			const res = await helpers.fetch("http://localhost:8080");
+			const res = await fetch("http://localhost:8181");
```
khassel, line 13: I think we should still use `helpers.fetch` here and in line 27 because I don't want to mix up using node-fetch and internal fetch. Tested this with `helpers.fetch` and this works too.

Other question: I think you looked into running with internal fetch, are there still problems? Tested internal fetch and all e2e tests run locally ...

KristjanESPERANTO: I accidentally closed the PR :facepalm:  I created a new one where I didn't change fetch: #3181

I still have issues with my PR for the internal fetch. I'm just stuck there and at the moment I don't have a good idea of how to proceed: https://github.com/MichMich/MagicMirror/actions/runs/6090278945/job/16524826973#step:5:1164 Hints would be very welcome.

khassel: thanks, will try to look into this within the next days ...

PR #3184, opened 2023-09-07
use internal fetch as replacement for node-fetch

khassel:
related to #2649

I was able to move to internal fetch and all tests seems fine so far.

But we have one problem with the calendar module. In the docs we have several authentication methods and one of them is `digest`. For this we used `digest-fetch` which needs `node-fetch` (this is not so clear from code but I was not able to get it working).

So we have 3 options:
- remove `digest` as authentication method for calendar module (this is what this PR does at the moment)
- find an alternative npm package or implement the digest stuff ourselves
- use `digest-fetch` and `node-fetch` for calendar module (so they would remain as dependencies in `package.json`)

Opinions? @KristjanESPERANTO @rejas @sdetweil @MichMich


rejas:
Looks like digest-fetch fixed that with their latest release, but thats the v3 line where they switched to ESM.

I dont like removing stuff, so maybe we can provide either a patch for their v2 branch (if there is any) and ask them nicely if they want to release that too.

Or look for an alternative library...

Or is digest-auth so uncommon that we can remove it?

khassel:
> Or is digest-auth so uncommon that we can remove it?

I don't know, I checked when it went to the doc-repo and this happened in the first commit, so I think that is in there because the `request` library supported it.

We have no tests for this and who knows if this ever worked after the change to `digest-fetch` ...

Will dig deeper and try to use the digest auth method with the current setup.

khassel:
I tried to setup a test for digest auth with current setup (`node-fetch` and `digest-fetch`).

Got it not working because [this line](https://github.com/MichMich/MagicMirror/blob/e87f50e64aeaf3a8587f4855f4679c3223959fd0/modules/default/calendar/calendarfetcher.js#L57) in `calendarfetcher.js` will never work because `digest` is an async funtion and without the missing `await` you only get a pending promise back:

```js
fetcher = new digest(auth.user, auth.pass).fetch(url, { headers: headers, agent: httpsAgent });
```

So I think we can remove digest auth because it has never worked since we replaced `request`.

I will make a documentation PR after this is merged.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ _This release is scheduled to be released on 2023-10-01._
 - Update engine node >=18. v16 reached it's end of life. (#3170)
 - Update typescript definition for modules
 - Cleaned up nunjuck templates
+- Replace node-fetch with internal fetch (#2649)
```
KristjanESPERANTO, line 28: Mention additionally the removal of digest-fetch?

KristjanESPERANTO (approved):

CHANGELOG.md
khassel: done.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ _This release is scheduled to be released on 2023-10-01._
 - Update engine node >=18. v16 reached it's end of life. (#3170)
 - Update typescript definition for modules
 - Cleaned up nunjuck templates
+- Replace `node-fetch` with internal fetch (#2649) and remove `digest-fetch`
```
rejas, line 29: I'd mention it in a BREAKING CHANGE section with "Removed digest emthod form calendar becuase it didnt wokr for ... month)" or similar. Make it more clear. Also maybe post it in the forum to ask if anybody complains?

rejas (changes requested):

CHANGELOG.md
khassel: added an extra entry under `Removed`.

> Also maybe post it in the forum to ask if anybody complains?

No, thats waste of time but if you want to do it ...

To be more precise:
- the bug using `digest-fetch` was introduced with release `2.15.0` and nobody complained about it until today (I follow the forum and discord and github). So this is a "good long running test" thats it is not used ...
- I don't know how often calendar auth is used in general, these are edge cases. If you use a public provider as e.g. google you get a private url. I don't know if public providers offer digest auth.
- If really someone is using digest auth (with a very old release) he can also easily transfer to basic auth. Using basic auth with https is more secure as digest auth reading [this article](https://en.wikipedia.org/wiki/Digest_access_authentication)

rejas: Awesome work! Thx a lot!!!

rejas (approved):

KristjanESPERANTO:
@khassel, many thanks for this solid work! 🚀👏

PR #3186, opened 2023-09-09
Mistake on Changelog

bugsounet:
Move AnimateCSS changeLog from  v2.24 to  v2.25


bugsounet:
@rejas: It’s really the only thing we don’t think to check!

rejas (approved):

Issue #3174, opened 2023-08-27
MagicMirror does not respect fullscreen option specificed in config.js


alejo4373:
**Platform**:  
* Raspberry Pi 4. Model B
* Raspberry Pi OS. Debian 11.7
* Electron 25.2.0
* Node : v18.17.1

**MagicMirror² Version**: 2.24.0

**Description**: Setting `electronOptions: { fullscreen: false }` in my config files didn't seem to be working. After digging a bit in `js/electron.js` I found the following snippet

```js
mainWindow.once("ready-to-show", () => {
	mainWindow.setFullScreen(true);
	mainWindow.show();
});
```

This snippet sets the window to full-screen regardless of what's specified on `electronOptions`. Once i removed or commented out the `mainWindow.setFullScreen(true);` statement I noticed that the value from `electronOptions` began applying 

**Steps to Reproduce**: 
1. Set `electronOptions: { fullscreen: false }` on your `config.js` file
2. Run MagicMirror with `npm run start`
3. Notice that the mirror app opens in full-screen despite what we set in `config.js`
4. Locate and open to edit `js/electron.js`
5. Locate the line with `mainWindow.setFullScreen(true)` and remove it or commet it out
6. Toggle the `fullscreen` setting in `electronOptions` in your config file and see how it actually has an effect now after restarting the MagicMirror app.

**Expected Results**:  Setting `electronOptions: { fullscreen: false }` opens the MM app in a window which is not full-screen. The setting is respected and applies

**Actual Results**:  Setting `electronOptions: { fullscreen: false }` in `config.js` is not respected and it does not apply, as such the MM app opens in full-screen

khassel:
You are correct with your analysis.

AFAIK magicmirror is designed to always run in fullscreen so this is probably not a bug but a feature.

What is your use case? What should be displayed on the rest of the screen?

alejo4373:
Well, I wanted it to respect `electronOptions: { fullscreen: false }` while I develop some modules on the PI itself in that way I could have MM running on one half of the screen and my code editor on the other half.

khassel:
AFAIR we did the above hack with `mainWindow.setFullScreen(true)` together with other changes to avoid a white flashing when starting electron. As this doesn't work anymore (white flashing is back) we could remove this line.

[...]

PR #3187, opened 2023-09-11
respect width/height (no fullscreen) if set in electronOptions...

khassel:
... in `config.js`.

Solves #3174 

With getting width/heigt from `electron.screen.getPrimaryDisplay().workAreaSize` introduced with https://github.com/MichMich/MagicMirror/pull/3161 the solution was to remove the `setFullscreen` line.

So per default the fullscreen resolution is used but when someone now uses `electronOptions.width`/`electronOptions.height` in `config.js` these parameters are used and so "no fullscreen" is possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 _This release is scheduled to be released on 2023-10-01._
 
-> ⚠️ This release needs nodejs version > `v18`, older release have reached end of life and will not work!
+> ⚠️ This release needs nodejs version >= `v18`, older release have reached end of life and will not work!
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


bugsounet:
Really!?

bugsounet:
@khassel @rejas

I have this:

image

It will not set fullscreen and I have my taskbar

that why I suggest: `mainWindow.setFullScreen(config.fullscreen);`

khassel:
you are right, I tested this on a system without taskbar, will look again into this ...

rejas:
Shall we revert this? 
And: Do you think its possbile to write a test for this usecase?

khassel:
> Shall we revert this?

no, will provide another PR later

> And: Do you think its possbile to write a test for this usecase?

no idea at the moment

bugsounet:
(revert not sure, You have commit something else after)

Issue #3105, opened 2023-05-16
Automatic Page Reload


QuinnyPig:
Is there a roadmap item (or reason it would be a terrible idea) to alert the client that the server has been restarted / upgraded, and thus to reload the page via 'window.location.reload(true);' or similar? (I run with separate client and server devices, if this sounds ridiculous to your usage model.)

For implementation I was thinking of injecting a UUID at startup into the server response headers that the client caches in local storage, then triggers the reload if responses start coming in with a different UUID; not sure if this should be built into the updatenotifications default module, its own module, or the MM code itself; thoughts?

sdetweil:
a general problem here is that there are no actives connections at the client side, so there is no way to know if the server stopped or was restarted.

 one could write a module that used some regular heartbeat with its node_helper, and cause a page reload . at which point the module starts over again, I think

mmm-watchdig does something similar but gets confused quite often in my experience. it uses time of response to its node_helper

PR #3188, opened 2023-09-11
added automatic client page reload

khassel:
solution for #3105 

~~not sure if updatenotification is the right place, so opinions?~~

now impleneted in `main.js`

### DIFF
--- a/js/defaults.js
+++ b/js/defaults.js
@@ -29,6 +29,11 @@ const defaults = {
 	// e.g. you need to add `frameguard: false` for embedding MagicMirror in another website, see https://github.com/MichMich/MagicMirror/issues/2847
 	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false },
 
+	// properties for checking if server is alive and has same startup-timestamp, the check is per default enabled
+	// (interval 30 seconds). If startup-timestamp has changed the client reloads the magicmirror webpage.
+	checkServerInterval: 30 * 1000,
+	reloadAfterServerRestart: false,
+
 	modules: [
 		{
 			module: "updatenotification",
--- a/js/main.js
+++ b/js/main.js
@@ -568,12 +568,33 @@ const MM = (function () {
 		 */
 		modulesStarted: function (moduleObjects) {
 			modules = [];
+			let startUp = "";
+
 			moduleObjects.forEach((module) => modules.push(module));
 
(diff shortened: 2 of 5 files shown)


bugsounet:
Hi, not sure it's better place
because some user disable updatenotification and it's not the fonction of it :/

For me, maybe best place is [main.js](https://github.com/MichMich/MagicMirror/blob/master/js/main.js#L64-L66)
when all promise of all DOMs are resolved (All DOM_OBJECTS_CREATED of all modules)
I consider that all modules are ready and you code can start from line 67 of main.js

In this case `checkServerInterval` will be a global feature and not a module feature

js/main.js
```diff
@@ -568,12 +568,33 @@ const MM = (function () {
 		 */
 		modulesStarted: function (moduleObjects) {
 			modules = [];
+			let startUp = "";
+
 			moduleObjects.forEach((module) => modules.push(module));
 
 			Log.info("All modules started!");
 			sendNotification("ALL_MODULES_STARTED");
```
bugsounet, line 576: I no longer thought about “ALL_MODULES_STARTED”

it makes more sense ;)

tests/e2e/helpers/mock-console.js
rejas, line 1: Are those changes here related to the PR or just general cleanups?

js/main.js
khassel: @bugsounet is this an "o.k." or a change request?

tests/e2e/helpers/mock-console.js
khassel: removed test changes with latest commit

js/defaults.js
```diff
@@ -29,6 +29,11 @@ const defaults = {
 	// e.g. you need to add `frameguard: false` for embedding MagicMirror in another website, see https://github.com/MichMich/MagicMirror/issues/2847
 	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false },
 
+	// properties for checking if server is alive and has same startup-timestamp, the check is per default enabled
+	// (interval 30 seconds). If startup-timestamp has changed the client reloads the magicmirror webpage.
+	checkServerInterval: 30 * 1000,
+	reloadAfterServerRestart: true,
```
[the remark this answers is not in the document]
khassel: changed it to false

js/main.js
bugsounet: It's ok for me now.
 I have says that before your change to main.js

rejas (approved):

PR #3113, opened 2023-05-28
AnimateCSS

bugsounet:
Hi,

This is my testing code for AnimateCSS for `show()`, `hide()`, `updateDom()`

Naturally, we have to do better !

I voluntarily modify `newsfeed` and `compliments` in order to test

Note: I will correct checks later... it's a test...

PR #3200, opened 2023-09-17
Fix: AnimateCSS merge hide() and show() animated css class when we do multiple call

bugsounet:
PR: #3113  

I see this bugs:

AnimateCSS merge hide() and show() animated css class when we do multiple call
--> result it will stay en hide state

I think event listener (is animateCSS file) is not a proper solution

I correct it with like traditional code with timer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ _This release is scheduled to be released on 2023-10-01._
 - Fix ipWhiteList test (#3179)
 - Fix newsfeed: Convert HTML entities, codes and tag in description (#3191)
 - Respect width/height (no fullscreen) if set in electronOptions (together with `fullscreen: false`) in `config.js` (#3174)
+- Fix: AnimateCSS merge hide() and show() animated css class when we do multiple call
 
 ## [2.24.0] - 2023-07-01
 
--- a/js/animateCSS.js
+++ b/js/animateCSS.js
@@ -130,36 +130,35 @@ const AnimateCSSOut = [
 
 /**
  * Create an animation with Animate CSS
- * resolved as Promise when done
  * @param {string} [element] div element to animate.
  * @param {string} [animation] animation name.
  * @param {number} [animationTime] animation duration.
  */
-function AnimateCSS(element, animation, animationTime) {
-	/* We create a Promise and return it */
-	return new Promise((resolve) => {
-		const animationName = `animate__${animation}`;
-		const node = document.getElementById(element);
-		if (!node) {
-			// don't execute animate and resolve
-			Log.warn(`AnimateCSS: node not found for`, element);
-			resolve();
-			return;
-		}
-		node.style.setProperty("--animate-duration", `${animationTime}s`);
-		node.classList.add("animate__animated", animationName);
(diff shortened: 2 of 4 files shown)


bugsounet:
seriously i have this in local:

```
Test Suites: 39 passed, 39 total
Tests:       565 passed, 565 total
Snapshots:   9 passed, 9 total
```

```
$ node -v
v20.5.1
```

rejas:
> Sometimes it works, sometimes it doesn't

yes, the tests are flaky, as was said 100 times.... What would I give if someone would find the reason for this....

bugsounet:
Tested with a special module (for remote using) with hide() show() function:

* passed: force  random hide/show timer 1000ms (visual testing with log following duration 10min)
* passed: animateIn/animateOut defined in module config 
* passed: with hiddenOnStartup feature
* passed: animation on first start (animateIn defined in module config)
* passed: hide() and show() animateCSS options
* passed updateDom() with animate option
* passed: default MagicMirror animation (if animation defined)

It's ok for me ;)

bugsounet:
@rejas: 

We are pushing this for v2.25

I plan to simplify the code for v2.26

Is it ok for you?

rejas:
> @rejas:
> 
> We are pushing this for v2.25
> 
> I plan to simplify the code for v2.26
> 
> Is it ok for you?

Ok with me as long as the config API doesnt change

bugsounet:
Don't worry about this.
It will not change config API

It's difficult to have what we want, so I will not change ;)

rejas:
> Don't worry about this. It will not change config API
> 
> It's difficult to have what we want, so I will not change ;)

so this PR is ready to merge in your opinion?

bugsounet:
Let me test until Wednesday.

I will be able to test it at home in depth on my MM² Prod (with rpi)
I'm going to put it everywhere to see the behavior !

(reality always takes precedence over a test with many modules rather than with one or two)
(especially when it's complex)

I will come back to you, once this is done

khassel:
> What would I give if someone would find the reason for this....

me too, but it is very hard to solve a timing issue which is 99% not reproducable locally

bugsounet:
I think it's +/- related to node v20 version used by github (maybe buggy version)

Actually we don't have "any" problem with node v18 test suite

We have faculty to choose another node v20 version ?
maybe display node version used for test suite can help ?

honestly, I never see how it works. 
I should look, it might help.
although you manage it very well, sometimes an extra eye can help ;)

khassel:
the e2e tests are also (sometimes) failing in my docker tests ... (more often with v20 but also sometimes with v18)

for me it looks like a previous test has sometimes not finished when next test is starting

feel free to look inside the e2e tests ...

bugsounet:
**Off topic**

@rejas I think to that: It's not possible to save current config.js file and restore it after suite test done.  (if needed) 

Why ? I just lost my test config (this is the last configuration of the test suite that has taken place) and I forget to make a bak

bugsounet:
@rejas:

Tested in my prod (since Sunday evening 10 p.m.) mirror without "animation collision"

It's ok for me, you can merge

Issue #3201, opened 2023-09-18
Config is overwritten when running tests


rejas:
As mentioned [here](https://github.com/MichMich/MagicMirror/pull/3200#issuecomment-1722567078) by @bugsounet the config.js is overwritten when running local tests.

Possible solution: Make a backup of the config before starting the testsuite.

bugsounet:
@rejas: is it really urgent or can it wait until v2.26?

I want to try to familiarize myself with the test suite
But if it's really urgent for you, code it :)

PR #3205, opened 2023-09-21
avoid overriding `config.js` when running tests

khassel:
solves #3201

### DIFF
--- a/tests/utils/weather_mocker.js
+++ b/tests/utils/weather_mocker.js
@@ -35,9 +35,11 @@ const injectMockData = (configFileName, extendedData = {}) => {
 	} else {
 		mockWeather = readMockData("current", extendedData);
 	}
-	let content = fs.readFileSync(path.resolve(`${__dirname}../../../${configFileName}`)).toString();
+	let content = fs.readFileSync(configFileName).toString();
 	content = content.replace("#####WEATHERDATA#####", mockWeather);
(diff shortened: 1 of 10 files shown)


tests/utils/weather_mocker.js
```diff
@@ -35,9 +35,11 @@ const injectMockData = (configFileName, extendedData = {}) => {
 	} else {
 		mockWeather = readMockData("current", extendedData);
 	}
-	let content = fs.readFileSync(path.resolve(`${__dirname}../../../${configFileName}`)).toString();
+	let content = fs.readFileSync(configFileName).toString();
 	content = content.replace("#####WEATHERDATA#####", mockWeather);
-	fs.writeFileSync(path.resolve(`${__dirname}../../../config/config.js`), content);
+	const newFile = configFileName.replace(".js", "_full.js");
```
rejas, line 40: should we name them "_temp" or "_test" maybe?

rejas (changes requested):
thx. the files dont get deleted at the end. maybe delete them after the test?

tests/utils/weather_mocker.js
khassel: found no good name, so going for "_temp"

khassel:
> maybe delete them after the test?

done, also for file `tests/configs/port_variable.js` which is created by template_spec test

rejas (approved):

PR #3207, opened 2023-09-26
Electron based on a severity vulnerability (develop)

bugsounet:
I just see `electron` package used in develop branch have `1 high severity vulnerability`

Detail is [there](https://github.com/advisories/GHSA-j7hp-h8jx-5ppr)

We can patch it with electron v26.2.2 (last version at this day) and will correct it

(ChangeLog is not needed in this case)


rejas:
We usually update dependencies shortly before the release. Might be a good time to update the other dependencies in this pr too, can you do that? (Only minor and patch versions though)

bugsounet:
humm... sure

I wondered if you had "dependabot", naturally yes

But he's weird

Why do you use `package-ecosystem: "github-actions"`  instead of  `package-ecosystem: "npm"` 
like me [there](https://github.com/bugsounet/MMM-GoogleAssistant/blob/dev/.github/dependabot.yml)

Actual result, it's check nothing.

I just replace with `package-ecosystem: "npm"`  and now I make a check
 Insights -> Dependency graph -> Dependabot 

I ask myself the question, is this intended or not?

bugsounet:
maybe it's why you have never any `Bump` for quite some time

bugsounet:
should I reverse `dependabot.yaml` ?

rejas:
> Actual result, it's check nothing.

oh it does check. the github actions we use. 

so, not sure of having dependabot check our npm dependencies too will result in too much noise. maybe worth checking out. so, please revert your change and maybe open a new PR with adding another entry in dependabot for npm dependencies

.github/dependabot.yaml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: "npm"
```
rejas, line 3: yes, please revert

rejas (changes requested):

.github/dependabot.yaml
bugsounet: So... I don't understand why this file because  dependency-review for test suite is [there](https://github.com/MichMich/MagicMirror/blob/master/.github/workflows/depsreview.yaml)

I think create an dependabot monthly is not bad.
sure... we have to check all dependency found but it can help
 * a little time saving
 * just checking is dependency is on package.json (for not update an sub-dependency)
 * able to test the dependabot branch

in all case i will reverse it

rejas (approved):

PR #3218, opened 2023-10-02
Event class bugfix

Ybbet:
Hello,

Bugfix. :-) The "break" line did not allow the "eventClass" option to execute.


Ybbet:
Hi,
I don't think we could use a specific test for that. The test for the option "eventClass" is already there for that.

Issue #3185, opened 2023-09-08
calendar bug ?


bugsounet:
Hi,

I use multi-calandar fetch on this default module (I use one instance)
And i see many fade out/in on each update

I use 4 calendars fetch so I have 4 fade out/in on updateDom (one for each calendar update)

It's not possible to work in async and merge all result and displaying ?
We will have only one fade out/in on update

honestly, I didn't look at the code, I don't know if it's possible (without breaking anything)

Themaster312:
Same here

bugsounet:
working arround ;)

bugsounet:
@Themaster312 

 * What Raspberry Pi are you using ?
 * What's OS ?

PR #3220, opened 2023-10-02
Avoid fade out/in on updateDom when many calendars are used

bugsounet:
related to #3185 

* I have limited updated dom to one update
-> `updateDom()` is activated by a timer which resets when a new event arrives (`CALENDAR_EVENTS`)

* I have set no speed to self update. I think it's not necessary
-> update it directly

If somebody can test and tell me result
In all case, I will patch my prod mirror for testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ❤️ **Donate:** Enjoying MagicMirror²? [Please consider a donation!](https://magicmirror.builders/donate) With your help we can continue to improve the MagicMirror².
 
+## [2.26.0] - Unreleased (`develop` branch)
+
+_This release is scheduled to be released on 2024-01-01._
+
+### Added
+
+### Removed
+
+### Updated
+
+### Fixed
+
 ## [2.25.0] - 2023-10-01
 
 Thanks to: @bugsounet, @dgoth, @dependabot, @kenzal, @Knapoc, @KristjanESPERANTO, @martingron, @NolanKingdon, @Paranoid93, @TeddyStarinvest and @Ybbet.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ _This release is scheduled to be released on 2024-01-01._
 
 ### Fixed
 
+- Avoid fade out/in on updateDom when many calendars are used
+
 ## [2.25.0] - 2023-10-01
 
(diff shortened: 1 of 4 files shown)


bugsounet:
Unfortunately, I see this morning that it doesn't work.
The problem remains the same.
I think it's because of the time taken to fetch  the data.
I'll see how to do it differently

bugsounet:
Really if node 18 is strong too...

```bash
$ npx jest tests/e2e/modules_empty_spec.js 
 PASS   e2e  tests/e2e/modules_empty_spec.js
  Check configuration without modules
    ✓ shows the message MagicMirror² title (203 ms)
    ✓ shows the url of michael's website (202 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        1.682 s

$ node -v
v18.18.0
```

bugsounet:
# v2: I clear all and start again.

* added `updateOnFetch`  feature:
  * `true`: (default) like actually update DOM on each calendar data received
  * `false`: delayed update DOM and automaticaly updated every minutes

* basicaly, I see an internal timer every minutes for update DOM.
   * this internal timer is called on `start()`
   * I move it to a function called : `selfUpdate()`
   * I have added this rule: if `updateOnFetch` is not activated, So updatedDom with animationSpeed

---> +/- no change: just add animation speed or not when DOM Updated

* When `CALENDAR_EVENTS` received (socketNotificationReceived)
  * if `updateOnFetch` is activated (default): no change, it's updateDom() on ach notification received
  * if `updateOnFetch` is NOT activated:
    * updateDom() on first data received for each calendars
    * When others data received (next cycle): don't update DOM and use `selfUpdate()` rules for update DOM

---> Final result must be, when `updateOnFetch` disabled: ONLY one update DOM every minute

I will patch my prod mirror with this new file for testing (with `updateOnFetch` disabled)
I will see tomorrow morning

bugsounet:
Good news; It works as requested!

rejas:
so, its ready for review @bugsounet ?

bugsounet:
yes, it's ready but something is really not clear for me

I have done some test

1) On my pi4 (without  animate and customCSS )
  * Tested with my prod version: I have the main issue (#3185)  
  * Tested with only clock / calendar: I have the issue too
  * Tested with `export ELECTRON_DISABLE_GPU=1`: I have the same issue

2) On my dev desktop (without animate, customCSS, ....)
 OS Ubuntu: 23.10
 Same prod config or only clock / calendar
 **I have not this main issue !**

maybe pi take many time for change CSS ?
So, I really don't understand

@rejas: what do you think about this ?

PR #3223, opened 2023-10-06
Fix API version in yr weather provider call

### LINKED ISSUE #3227: Weather module (yr) hung on 'Loading ...'
Hi!

Using default weather module with provider 'yr' on MM v2.25.00

Is it just me or is there a problem with the api for 'sunrise'?

In the MM all 3 of my weather modules are stuck on 'Loading ...' since yesterday.

In the Console I can see the following errors:
`Could not load weather data. SyntaxError: Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)
    at JSON.parse (<anonymous>)
    at performWebRequest (utils.js:27:29)`
    
Digging a bit deeper into the Chrome Console I find that the request for:
`https://api.met.no/weatherapi/sunrise/2.3/sun?lat=<XX.XXXX>&lon=<YY.YYYY>&date=2023-10-11&offset=%2B02%3A00`

Gives a response of:
`404 Not Found`
`The specified version number is not supported by this product`

I get the same error when trying the above with curl in a terminal on the MM server.

Apparently the api for 'locationforecast' still works.

Running this with curl in a terminal produces nice json output:
`https://api.met.no/weatherapi/locationforecast/2.0/complete?&altitude=1&lat=<XX.XXXX>&lon=<YY.YYYY>`

*) Yes, I have exchanged my 'lat' and 'lon' with placeholders.

Regards,

   /Jonas...

rejas:
Fixes #3227 

and also
- removes unused code
- de-duplicates code fragments
- fixes typos
- inlines code
- adds more weather util tests

@martingron and @Justheretoreportanissue would you be so kind to check if I didnt mess anything up?


khassel:
LGTM but what about my comment in the issue

> We should introduce apiVersionSunrise and apiVersionLocationforecast so users can change this in config.js.

can you add this too? So users could change this in `config.js` without waiting for next release or hacking the code

rejas:
> can you add this too? So users could change this in `config.js` without waiting for next release or hacking the code

will do that, until then I revert this to a "draft"

matt-thurling:
I noticed that sometime in the last couple of days they've switched off the 2.3 API for the sunset data too (i.e. line 355) so that's now returning a 404.
I've had a quick look at the [doco](https://api.met.no/weatherapi/sunrise/3.0/documentation) and api response, and it appears that the 3.0 API has changed a fair bit for sunset - so wouldn't be as simple as bumping the number. Though it does look like you can now give it a date including timezone to get the sunrise/sunset data so the code in this provider which has logic for whether it should look at the "today" data or the "tomorrow" data could probably be simplified to just request whichever is needed. Would you prefer me to raise that as a seperate issue?

khassel (approved):

Issue #3212, opened 2023-09-29
Feature - Custom Positions


Entropy159:
It is really annoying being stuck to the limiting default positions built into Magic Mirror. I would like to be able to set custom X and Y values for positioning modules.

sdetweil:
You can use position absolute in the CSS and move your module anywhere you like, regardless of the position setting

RahulSoni2532:
can you tell me where you want to move this 

![Screenshot 2023-10-10 224656](https://github.com/MichMich/MagicMirror/assets/146342894/9a77b1f3-d645-42b6-a855-a34d95dc38b3)

RahulSoni2532:
i can work on this

Entropy159:
Something interesting I've noticed is when I set an absolute position in the CSS, it's still relative to the area and gets messed up when other modules are toggled.

khassel:
[this thread in the forum](https://forum.magicmirror.builders/topic/18069/multiple-modules-in-a-region) is an example how to override module positions.

I don't think we will change something concerning the module positions which will break existing mm instances.

sdetweil:
@Entropy159   
> when I set an absolute position in the CSS, it's still relative to the area and gets messed up when other modules are toggled.

that is not correct... absolute is absolute. 

see here on how to use the developers console.
https://forum.magicmirror.builders/topic/14862/help-with-a-couple-css-issues?_=1696866322265

you can see and change the classes & styles live..

Entropy159:
I have the CSS for the custom positioning, and it looks like this:
#module_14_MMM-TouchButton {
 top: 150px;
 left: 200px;
 position: absolute !important;
}
When I toggle the modules above it, the button moves to the right. I used the developer console, and it's applying the position correctly, but it still seems like it's relative.

sdetweil:
yes, you are using the relative dimension tags..  top and left

use x and y

Entropy159:
X and Y don't seem to do anything...

PR #3231, opened 2023-10-12
allow custom module positions

khassel:
I'm not sure if this should be merged but I tested this because of these discussions:
- https://github.com/MichMich/MagicMirror/issues/3212
- https://forum.magicmirror.builders/topic/18069/multiple-modules-in-a-region/17

With this PR you can
- define own module positions as html in `config.js`, key is `modulePositions`
- define own css styles in a new file in the `css` directory, the name of this file is specified in `config.js`, key is `modulePositionsCss`

So let me know whether we should offer this or not (no problem for me if not). This PR could be a basis for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ _This release is scheduled to be released on 2024-01-01._
 
 ### Added
 
+- custom module positions
+  reason see e.g.
+
+  - https://github.com/MichMich/MagicMirror/issues/3212
+  - https://forum.magicmirror.builders/topic/18069/multiple-modules-in-a-region/17
+
+  usage:
+
+  - define own module positions as html in `config.js`, key is `modulePositions`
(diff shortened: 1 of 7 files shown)


bugsounet:
@khassel 

and do you think about change module position when MM² is running ?
some users ask to me this feature when whey use some "pages" 3rd party modules

sample using with an pages module
  * page 1: weather in `top_right` // clock in `top_left` // ...
  * page 2 : weather in `top_left` // clock in `top_right` // ...

Actually they have to use multi-instance of a module with `hide` // `show` feature

sdetweil:
there is a module that does this now in 3rd party list

khassel:
> Note: sometimes it would be good to have a TODO file :p

I think we should do this with issues, there is already #3212 but I can create a new one for this feature (and assign it to you if you want).

sdetweil:
I submitted #3518 which reads index.html to get the defined regions for this run

sdetweil:
this is superceded by #3518

PR #3247, opened 2023-10-23
Update github test action

rejas:
... add node 21 to the tests and also update dependencies


khassel:
I was thinking about adding node v21 to the tests, may you can add this one liner here?

rejas:
> I was thinking about adding node v21 to the tests, may you can add this one liner here?

sure

Issue #3256, opened 2023-10-31
calendar crashes when bad dates returned from rrule.between in calendarfetcherutils.js


sdetweil:
version, all

when rrule (provided by node-ical, now called ical)  gets confused, it returns an array of invalid date objects  from the between function.
```
 dates: [ null,null,null,null....]
```
the code checking them then crashes because of the invalid date object.

there are an unknown number of causes

like this event from in the past 
```
BEGIN:VEVENT
DTSTAMP:20210413T203456Z
UID:E689AEB8C02C4E2CADD8C7D3D303CEAD0
DTSTART;TZID="Amsterdam, Belgrade, Berlin, Brussels, Budapest, Madrid, Paris, Prague, Stockholm":20210415T190000
DTEND;TZID="Amsterdam, Belgrade, Berlin, Brussels, Budapest, Madrid, Paris, Prague, Stockholm":20210415T210000
CLASS:PUBLIC
LOCATION:albert heijn
SUMMARY:xxx xxxx
SEQUENCE:10
RRULE:FREQ=DAILY;UNTIL=20210418T170000Z
EXDATE;TZID="Amsterdam, Belgrade, Berlin, Brussels, Budapest, Madrid, Paris, Prague, Stockholm":20210417T190000
EXDATE;TZID="Amsterdam, Belgrade, Berlin, Brussels, Budapest, Madrid, Paris, Prague, Stockholm":20210416T190000
EXDATE;TZID="Amsterdam, Belgrade, Berlin, Brussels, Budapest, Madrid, Paris, Prague, Stockholm":20210415T190000
BEGIN:VALARM
ACTION:DISPLAY
TRIGGER;RELATED=START:-PT15M
END:VALARM
END:VEVENT
```

sdetweil:
I have a PR for this, but can't commit it as the lint stage fails for the new ics file..  
```
(base) sam@sams:~/MagicMirror$ git commit -m "fix rrule.between crash on bad dates #3256"
/usr/local/bin/npm

> magicmirror@2.26.0-develop lint:staged
> lint-staged

✔ Preparing lint-staged...
⚠ Running tasks for staged files...
  ❯ package.json — 4 files
    ❯ * — 4 files
      ✖ prettier --write [FAILED]
    ❯ *.js — 3 files
      ✖ eslint [KILLED]
    ↓ *.css — no files
↓ Skipped because of errors from tasks.
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

[...]

PR #3257, opened 2023-10-31
fix  crash on rrule.between returned bad dates #3256

sdetweil:
Fixes: #3256 

BUT.. the testcase is inconclusive.. as the code FAILS without the fix, BUT somehow RETURNS 0 entries..
in real life run the node helper fails, and all calendar processing stops.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,6 @@
+.eslintignore
+.prettierignore
 /config
 /coverage
 package-lock.json
+**.ics
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ _This release is scheduled to be released on 2024-01-01._
 - Fix cloneObject() function to respect RegExp (#3237)
 - Fix newsfeed module for feeds using "a10:updated" tag (#3238)
 - Fix issue template (3167)
+- Fix #3256 filter out bad results from rrule.between
 
 ## [2.25.0] - 2023-10-01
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -30,7 +30,8 @@ _This release is scheduled to be released on 2024-01-01._
 - Fix yr API version in locationforecast and sunrise call (#3227)
 - Fix cloneObject() function to respect RegExp (#3237)
 - Fix newsfeed module for feeds using "a10:updated" tag (#3238)
-- Fix issue template (#3167)
+- Fix issue template (3167)
```
rejas, line 33: please re-add the # :-)

tests/configs/modules/calendar/bad_rrule.js
```diff
@@ -0,0 +1,27 @@
+/* MagicMirror² Test config default calendar with auth by default
+ *
+ * By Rodrigo Ramírez Norambuena https://rodrigoramirez.com
```
rejas, line 3: you can put your own name here

rejas (changes requested):

.prettierignore
```diff
@@ -1,3 +1,6 @@
+.eslintignore
+.prettierignore
```
KristjanESPERANTO, line 2: Why should prettier ignore these two files?

sdetweil:
because they produced errors in prettier about no ruleset.  same as the ics

.prettierignore
khassel: these changes are meanwhile already on `develop`

sdetweil: yes, I took them from the other calendar fix

khassel (approved):

rejas (approved):

Issue #3254, opened 2023-10-30
unit test fails


khassel:
see https://github.com/MichMich/MagicMirror/pull/3251#issuecomment-1786181663

I think this was introduced with https://github.com/MichMich/MagicMirror/pull/3087 , see also discussions here https://github.com/MichMich/MagicMirror/pull/3073

Test were fine as long we had Central European Summer Time

khassel:
I think the solution is to use `UTC` as timezone because because it has [no daylight saving time](https://stackoverflow.com/questions/5495803/does-utc-observe-daylight-saving-time).

khassel:
back again, don't know if related to some of the latest changes

```bash
FAIL unit tests/unit/modules/default/utils_spec.js
  Default modules utils tests
    performWebRequest
      When using cors proxy
        ✓ Calls correct URL once (13 ms)
        ✓ Sends correct headers
      When not using cors proxy
        ✓ Calls correct URL once (1 ms)
        ✓ Sends correct headers (1 ms)
      When receiving json format
        ✓ Returns undefined when no data is received
        ✓ Returns object when data is received (2 ms)
        ✓ Returns expected headers when data is received (1 ms)
    formatTime
      ✓ should convert correctly according to the config (1 ms)
      ✕ should convert correctly into another timezone (5 ms)
  ● Default modules utils tests › formatTime › should convert correctly into another timezone
    expect(received).toBe(expected) // Object.is equality
    Expected: "09:13"
    Received: "08:13"
      161 | 					time
      162 | 				)
    > 163 | 			).toBe("09:13");
          | 			  ^
      164 | 		});
      165 | 	});
      166 | });
      at Object.toBe (tests/unit/modules/default/utils_spec.js:163:6)
```

khassel:
not related to latest changes, but the test uses as timezone `America/Toronto` and the summertime there ended at 5. Nov. so new problem since then.

Looking into this I think this will never work because the tested function `formatTime` uses `let date = moment(time);` which will always give a local date.

So I will remove this test now to get the unit tests working again.

@rejas because you introduced this with 32192d16983f9b9c84f298c70b15bcb97fa91a1d feel free to provide a new test for this which work after every timezone change ...

PR #3258, opened 2023-10-31
hotfix for failing unit test

khassel:
I know this isn't a real solution, but it's the quickest way to get it working again.

A real solution must be found later.


rejas (approved):

rejas:
Could you open an issue for the real fix too?

khassel:
see #3254

PR #3270, opened 2023-11-20
Enable and apply ESLint Jest rules

KristjanESPERANTO:
Jest was in the plugin array of the ESLint configuration, but no rules were enabled. So ESLint hasn't checked any Jest rules yet.

So I activated the recommended Jest rules and added a few more. Then I fixed the issues (mostly automatically). I have deactivated the rules "jest/expect-expect" and "jest/no-done-callback" for the time being, as they would have entailed major changes. I didn't want to make the PR too big.

I'm not a Jest expert, but the changes so far look good to me. What do you think of that @khassel? :slightly_smiling_face:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
-	"extends": ["eslint:recommended", "plugin:import/recommended", "plugin:jsdoc/recommended", "plugin:prettier/recommended"],
-	"plugins": ["jest"],
+	"extends": ["eslint:recommended", "plugin:import/recommended", "plugin:jest/recommended", "plugin:jsdoc/recommended", "plugin:prettier/recommended"],
+	"plugins": [],
 	"env": {
 		"browser": true,
 		"es2022": true,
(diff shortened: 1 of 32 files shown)


khassel (approved):

khassel:
Thanks, LGTM!

Will merge it if @rejas has no complaints ...

rejas:
No complaints

rejas (approved):

KristjanESPERANTO:
If anyone wants to take a look. I've been working on enabling the "jest/expect-expect" and "jest/no-done-callback" rules for a while now and adapting the code accordingly. I haven't come to an acceptable result and am putting it aside.

khassel:
is your work [here](https://github.com/KristjanESPERANTO/MagicMirror/tree/eslint-jest)?

KristjanESPERANTO:
Ah, that's an attempt that I wasn't happy with either. If you want to investigate this, it might be better to ignore that attempt and change the rules yourself from "off" to "warn" and look at the issues from `npm run test:js`.

KristjanESPERANTO:
In any case, I think it's worth taking a look especially at rule ["jest/no-done-callback"](https://github.com/jest-community/eslint-plugin-jest/blob/v27.6.0/docs/rules/no-done-callback.md), because its implementation might help us to avoid problems that we sometimes have with asynchronous code.

PR #3281, opened 2023-12-01
Fix updatenotification

bugsounet:
Sometime, `pm2 jlist` don't send an json reponse
Catch error if happen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _This release is scheduled to be released on 2024-01-01._
 - Fix #3256 filter out bad results from rrule.between
 - Fix calendar events sometimes not respecting deleted events (#3250)
 - Fix electron loadurl locally on Windows when address "0.0.0.0" (#2550)
+- Fix updatanotification (update_helper.js): catch error if reponse is not an JSON format (check PM2)
 
 ## [2.25.0] - 2023-10-01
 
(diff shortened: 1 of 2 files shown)


bugsounet:
Note:

For me, best way is using `pm2` library directly because with command line sometine we can have some pm2 warn (update warn for example) and it break response
but there is some `moderate severity vulnerabilities` with this library

@rejas: So, what can I do ?

rejas:
@bugsounet do you have a link to that vulnarbility?

bugsounet:
seams to be last current `axios` vulnerabilities

```sh
# npm audit report

axios  0.8.1 - 1.5.1
Severity: moderate
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
fix available via `npm audit fix --force`
Will install pm2@2.10.4, which is a breaking change
node_modules/axios
  @pm2/js-api  *
  Depends on vulnerable versions of axios
  node_modules/@pm2/js-api
    pm2  >=3.0.0
    Depends on vulnerable versions of @pm2/js-api
    node_modules/pm2

3 moderate severity vulnerabilities
```

rejas (approved):

bugsounet:
I try to code (recode) it why `pm2` library in draft and we see ?

PR #3284, opened 2023-12-04
Show moon phase in clock module

bnitkin:
This change replaces the font-awesome moon icon and percent-lit with an icon showing the current lunar phase.

It uses emoji, which may not be installed on all machines. The fallback text version is backwards (the dark part of the moon is text-color, which is normally black but white in MagicMirror).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ _This release is scheduled to be released on 2024-01-01._
 - Update newsfeed: Use `html-to-text` instead of regex for transform description
 - Review ESLint config (#3269)
 - Updated dependencies
-- Clock module: swapped moon percent lit for current phase icon
+- Clock module: optionally display current moon phase in addition to rise/set times
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ _This release is scheduled to be released on 2024-01-01._
 - Update newsfeed: Use `html-to-text` instead of regex for transform description
 - Review ESLint config (#3269)
 - Updated dependencies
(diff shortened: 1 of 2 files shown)


bnitkin:
Here's the clock showing updates

2023-12-04-165622_370x181_scrot

KristjanESPERANTO:
I like the idea and the implementation with emojis! :smiley:  Can you provide a screenshot of what it looked like before for comparison? I can't test it right now, but wasn't there a percentage behind the moon?

khassel:
grafik

I'm not sure if all people like the yellow color so maybe a better approach would be to add a new config option or change `showMoonTimes` from `true/false` to e.g. `none/percent/emoji`

bnitkin:
I like that idea. I can take a swing at that over the weekend and push some updates!

bnitkin:
Just pushed an update - thoughts? Here's a screenshot of the various configs:
2023-12-12-185901_248x597_scrot

The parsing logic is fairly relaxed. A falsy value with turn times off; anything truthy enables times (so people with `showMoonTimes=true` will keep seeing times). "percent", "phase", and "both" enable the percent-lit and/or phase image.

Top to bottom is
```
false
"times"
"percent"
"phase"
"both"
```

rejas:
Thanks. Can you now prepare a PR for the Documentation repo?

khassel:
@rejas I'm getting 

```bash
> magicmirror@2.26.0-develop test:prettier
> prettier . --check
Checking formatting...
[warn] js/class.js
[warn] Code style issues found in the above file. Run Prettier to fix.
```

since this is merged on `develop`

khassel:
seems unrelated, maybe new deps

PR #3285, opened 2023-12-06
updatenotification: update_helper.js recode with pm2 library

bugsounet:
Like i have say some few days ago.

 * recode: `update_helper.js` with `pm2` library
 * fix: default config -> `updates` is a array
 * delete: `command-exists` library (not used)
 * delete: `PM2_GetList()`  function (not used)
 * add: check `updates.length` (prevent crash)
 * add: `[PM2]` tag in log (for better visibility)
 * add: `pm2` library


bugsounet:
Like I have says [this](https://github.com/MichMich/MagicMirror/pull/3281#issuecomment-1837084761)

sdetweil:
we need node-ical 0.17.0  we aren't using the url ics approach node-ical provides..

bugsounet:
> yeah, not sure about that axios vulnarbility from node-ical.... maybe we can add the v1.6.0 manually in our package.json and later remove it once node-ical update is working? would that work?

tested and not work because [@pm2/js-api](https://www.npmjs.com/package/@pm2/js-api) force  another version in its own package.json

rejas:
The headache I have is that our "Review Dependencies" check will always error out once we merge this. And a check that always fails is worthless.

CHANGELOG.md
```diff
@@ -41,8 +41,7 @@ _This release is scheduled to be released on 2024-01-01._
 - Fix #3256 filter out bad results from rrule.between
 - Fix calendar events sometimes not respecting deleted events (#3250)
 - Fix electron loadurl locally on Windows when address "0.0.0.0" (#2550)
-- Fix updatanotification (update_helper.js): catch error if reponse is not an JSON format (check PM2)
-- Fix missing typeof in calendar module
```
KristjanESPERANTO, line 45: The deletion of line 45 was not intentional, was it?

rejas: maybe a oversight when i merged develop into it. will merge develop again now that other PRs got merged and re-add it

bugsounet:
So I have 2 solutions:

*  Waiting for an `pm2` fix (an issue/PR) is ready for this
   * make a draft of this PR and waiting...
 
or

* make a fork of `pm2`
  * correct this vulnerability
  * publish it into my npm account and use it
  * waiting for original `pm2` fix and switch into it in next MM² version

bugsounet:
That you can see my fix works
Note: We don't use `@pm2/js-api` in this part of code but ... it's on main package.json of `pm2`

now @rejas what I do ?
 * Publish my own package (pm2 and js-api) to npm and use it temporarily?
 * use it directly from my github account (like actually)
 or
 * waiting for an official fix ? and still this PR in draft ?

bugsounet:
again another  solution: recode `pm2` and limit it to our own use (for MagicMirror² only)

  - just using: start / stop / restart / list / status / log ... (with no advanced function)
  - it will be automaticaly linked to original `pm2` (like actually)

we can call It something like 'MM-pm2` (I'm able to do it)

sdetweil:
I would wait. there is no rush for this 

as a user I do not want auto update.

bugsounet:
humm... so best way is not code this functionality and close this PR ?

sdetweil:
no. wait til the needed support is there.

package.json
```diff
@@ -87,6 +86,7 @@
 		"module-alias": "^2.2.3",
 		"moment": "^2.29.4",
 		"node-ical": "0.16.1",
+		"pm2": "https://github.com/bugsounet/pm2#build",
```
KristjanESPERANTO, line 89: Why is the official package not used here?

bugsounet: it's a draft... I test something but i plan to use official packet
I'm working about something else actually. I will see after ;)
Don't worry

bugsounet:
recoded for v2.27.x with #3332

Issue #3226, opened 2023-10-09
disable/enable gpu when running under electron


BKeyport:
Does anyone know what this is about, and how to fix it? 

[28358:1008/184025.188520:ERROR:object_proxy.cc(590)] Failed to call method: org.freedesktop.portal.Settings.Read: object_path= /org/freedesktop/portal/desktop: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.portal.Desktop was not provided by any .service files

-- while we're at it, can we modify base code to start up with the mesa-loader error disabled? 

Thanks!

sdetweil:
for the mesa loader thing add
```
export ELECTRON_DISABLE_GPU=1
```
in from of the npm start 
if you use my install script, that's
```
~/MagicMirror/installers/mm.sh
```

the other error is an assumption on electrons part that there is a desktop control file there  but it doesn't.exist on most of our systems anymore. nothing we can do about it

BKeyport:
I'm aware how to fix it after using your installer - I'm suggesting we come up with some way to proactively disable it.

As for the other one, good to know it's something not able to be fixed. Darn assumptions...

khassel:
this is coded [here](https://github.com/MichMich/MagicMirror/blob/343e7de7bd295a69e68e4ee520552e0785a99e1c/js/electron.js#L13)

I introduced this with the defensive approach, I'm open to doing it the other way around (e.g. with a new env var `ELECTRON_ENABLE_GPU`).

Opinions @sdetweil @rejas

sdetweil:
@khassel I'm ok with enable_gpu
most wouldn't need it anyhow.

jalibu:
Modules with CSS transition animations (like stock tickers) still need it.
Some MMM-Jast users report choppy rendering since 2.26.0

I would welcome a more user-friendly way to activate this option. Not all users know how to customize the startup commands and how to set ENV variables.

khassel:
> What about introducing .env files for configuration?

AFAIS this is built-in with nodejs version v20.6.0, electron is still on v18 so we would need an extra npm package for this (e.g. `dotenv`).

Beside this we have only 3 env vars: `ELECTRON_DISABLE_GPU`, `MM_CONFIG_FILE`, `MM_PORT` and the 2 last one are for running 2 mm instances from the same mm dir so it makes no sense to put them in an `.env` file.

So as long as we are talking about 1 variable I would prefer to add another line in `package.json`

[...]

PR #3290, opened 2023-12-13
start electron with --disable-gpu flag

khassel:
can be overriden with env var ELECTRON_ENABLE_GPU=1.

see #3226 

Tests will fail as long as https://github.com/MichMich/MagicMirror/pull/3289 is not merged.


sdetweil (approved):
seems ok. process.env will always be defined

rejas (approved):

PR #3303, opened 2023-12-24
Replace prettier by stylistic to lint JavaScript

KristjanESPERANTO:
In the latest versions of ESLint, more and more formatting rules were removed or declared deprecated. These rules have been integrated into the new Stylistic package (https://eslint.style/guide/why) and expanded.

Stylistic acts as a better formatter  for JavaScript as Prettier.

With this PR there are many changes that make the code more uniform, but it may be difficult to review due to the large amount. Even if I have no worries about the changes, perhaps this would be something for the release after next.

Let me know what you think.


khassel (approved):

khassel:
Thanks! I'm fine with this.

rejas:
I dont mind changing this and see how the stylistic package holds up during 2024 :-)
Cant we get rid of the whole prettier things then? Or why do these still stay in there?

KristjanESPERANTO:
> Cant we get rid of the whole prettier things then? Or why do these still stay in there?

Prettiert makes other files formats (e.g. HTML, JSON, Markdown, YAML) pretty in our project. There are also formatter and linter or ESLint plugins for some of these formats, but at the moment I don't see us getting rid of prettier. And I think prettier does a pretty job for the non JavaScript formats.

rejas (approved):

Issue #3328, opened 2024-01-01
Add additional infos into log


KristjanESPERANTO:
When troubleshooting in the forum or in the GitHub issues, we often get screenshots or photos [like this](https://private-user-images.githubusercontent.com/62189127/293600498-125060ef-8025-4074-89b2-a377c2ceedaa.JPG?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDQxNDQxMTEsIm5iZiI6MTcwNDE0MzgxMSwicGF0aCI6Ii82MjE4OTEyNy8yOTM2MDA0OTgtMTI1MDYwZWYtODAyNS00MDc0LTg5YjItYTM3N2MyY2VlZGFhLkpQRz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAxMDElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMTAxVDIxMTY1MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWNjYTEyM2I4MzY1MmNlODE3MzEwNDc3M2MzZTVkZTQ1MmM3MDcyMzFkZDBjZjI2YTg3ZmFlN2VhNjI5OThkNGQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.600iVgD3CIfvDO9crbsst4-wuykNxTK9qz9qQ1oPDLY).

Wouldn't it be helpful if we could add some more information like node version or OS into the log? Then we wouldn't have to ask the users for it.

Something like this:

```sh
[01.01.2024 22:22.15.318] [LOG]   Starting MagicMirror: v2.26.00
[...]

PR #3331, opened 2024-01-03
Add systeminfo

KristjanESPERANTO:
This is a first attempt to bring additional system information into the console (see #3328). It's certainly not yet perfect, but with the PR we have a better basis for discussion.

I tried to keep the output small so that we get as much information as possible in screenshots.

This is how it looks on my development system.

```bash
[03.01.2024 00:50.19.226] [INFO] System information:
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12
 ### VERSIONS: MagicMirror: 2.27.0-develop; electron: 27.2.0; kernel: 5.10.0-20-amd64; node: 21.1.0; npm: 10.2.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ```
 
 Why is it still a draft:
- [x] I have doubts that utils.js is the right place for the function. What do you think?
=> Update: As long as there is no better idea, it stays there.
- [x] Instead of working through all wishes you expressed in the issue #3328, I only implemented what was easy to achieve. And wanted to hear what you think about this approach.
=> Update: Some added. Of course, more information could be added later, as soon as experience has been gained in productive use.
- [x] I don't quite like the introductory line ("The following lines provide information..."). Should I perhaps simply replace it with "System information:"?
=> Update: Changed to "System information:"
 
 [Here](https://github.com/sebhildebrandt/systeminformation#function-reference-and-os-support) you can see what information we could easily add with the systeminformation package.
 
 It would be interesting how the raspberry field is filled on a raspi system and with docker there should be another line, but I can't easily test that now.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -21,6 +21,9 @@ const defaultModules = require(`${__dirname}/../modules/default/defaultmodules`)
 global.version = require(`${__dirname}/../package.json`).version;
 Log.log(`Starting MagicMirror: v${global.version}`);
 
+// Log system information.
+Utils.logSystemInformation();
+
 // global absolute root path
 global.root_path = path.resolve(`${__dirname}/../`);
 
--- a/js/utils.js
+++ b/js/utils.js
@@ -25,7 +25,7 @@ module.exports = {
 	async logSystemInformation  () {
 		try {
 			const staticData = await si.getStaticData();
-			let systemDataString = "The following lines provide information about your system and may be of interest when troubleshooting.";
+			let systemDataString = "System information:";
 			systemDataString += `\n ### SYSTEM:   manufacturer: ${staticData["system"]["manufacturer"]}; model: ${staticData["system"]["model"]}; raspberry: ${staticData["system"]["raspberry"]}; virtual: ${staticData["system"]["virtual"]}`;
 			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}`;
 			systemDataString += `\n ### VERSIONS: MagicMirror: ${global.version}; electron: ${global.electronVersion}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
--- a/js/utils.js
+++ b/js/utils.js
@@ -11,7 +11,7 @@ const si = require("systeminformation");
 try {
 	global.electronVersion = require(`${__dirname}/../node_modules/electron/package.json`).version;
 } catch (error) {
-	Log.error(`Can't find electron. Have you performed 'npm run install-mm'? ${error}`);
+	global.electronVersion = "not installed";
 }
 
 module.exports = {
--- a/js/utils.js
+++ b/js/utils.js
@@ -23,7 +23,6 @@ module.exports = {
 			systemDataString += `\n ### SYSTEM:   manufacturer: ${staticData["system"]["manufacturer"]}; model: ${staticData["system"]["model"]}; raspberry: ${staticData["system"]["raspberry"]}; virtual: ${staticData["system"]["virtual"]}`;
 			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}; arch: ${staticData["os"]["arch"]}; kernel: ${staticData["versions"]["kernel"]}`;
 			systemDataString += `\n ### VERSIONS: electron: ${process.versions.electron}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
-			if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;
 			systemDataString += `\n ### OTHER:    timeZone: ${Intl.DateTimeFormat().resolvedOptions().timeZone}`;
 			Log.info(systemDataString);
 		} catch (e) {
--- a/js/utils.js
+++ b/js/utils.js
@@ -24,6 +24,7 @@ module.exports = {
 			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}; kernel: ${staticData["versions"]["kernel"]}`;
 			systemDataString += `\n ### VERSIONS: electron: ${process.versions.electron}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
 			if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;
+			systemDataString += `\n ### OTHER:    timeZone: ${Intl.DateTimeFormat().resolvedOptions().timeZone}`;
 			Log.info(systemDataString);
 		} catch (e) {
 			Log.error(e);
--- a/js/utils.js
+++ b/js/utils.js
@@ -8,12 +8,6 @@ const colors = require("colors/safe");
 const Log = require("logger");
 const si = require("systeminformation");
 
-try {
-	global.electronVersion = require(`${__dirname}/../node_modules/electron/package.json`).version;
-} catch (error) {
-	global.electronVersion = "not installed";
-}
-
 module.exports = {
 	colors: {
 		warn: colors.yellow,
@@ -28,7 +22,7 @@ module.exports = {
 			let systemDataString = "System information:";
 			systemDataString += `\n ### SYSTEM:   manufacturer: ${staticData["system"]["manufacturer"]}; model: ${staticData["system"]["model"]}; raspberry: ${staticData["system"]["raspberry"]}; virtual: ${staticData["system"]["virtual"]}`;
 			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}`;
-			systemDataString += `\n ### VERSIONS: MagicMirror: ${global.version}; electron: ${global.electronVersion}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
+			systemDataString += `\n ### VERSIONS: MagicMirror: ${global.version}; electron: ${process.versions.electron}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
 			if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;
 			Log.info(systemDataString);
 		} catch (e) {
--- a/js/utils.js
+++ b/js/utils.js
@@ -22,7 +22,7 @@ module.exports = {
 			let systemDataString = "System information:";
 			systemDataString += `\n ### SYSTEM:   manufacturer: ${staticData["system"]["manufacturer"]}; model: ${staticData["system"]["model"]}; raspberry: ${staticData["system"]["raspberry"]}; virtual: ${staticData["system"]["virtual"]}`;
 			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}`;
-			systemDataString += `\n ### VERSIONS: MagicMirror: ${global.version}; electron: ${process.versions.electron}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
+			systemDataString += `\n ### VERSIONS: electron: ${process.versions.electron}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
 			if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;
 			Log.info(systemDataString);
 		} catch (e) {
--- a/js/utils.js
+++ b/js/utils.js
@@ -21,8 +21,8 @@ module.exports = {
 			const staticData = await si.getStaticData();
 			let systemDataString = "System information:";
 			systemDataString += `\n ### SYSTEM:   manufacturer: ${staticData["system"]["manufacturer"]}; model: ${staticData["system"]["model"]}; raspberry: ${staticData["system"]["raspberry"]}; virtual: ${staticData["system"]["virtual"]}`;
-			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}`;
-			systemDataString += `\n ### VERSIONS: electron: ${process.versions.electron}; kernel: ${staticData["versions"]["kernel"]}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
+			systemDataString += `\n ### OS:       platform: ${staticData["os"]["platform"]}; distro: ${staticData["os"]["distro"]}; release: ${staticData["os"]["release"]}; kernel: ${staticData["versions"]["kernel"]}`;
+			systemDataString += `\n ### VERSIONS: electron: ${process.versions.electron}; node: ${staticData["versions"]["node"]}; npm: ${staticData["versions"]["npm"]}; pm2: ${staticData["versions"]["pm2"]}; docker: ${staticData["versions"]["docker"]}`;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: js/app.js [hunk 1]
```diff
@@ -21,6 +21,9 @@ const defaultModules = require(`${__dirname}/../modules/default/defaultmodules`)
 global.version = require(`${__dirname}/../package.json`).version;
 Log.log(`Starting MagicMirror: v${global.version}`);
 
+// Log system information.
+Utils.logSystemInformation();
```
rejas, line 25: I think putting it in Utils is ok for now. Moving it to somewhere else if someone has a better idea is always possible

js/utils.js
```diff
@@ -5,12 +5,34 @@
...
 		warn: colors.yellow,
 		error: colors.red,
 		info: colors.blue,
 		pass: colors.green
+	},
+
+	async logSystemInformation  () {
+		try {
+			const staticData = await si.getStaticData();
+			let systemDataString = "The following lines provide information about your system and may be of interest when troubleshooting.";
```
rejas, line 28: Yeah, something short like "System Information" should be enough

rejas:
Thanks for whipping up this PR. Cant wait to see it moved from Draft into real.

KristjanESPERANTO:
Thanks for the feedback. I have changed the first line. With this I take back the draft state 🙂

rejas (approved):

rejas:
If nobody comes up with more wishes for the ouput in the next 24h I will merge it :-)

bugsounet:
why not using `process.versions.electron`  instead of `require(`${__dirname}/../node_modules/electron/package.json`).version;
` ?

many values are inside process object (and many other) just check it :)

sdetweil:
also on armv6l, there is no electron, as its optional, and won't be installed..

AND we are seeing a lot of electron fail install on upgrade .. but work manually after,, don't know why

sdetweil:
virtual.. does this detect in a VM? (vmware, ...) , or macos

bugsounet:
[systeminformation docs](https://systeminformation.io/#docs) says mac is supported, VM... humm  virtual box VMs only ?

KristjanESPERANTO:
> why not using process.versions.electron instead of ...

I have tested it. At the beginning electron is not yet initialized and then this variable does not exist. At what point did you call the variable? I would like to output the system information before the modules.

> also on armv6l, there is no electron

I have modified it a bit, now no message is displayed if electron is not installed. It will only indicate in the list that it is not installed. Like this:

```bash
[03.01.2024 18:35.53.972] [INFO]  System information:
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12
 ### VERSIONS: MagicMirror: 2.27.0-develop; electron: not installed; kernel: 5.10.0-20-amd64; node: 21.1.0; npm: 10.2.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ``

**Comment by khassel** [comment]
LGTM!

Here the results on my windows machine using docker container with wsl2:

```bash

npm run server
 ### SYSTEM:   manufacturer: Microsoft; model: Docker Container; raspberry: undefined; virtual: true
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12
 ### VERSIONS: MagicMirror: 2.27.0-develop; electron: 27.2.0; kernel: 5.15.133.1-microsoft-standard-WSL2; node: 21.5.0; npm: 10.2.4; pm2: ; docker:

npm run start
 ### SYSTEM:   manufacturer: Microsoft; model: Docker Container; raspberry: undefined; virtual: true
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12
 ### VERSIONS: MagicMirror: 2.27.0-develop; electron: 27.2.0; kernel: 5.15.133.1-microsoft-standard-WSL2; node: 18.17.1; npm: 10.2.4; pm2: ; docker:
```

The difference between the two start variants is the node version, under electron the node version of electron is used, so everything works as expected :)

One cosmetic thing: The magicmirror version is already shown twice in the logs, e.g.

```bash
> magicmirror@2.27.0-develop start
> DISPLAY="${DISPLAY:=:0}" ./node_modules/.bin/electron js/electron.js
[03.01.2024 19:15.41.671] [LOG]   Starting MagicMirror: v2.27.0-develop
```

so I think we don't need this information again.

Will do some tests later using a raspberry pi.

sdetweil:
is there CPU architecture info 
armv6l/armv7l/aarch64, ......

khassel:
So here my test on a raspi4 but - as always - with docker container:

```bash
 ### SYSTEM:   manufacturer: ; model: Docker Container; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12
 ### VERSIONS: MagicMirror: 2.27.0-develop; electron: 27.2.0; kernel: 6.1.19-v8+; node: 18.17.1; npm: 10.2.4; pm2: ; docker:
```
> is there CPU architecture info

I was searching for this too and it is available:

```json
  "os": {
    "platform": "linux",
    "distro": "Debian GNU/Linux",
    "release": "12",
    "codename": "bookworm",
    "kernel": "6.1.19-v8+",
    "arch": "arm",
    "hostname": "raspberrypi",
    "fqdn": "raspberrypi",
    "codepage": "UTF-8",
    "logofile": "debian",
    "serial": "85b5a61dd83b459d884a321797727254",
    "build": "",
    "servicepack": "",
    "uefi": false
  },
```

khassel:
> armv6l/armv7l/aarch64, ......

tested now 64-bit on raspi so we are getting only the values `arm`, `arm64` and `x64` for architecture, but should be added anyway.

khassel:
found one additional info I'm interested in: `timezoneName`

KristjanESPERANTO:
>> why not using process.versions.electron instead of ...

> I have tested it. At the point where I have set the function, electron is not yet initialized ...

I made a mistake. The variable works! I've changed it accordingly, it's much better now. Thanks for the hint!

> Here the results on my windows machine using docker container with wsl2:

Strange. Actually, I would have expected an extra line with Docker info.

`if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;`

So this line seems to have no effect at all.

> is there CPU architecture info

I've just added it.

> found one additional info I'm interested in: timezoneName

I've just added it.

Current status on my system:

```sh
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.10.0-20-amd64
 ### VERSIONS: electron: 27.2.0; node: 18.17.1; npm: 10.2.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ### OTHER:    timeZone: Europe/Berlin
 ```

**Comment by khassel** [comment]
> Strange. Actually, I would have expected an extra line with docker info.

> if (typeof staticData["dockerInfo"] !== "undefined") systemDataString += `\n ### DOCKER:   containers: ${staticData["dockerInfo"]["containers"]}; operatingSystem: ${staticData["dockerInfo"]["operatingSystem"]}; osType: ${staticData["versions"]["osType"]}; architecture: ${staticData["versions"]["architecture"]}; serverVersion: ${staticData["versions"]["serverVersion"]}`;

> So this line seems to have no effect at all.

I think this line can be removed. For getting such infos you must have installed docker. With that you get such info's but I think this is useless here. I'm only interested if mm is running as docker container and this is displayed in the model section `### SYSTEM:   manufacturer: Microsoft; model: Docker Container; raspberry: undefined; virtual: true`

**Comment by KristjanESPERANTO** [comment]
> I think this line can be removed. For getting such infos you must have installed docker. With that you get such info's but I think this is useless here. ...

Okay, I've just removed it :+1:

I think we have now reached a status where we can merge :slightly_smiling_face:

**Comment by KristjanESPERANTO** [comment]
Thanks to everyone who contributed! 😀

**Comment by bugsounet** [comment]
I just see That from ma dev deskyop:

```sh
[05.01.2024 19:47.24.506] [INFO]  System information:
 ### SYSTEM:   manufacturer: FUJITSU; model: ESPRIMO P420; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Ubuntu; release: 23.10; arch: x64; kernel: 6.5.0-14-generic
 ### VERSIONS: electron: 27.2.0; node: 18.17.1; npm: 10.2.0; pm2: 5.3.0; docker: 
 ### OTHER:    timeZone: Europe/Paris
```
`node: 18.17.1` is related to internal node version of electron using

and not the installed node version
 ```sh
bugsounet@Kubuntu:~/MagicMirror/modules$ node -v
v21.5.0
```

This is my sysinfo from my MMM-GoogleAssistant Website:

image

So, maybe checking installed node version is needed no ? with something like [this](https://github.com/bugsounet/MMM-GoogleAssistant/blob/prod/components/systemInformation.js#L85)

**Comment by khassel** [comment]
For me it is o.k. because

- when running `npm start` you see the above info's with electron version and node version (from electron)
- when running `npm run server` you see no electron version und version of installed node

But we can add the 2 node versions when running `npm start` if you need it.

**Comment by sdetweil** [comment]
the installed node version is where we see all the fetch  problems in the node_helper.  so that would be more useful than the  electron version

**Comment by khassel** [comment]
as I said, we can add the missing one ...

But which node version is used?

- running `npm run server` is clear, the locally installed node version is used
- running `npm start` I think the locally installed node version is **not** used (only npm as wrapper for the start), instead the electron binary is started and AFAIK the node version coming with electron is used

But correct me if I'm wrong.

**Comment by khassel** [comment]
From https://www.electronjs.org/docs/latest/tutorial/tutorial-prerequisites#nodejs-and-npm

grafik

**Comment by bugsounet** [comment]
but i think that node version is used in node_helper (I will check)

**Comment by sdetweil** [comment]
but this is not true.  

the node modules are loaded before electron is launched.

and the node_helpers fail because fetch is not there. if we upgrade the base  then things are ok.

so node helpers run under platform node

**Comment by khassel** [comment]
I added an additional line to `js/node_helper.js`

```js
        start () {
                Log.log(`Starting module helper: ${this.name}`);
                Log.log(process.versions);
        },
```

and this is the result when starting with `npm start`:

```bash
[05.01.2024 23:17.31.742] [LOG]   Server started ...
[05.01.2024 23:17.31.742] [LOG]   Connecting socket for: updatenotification
[05.01.2024 23:17.31.742] [LOG]   Starting module helper: updatenotification
[05.01.2024 23:17.31.743] [LOG]   {
  node: '18.17.1',
  acorn: '8.8.2',
  ada: '2.5.0',
  ares: '1.19.1',
  brotli: '1.0.9',
  cldr: '43.0',
  icu: '73.1',
  llhttp: '6.0.11',
  modules: '118',
  napi: '9',
  nghttp2: '1.52.0',
  openssl: '1.1.1',
  simdutf: '3.2.12',
  tz: '2023c',
  undici: '5.22.1',
  unicode: '15.0',
  uv: '1.44.2',
  uvwasi: '0.0.18',
  v8: '11.8.172.18-electron.0',
  zlib: '1.2.13.1-motley',
  electron: '27.2.1',
  chrome: '118.0.5993.159'
}
[05.01.2024 23:17.31.743] [LOG]   Connecting socket for: calendar
[05.01.2024 23:17.31.743] [LOG]   Starting node helper for: calendar
[05.01.2024 23:17.31.743] [LOG]   Connecting socket for: newsfeed
```

sdetweil:
makes no sense

khassel:
my test or the result?

bugsounet:
Like I says installed node version must be ~ to electron node version for have all functions
after Electron use his own version of node

I know... there is no sense
In other hand... dependencies use installed node version

sdetweil:
>. dependencies use installed node version

right, but they are loaded by mode_helper
so IT must using installed node version.

what is that junk from start then?

server mode just doesn't launch electron . it does all the rest using installed node.

khassel:
> server mode just doesn't launch electron

yes, server mode always uses the installed node version, but not when running as electron (`npm start`).

If I put the line `Log.log(process.versions);` e.g. in the `node_helper.js` of the default calendar I get the same results as above.

KristjanESPERANTO:
I would not have expected anything like this. Even though it may not have any added value, that's just how I looked at it:

I added the following two lines to the new `logSystemInformation` function:

```js
Log.info("systeminformation:     " + staticData["versions"]["node"]);
Log.info("process.versions.node: " + process.versions.node);
```

Running with electron:
```bash
[06.01.2024 00:21.07.909] [INFO]  systeminformation:     18.17.1
[06.01.2024 00:21.07.909] [INFO]  process.versions.node: 18.17.1
```
Running as server:
```bash
[06.01.2024 00:21.13.738] [INFO]  systeminformation:     21.1.0
[06.01.2024 00:21.13.738] [INFO]  process.versions.node: 21.1.0
```

Getting the information from 'node -v' or reading the package.json in the node_modules directory are definitely options.  Of course we can also display both (the information we get from systeminformation and the installed version).

But to understand the whole system, it would be good if we could get to the bottom of it. Even if I think it's unlikely, but could it be that if we run with electron, everything runs with the node version of electron? That would be a very interesting discovery. And it would not only be relevant for what we want to display, but also for troubleshooting in general.

sdetweil:
but it's not correct.  

keep the current electron, install node 16 and the calendar node_helper will fail cause fetch isn't found

fetch is in electron. what node is being used by node helpers?

KristjanESPERANTO:
I just did that:

```bash
kristjan@debian:~/MagicMirror$ node -v
v16.20.0
kristjan@debian:~/MagicMirror$ npm run start

> magicmirror@2.27.0-develop start
> DISPLAY="${DISPLAY:=:0}" ./node_modules/.bin/electron js/electron.js

[06.01.2024 00:46.04.220] [LOG]   Starting MagicMirror: v2.27.0-develop
[06.01.2024 00:46.04.224] [LOG]   Loading config ...
[06.01.2024 00:46.04.226] [DEBUG] config template file not exists, no envsubst
....
[06.01.2024 00:46.06.083] [INFO]  System information:
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.10.0-20-amd64
 ### VERSIONS: electron: 27.2.0; node: 18.17.1; npm: 8.19.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ### OTHER:    timeZone: Europe/Berlin
[06.01.2024 00:46.06.083] [INFO]  systeminformation:     18.17.1
[06.01.2024 00:46.06.084] [INFO]  process.versions.node: 18.17.1
...
```

And the calendar and other modules using fetch are working 🤯

sdetweil:
yes, mind blown...

bugsounet:
really... I do not understand anything anymore

KristjanESPERANTO:
I give up trying to understand that today! 😴

khassel:
I managed to get mm running with electron but without even nodejs/npm installed ...

KristjanESPERANTO:
> I managed to get mm running with electron but without even nodejs/npm installed ...

Crazy. Just to get it right: You have executed `npm run start` without having npm installed on the system?

bugsounet:
and `npm i` works too??

sdetweil:
@bugsounet  he probably had to do the install ., then remove node and npm and then launch the electron bin directly

i remember others having to do npm rebuild on onoff for server mode,  or electron-rebuild for electron mode
but then fetch makes my head hurt

khassel:
> then remove node and npm and then launch the electron bin directly

this was the first approach

Electron releases prebuilt binaries as described [here](https://www.electronjs.org/docs/latest/tutorial/application-distribution#with-prebuilt-binaries).

So I did
- a normal `npm install` with `node` an `npm`, I omitted devDependencies and optional dependencies (electron) and saved the mm folder
- used a debian image (without `node` an `npm`) and installed the debian packages which electron needs
- installed the electron prebuilt binaries
- copied the mm folder
- started mm with `./electron path-of-mm-folder`

Here is the complete `Dockerfile`, tested this on x64-arch with debian-wsl2:

```Dockerfile
FROM node:21-alpine as builder

WORKDIR /opt/magic_mirror

RUN set -e; \
    apk add --no-cache git;

RUN set -e; \
    node -v; \
    git clone --depth 1 -b develop --single-branch "https://github.com/MagicMirrorOrg/MagicMirror.git" .; \
    git log -1; \
    npmargs="--no-audit --no-fund --no-update-notifier --arch=arm64 --omit=optional --omit=dev"; \
    rm -f package-lock.json; \
    echo "now executing: npm install ${npmargs}"; \
    npm install ${npmargs}; \
    cat package.json; \
    sed -i "s:address\: \"localhost\":address\: \"0.0.0.0\":" config/config.js.sample; \
    sed -i "s:ipWhitelist\: \[.*\],:ipWhitelist\: \[\],:" config/config.js.sample; \
    cp config/config.js.sample config/config.js;

FROM debian:bookworm
LABEL maintainer="Karsten Hassel"

RUN set -e; \
    apt-get update; \
    DEBIAN_FRONTEND=noninteractive apt-get -qy --no-install-recommends install git nano sudo wget ca-certificates libgtk-3-0 libx11-xcb-dev libnss3-dev \
      libxss1 libxtst6 libasound2 libdrm2 libgbm1 libxshmfence1 fonts-arphic-uming procps unzip; \
    apt-get clean; \
    rm -rf /var/lib/apt/lists/*; \
    useradd -u 1000 node; \
    usermod -a -G video node; \
    echo "node ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers; \
    groupadd --gid 997 gpio; \
    usermod -a -G gpio node;

ENV ELECTRON_DISABLE_SANDBOX=1 \
    DBUS_SESSION_BUS_ADDRESS="unix:path=/var/run/dbus/system_bus_socket"

COPY --from=builder --chown=node:node /opt/magic_mirror /opt/magic_mirror

USER node

WORKDIR /home/node/electron

RUN set -e; \
    wget -O /tmp/electron.zip https://github.com/electron/electron/releases/download/v28.1.1/electron-v28.1.1-linux-x64.zip; \
    unzip /tmp/electron.zip -d /home/node/electron/; \
    rm -f /tmp/electron.zip;

ENTRYPOINT ["/home/node/electron/electron", "/opt/magic_mirror"]
```

KristjanESPERANTO:
So what do we learn from all this?

1. When running with electron, we only need npm and the locally installed node version is irrelevant?
2. The locally installed version is only relevant for server mode?
3. The new system information in the log shows always the relevant node version: In electron mode the node version of electron and in server mode the locally installed version. So we don't have to change anything?

khassel:
yes, I think so

- to 1.: `npm` needs `node` to run, so for running `npm start` you need `node`. The command `./node_modules/.bin/electron` (used when running `npm start`) also needs `node` because this is only a small nodejs program which calls the `electron` binary. You can take a look into `node_modules/electron` folder. But if I remove the `node` binary I can still start mm with `node_modules/electron/dist/electron js/electron.js`
- to 2.: For running mm yes, but you need it e.g. for `npm install` (for mm and modules)
- to 3.: I'm fine with current state, but if others want this information, no problem

bugsounet:
@KristjanESPERANTO: maybe add an Log if GPU accelerator is activated or not?
Because I just see that is disabled by default. I turn around "my MM² works slowly* since last release :/

sdetweil:
yes  we changed the default to disabled to stop the errors

export ELECTRON_ENABLE_GPU=1

To turn it on

bugsounet:
In my case: error/warn vs MM² running slowly ... `ELECTRON_ENABLE_GPU=1` win :) :)

KristjanESPERANTO:
> maybe add an Log if GPU accelerator is activated or not?

Sure. Like that?

Hardware acceleration enabled:

```bash
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.10.0-20-amd64
 ### VERSIONS: electron: 27.2.0; node: 18.17.1; npm: 10.2.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ### OTHER:    timeZone: Europe/Berlin; ELECTRON_ENABLE_GPU: 1
``` 

Hardware acceleration not enabled: 

```bash
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.10.0-20-amd64
 ### VERSIONS: electron: 27.2.0; node: 18.17.1; npm: 10.2.4; pm2: 5.3.0; docker: 20.10.24+dfsg1
 ### OTHER:    timeZone: Europe/Berlin; ELECTRON_ENABLE_GPU: undefined
 ```
 
 
 @khassel Is it even relevant to display the docker version? If mm is running in a container, there is no docker installed in it and then no version is displayed, right?

KristjanESPERANTO:
Before I do a PR tomorrow, you are welcome to give me feedback here.

Changes so far:
- Add ELECTRON_ENABLE_GPU
- Remove docker version (depending on response from @khassel)
- Differentiation between installed and used node version

```bash
 ### SYSTEM:   manufacturer: Notebook; model: N650DU; raspberry: undefined; virtual: false
 ### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.10.0-20-amd64
 ### VERSIONS: electron: 27.2.0; used node: 18.17.1; installed node: 21.1.0; npm: 10.2.4; pm2: 5.3.0
 ### OTHER:    timeZone: Europe/Berlin; ELECTRON_ENABLE_GPU: 1
 ```

khassel:
Thanks, I'm fine with these changes

KristjanESPERANTO:
Okay, thanks for the feedback!

Further discussion on the subject of system information in the new PR #3337.

sdetweil:
looks good to me..

PR #3332, opened 2024-01-03
updatenotification: update_helper.js recode with pm2 library (v2.27.x)

bugsounet:
#3285

Because there is so many conflit with package,
I have rewrite the code with v2.27.0-develop

For remember:

 * recode: `update_helper.js` with `pm2` library
 * fix: default config -> `updates` is a array
 * delete: `command-exists` library (not used)
 * delete: `PM2_GetList()`  function (not used)
 * add: check `updates.length` (prevent crash)
 * add: `[PM2]` tag in log (for better visibility)
 * add: `pm2` library
 
advantage:
  * we use the pm2 library directly
  * avoids weird returns from child_process.exec when requesting a json format from pm2
  * simplified the code

inconvenient:
  * we have vulnerabilities with axios

240120 Fix:
 * use `pm2_env.pm_cwd` instead of `pm2_env.PWD` : prevent using `pm2 restart <id> --update-env` in other directory (for enable GPU rendering for exemple)
 * resolve packages (again)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,9 @@
 {
 	"extends": ["eslint:recommended", "plugin:@stylistic/all-extends", "plugin:import/recommended", "plugin:jest/recommended", "plugin:jsdoc/recommended"],
-	"plugins": [],
+	"plugins": ["unicorn"],
 	"env": {
 		"browser": true,
-		"es2023": true,
+		"es2024": true,
 		"jest/globals": true,
 		"node": true
 	},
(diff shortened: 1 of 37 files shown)


bugsounet:
Testing and correct some new part of code (new dependency format)

@rejas should i remove the exception of dependency ?

rejas:
I actually like the allow-ghsas solution which can stay there as long as the dependabot vulanribility is there.

bugsounet:
if it's ok for you, it's ok for me ;)

rejas (approved):

PR #3336, opened 2024-01-07
[newsfeed] Suppress unsightly animation edge cases when there are 0 or 1 active news items

crazyscot:
When the newsfeed module has an items list of size 1, every `updateInterval` the animation runs to transition from the active story to itself. This is unsightly. This PR suppresses that.

To reproduce: configure newsfeed with a single news source, `ignoreOldItems` true, a short `updateInterval` (e.g. 3000), and a carefully-chosen small `ignoreOlderThan` lining up with the current contents of your news source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ _This release is scheduled to be released on 2024-04-01._
 ### Fixed
 
 - Skip changelog requirement when running tests for dependency updates (#3320)
+- [newsfeed] Suppress unsightly animation cases when there are 0 or 1 active news items
+- [newsfeed] Always compute the feed item URL using the same helper function
 
 ### Deleted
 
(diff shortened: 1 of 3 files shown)


rejas:
Thanks for the PR. Quick question though: if the newsfeed starts with 1 item only, and has more items only a while later, this will still update as expected?

crazyscot:
node_helper sends a `NEWS_ITEMS` notify, this calls generateFeed which updates this.newsItems.
Next updateInterval, `newsItems.length > 1`, no problem.

However I think I have spotted an edge case: if there is still precisely one item, but a _different_ item, the display won't be updated.

An obvious workaround is to forcibly redraw the DOM after calling generateFeed(), but I think that will cause the same unsightliness, alebit just at a lower rate.
There may be a better workaround by special-case cacheing the currently displayed item, but it feels messy?

rejas:
> However I think I have spotted an edge case: if there is still precisely one item, but a _different_ item, the display won't be updated.

how about storing a hash value of the single item and compare that against the next?

crazyscot:
I have thought this through a bit more and realised there was a further animation to suppress: the case where there are 0 active items. See the detailed comment in `newsfeed.js`.
- Branch rebased onto the tip of `develop`.
- The branch also includes two minor bugfix commits in `getTemplateData`; they weren't affecting me, but they jumped out when I was working there.

crazyscot:
Fixed linter issues.

rejas (approved):

PR #3340, opened 2024-01-08
Use node prefix for build-in modules

KristjanESPERANTO:
It is basically a cosmetic thing, but has the following advantages:

1. Consistency with the official node documentation. The prefix is used there.
2. It is easier to recognize the build-in modules.


rejas:
> Unfortunately, I have not found an ESLint rule for this.

I think this one is what you are looking for: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md

KristjanESPERANTO:
Okay, nice. I will check this out and add it if it fits. This way we will have a nice complete commit.

Unicorn has other good rules, so maybe we can activate some of them later.

KristjanESPERANTO:
Great, I added the rule and it even found a place that I had overlooked. I love linter 😀

I also tested activating the recommended rules from unicorn, but that would have result in a lot of changes. That's why I didn't integrate it into this PR.

rejas (approved):

Issue #3291, opened 2023-12-13
one calendar test fails since today


khassel:
```bash
  ● Calendar module › exdate check › should show the recurring event 51 times (excluded once) in a 364-day (inclusive) period

    expect(received).resolves.toBe()

    Received promise rejected instead of resolved
    Rejected to value: [Error: expect(received).toHaveLength(expected)·
    Expected length: 51
    Received length: 50
[...]

PR #3342, opened 2024-01-09
Better fixes for #3291 and the underlying exdate issues

jkriegshauser:
* Worked around several issues in the RRULE library that were causing deleted calender events to still show, some initial and recurring events to not show, and some event times to be off an hour. (#3291)
* Renamed variables in *calendarfetcherutils.js* to be more clear about use of `moment` and js's `Date` class.
* Added calendar config option `forceUseCurrentTime` (default:`false`) which will ignore overridden `Date.now` in the config in order to keep some tests consistent.
* Added several unit tests for crossing DST in different timezones with excluded events.


rejas:
so, whats this PR about actually ;-) ?

jkriegshauser:
> so, whats this PR about actually ;-) ?

It’s a WIP for better fixes for #3291 and the underlying exdate issues. I’m still seeing breakthrough events that have been excluded showing up. Stay tuned…

sdetweil:
@jkriegshauser  note that we changed rrule processors this release, and are seeing weird results. skipped dates (null returned where date should be)

jkriegshauser:
> @jkriegshauser note that we changed rrule processors this release,(dragged in from node-ical) and are seeing weird results. skipped dates (null returned where date should be)

Odd, where is the null date being returned?

sdetweil:
from the rrule.between  in calendarfetcherutils.js
```
					Log.debug(`Search for recurring events between: ${pastLocal} and ${futureLocal}`);
					let dates = rule.between(pastLocal, futureLocal, true, limitFunction);   // <---- here 
					Log.debug(`Title: ${event.summary}, with dates: ${JSON.stringify(dates)}`); // you can see them in the debug output here 
					dates = dates.filter((d) => {      // this filter added in this release to remove null entries which cause crash
						if (JSON.stringify(d) === "null") return false;
						else return true;
					});
```

from a user with problem 

1704763166785-screenshot_20240108_191819_chrome-resized

jkriegshauser:
> from a user with problem

@sdetweil thanks, can you provide a link to the forum comment? The URL in the image doesn't seem to take me anywhere.

sdetweil:
@jkriegshauser 
see 
https://forum.magicmirror.builders/topic/18363/calendar-module-missing-recurring-events-schedule-after-7pm-est

jkriegshauser:
@rejas @sdetweil @khassel this is ready for review now. Sam, I kept the `null` filtering that you mentioned above, and I believe this will also address the problems mentioned in the forum post above.

KristjanESPERANTO:
Thank you for your work! In case we need to look at the PR again later: Can you write in the initial comment what was changed? I think the text from the changelog would suffice.

jkriegshauser:
Ah I was just working on this.

khassel:
> Ah I was just working on this.

Sorry, you can update again ...

I'm fine with this PR, can this be merged? @sdetweil @rejas

jkriegshauser:
Alright, should be gtg now

jkriegshauser:
@rejas @sdetweil Can this be merged? I've been running it on two local instances for a week and had no issues. @khassel gave approval above

rejas:
Sorry for delaying this longer than needed. Thanks for your work (and the insane amount of tests ;-)
Now only some doc updates are needed ;-)

jkriegshauser:
Yep, planning on finishing those up when I get some time 😄

Issue #3345, opened 2024-01-11
Precipitation Probability not displayed when it is 0%


vppencilsharpener:
OS: Ubuntu 22.04.3 LTS
Node: v21.5.0
MagicMirror Version: 2.26.0

In the hourly weather display, when the the Precipitation Probability is 0 the table cell is blank/empty. For consistency I would expect it to display “0%”

**Steps to Reproduce:**
I am using the WeatherGov module, but I believe this would be reproducible on any module that uses the weather.js library. To reproduce display the precipitation probability in the hourly weather module and have a dry time period where the precipitation probably is returned as 0%.

**Expected Results:** 
To be consistent in the hourly chart, I would expect the table cell to be populated with 0% instead of blank. That gives confidence that data is being returned instead of it being an unknown (i.e. is it zero or is it missing data).

**Additional Notes:**
I think the problem is with modules\default\weather\weather.js line 238.
If the type is “precip” and the value is 0 or the formated value is “0.00”, then a blank value is returned.
I don’t believe the zero check to be valid for the Precipitation Probability data and instead would send zero values to the WeatherUtils.convertPrecipitationUnit function. There does not appear to be anything in that function that cannot handle a zero value so it should not cause other problems. 
Null values should continue to return a blank/empty value as they do now.

I've made the change locally and will submit a pull request shortly.

PR #3346, opened 2024-01-11
Develop

vppencilsharpener:
Fixes issue #3345. 

I think I submitted this correctly, but don't do this often so let me know if anything needs to be corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ _This release is scheduled to be released on 2024-04-01._
 ### Fixed
 
 - Skip changelog requirement when running tests for dependency updates (#3320)
+- Display precipitation probability when it is 0% instead of blank/empty (#3345)
 
 ### Deleted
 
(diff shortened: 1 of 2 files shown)


rejas:
Sorry for not reviewing earlier... Might it be possible that you write a test for this?

rejas (approved):

rejas:
> > Sorry for not reviewing earlier... Might it be possible that you write a test for this?
> 
> I am willing, but in all honesty I have never written an automated test so there is going to be a bit of a learning curve. If I could be directed to something existing I can base it off of that would be appreciated.

You can take a look at `tests/e2e/modules/weather_hourly_spec.js`
There are tests for checking the precipitation is displayed. Your test should check if the 0 is displayed if the pop in the mocked data `tests/mocks/weather_hourly.json` is 0

rejas:
Merging this so the next release has it. If you find the time @vppencilsharpener you can always add a test later ;-)

apoulalier:
Thanks for your job but since this change, the amount of precipitaton (quantity of water) seems to be 0.00 mm and for me it's not relevant. When your return null, the column could disappaear if we have a week without rain !
Right now, i see a value for all days. Can we make a revert or a better fix ?

PR #3361, opened 2024-01-17
[newsfeed] Fix bug where the newsfeed sometimes stops

crazyscot:
It appears that #3336 introduced a bug where a newsfeed with >1 items would stop updating after a while (usually after `activeItem` wraps around the end of the list). Sorry! My bad, I hadn't tested that case well enough.


crazyscot:
I presume the test failure is related to the ongoing work on 3291.

KristjanESPERANTO:
> I presume the test failure is related to the ongoing work on 3291.

Yes, the test fails usually on wednesdays.

khassel (approved):

rejas (approved):

PR #3370, opened 2024-01-27
Update husky and let lint-staged fix ESLint issues

KristjanESPERANTO:
The new version of husky makes it possible to simplify the pre-commit hook a little.

And since prettier no longer takes care of the JavaScript files in our project, it can no longer come into conflict with ESLint while running lint-staged. Therefore we can activate the correction of ESLint issues here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
 		"lint:js": "eslint . --fix",
 		"lint:css": "stylelint 'css/main.css' 'fonts/*.css' 'modules/default/**/*.css' 'vendor/*.css' --config .stylelintrc.json --fix",
 		"lint:staged": "lint-staged",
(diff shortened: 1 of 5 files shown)


khassel (approved):

package.json
```diff
@@ -88,7 +88,7 @@
 	},
 	"lint-staged": {
 		"*": "prettier --write",
-		"*.js": "eslint",
+		"*.js": "eslint --fix",
 		"*.css": "stylelint"
```
rejas, line 92: Stylelint can/should fix too:

```suggestion
		"*.css": "stylelint --fix"
```

rejas (changes requested):

package.json
KristjanESPERANTO: Okay, I've just changed it :slightly_smiling_face:

rejas (approved):

PR #3386, opened 2024-02-24
update electron to v29 and other deps

khassel:


rejas (approved):

bugsounet:
Hi,

Are you really sure that is a good idea to use electron v29.x ?
Do you test some electron rebuild test ? (onoff, node-pty, usb, drivelist, ...)

khassel:
we have time to test until next release and can revert if someone find problems.

At the moment we have some problems with internal fetch under electron, I hope this will be better with electron v29 because they upgraded from node v18 to v20.

bugsounet:
In this case, you are right, 3rd will sort itself out in due time :)

PR #3392, opened 2024-03-02
Add electron-rebuild to suite test

bugsounet:
because actually i'm not able to rebuild any libraries to works with electron v29.x
I write a suite test to check `electron-rebuild`

Note:  works with [v28.x](https://github.com/MagicMirrorOrg/MagicMirror/actions/runs/8122468177/job/22201931385)


bugsounet:
force to use electron v28 result:

```
Run npx electron-rebuild
- Searching dependency tree
✔ Rebuild Complete
```

electron v29 failed

rejas:
Hi @bugsounet so what do we get with this new test? what benefit do we have?

rejas (commented):
Hi @bugsounet so what do we get with this new test? what benefit do we have? just trying to understand

bugsounet:
Some complex modules use electron rebuild
perhaps in line with a new electron version, it might be important to know if our development version works before publishing the release
I don't... it's the job of 3rd party module
but it can help them

It's up to you if you want to add it or not.
In my case I already have a test suite to see if it is compatible or not with the master branch of MM²

sdetweil:
many modules use libraries which have binary components  tied to node and electron versions 

our update causes them to fail, they need to rebuild.

we should ship a compatible version of electron rebuild for this purpose. it must be located in the same node_modules folder as the electron install now

khassel:
> electron v29 failed

```bash
node@80ef79387f2d:/opt/magic_mirror$ python3 --version
Python 3.11.2
node@80ef79387f2d:/opt/magic_mirror$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
node@80ef79387f2d:/opt/magic_mirror$ npx electron --version
v29.1.5
node@80ef79387f2d:/opt/magic_mirror$ npx electron-rebuild --version
Electron Rebuild Version: 3.6.0
node@80ef79387f2d:/opt/magic_mirror$ node -v
v21.7.1
node@80ef79387f2d:/opt/magic_mirror$ npx electron-rebuild
✔ Rebuild Complete
```

bugsounet:
@khassel: yes it's solved since last week ;)

khassel:
> Hi @bugsounet so what do we get with this new test? what benefit do we have? just trying to understand

@rejas this should be merged. we need such a test to see if electron-rebuild fails if we update electron

bugsounet:
After, I don't know if we can do this test only for information

If test failed -> just warn but we can merge

sdetweil:
is electron/rebuild in package.json  for production?

bugsounet:
> is electron/rebuild in package.json for production?

@sdetweil : Actually no, but you want to add it ?
@rejas, @khassel : what do you think ?

khassel:
> @rejas, @khassel : what do you think ?

I like the idea to test electron-rebuild against the electron version but I don't want to have electron-rebuild in `package.json`.

I think only a minority of users are using such hardware related modules so we should not install this for all. Next point is, that electron-rebuild needs python3 and build-essential rpm's to work, but these are not installed on every os per default, e.g. not in the docker images (which would increase size of them massively). So we should stay with current approach where these modules must install electron-rebuild if needed.

bugsounet:
@khassel: that why i install it only for testing in test-suite ;)

khassel (approved):

KristjanESPERANTO (approved):

rejas (approved):

PR #3396, opened 2024-03-07
Change multiday fullDay Event behaviour

Paranoid93:
Hey!

This PR should change the behaviour of starting fullDay events that last several days. The goal was to change the behavior of the "Starting today, ends T" (T=Tomorrow) event, so it should show how many days it will occur from the first day on

Before situation:

a fullDay event that started 'today' and ends several days later showed Today on the first day. The rest of the days it showed X days left.

grafik

Y => Yesterday
T => Tomorrow

Target situation with this commit:
a fullDay event that started 'today' shows 'X days left' from the first day on and 'Today' on the last day.

grafik


rejas:
Thanks for the PR. Makes sense for me, but the tests are failing for it :-(

Paranoid93:
Okay.. they were green before and as far as I see, nothing changed in the Code. Would you be fine that I edit the tests to only assert that the text "ends in" is contained, no matter how many days are following? That's the main goal of the PR.

rejas (approved):

PR #3401, opened 2024-03-22
move suncalc dependency out of dev, as it is used by the clock module

sdetweil:
user reported problem with clock module, checking code found dependency on suncalc library, but it is not loaded in production mode..  move dependency


rejas (changes requested):
shouldnt the lock file also get updated by this?

sdetweil:
it is referenced in the lock file. and
I don't think it matters   as one of the release steps is to rebuild it

rejas (approved):

PR #3406, opened 2024-03-27
update deps and package.json's (eslint)

khassel:


package.json
```diff
@@ -28,7 +28,10 @@
 		"lint:staged": "lint-staged",
 		"prepare": "[ -f node_modules/.bin/husky ] && husky || echo no husky installed."
 	},
-	"repository": "MagicMirrorOrg/MagicMirror",
+	"repository": {
```
rejas, line 31: this reverts https://github.com/MagicMirrorOrg/MagicMirror/commit/b0161fe0115acd0f2cfed4abde69120e260490ac#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

what format do we want to use in the future?

khassel: I only updated `"eslint-plugin-package-json": "^0.12.0"` and after this the js test failed.
After this I used the lint run and this made these changes.

It's more your topic so I'm open for changes ...

rejas (approved):

PR #3407, opened 2024-03-28
MM² Icon

bugsounet:
* Create `MM²` icon
 * Allow to change default electron icon to this icon


MM².png
rejas, line 1: nice idea. but can you name the file jsut mm2.png without the high 2 ? dont want to risk the spelling to condfuse some filesystems :-)

rejas (changes requested):

KristjanESPERANTO:
Out of curiosity: When and where is the icon displayed?

bugsounet:
@rejas: wait, I will change "MM" color because it's so darkness... we see nothing lol

bugsounet:
@KristjanESPERANTO 

something like this ?

image

and task bar (when minified or when multi-screen using)
image

bugsounet:
@veeck, it's ok (i think)

rejas (approved):

khassel:
so this is the new favicon when running under electron (which is only visible if not running fullscreen), but for server only we have another favicon in `index.html`. So should we use the same icon for both?

rejas:
that index.html favicon is this base64 encoded small string?

khassel:
I think so, displayed is 
grafik

khassel:
you can replace this line e.g. with

```html
<link rel="icon" type="image/png" href="path/file.png">
```

to test other icons (the path must served by mm)

bugsounet:
will not works `mm2.png` have no entry on express

khassel:
that was meant with "the path must served by mm", for testing is must be copied to e.g. config or modules or ...

bugsounet:
add an  `app.get` to serve `mm2.png` in server.js ;)
I'm in lazy mode :D

rejas:
unifying them would make sense. and maybe also ask in the forum if we want to make a "design the mm2 logo" contest (no offence against your design, bugsounet, but we are developers, no designers :-)

bugsounet:
sure it's an first approach... I have planned to do that in next version of MM²

PR #3415, opened 2024-04-01
Fix check for mastermerge label

rejas:
Mastermerge label wasnt checked correctly, this PR should hopefully fixes it for good


khassel:
@rejas ready to merge?

rejas:
Would like to wait until release is done

rejas:
or we merge this and re-build the release?

rejas:
@khassel updated it with some dep updates so I think we can merge it before we do the real 2.27.0 release

PR #3425, opened 2024-04-18
add wrapper ClassName // DOM_CONTENT_UPDATED

bugsounet:
Added:
 * ~~`DOM_CONTENT_UPDATED` notification when a module is updated with `updateDom()`~~ to do better
 * set className for njk wrapper: 
     * There is no class name to the main wrapper of the module and It's dificult to identify it
     * ---> define it to `<module_name>-njk`

Idea:
  Add a callback option in `updateDom()`

```
  if (typeof updateOptions?.options?.callback === "function") {
    updateOptions.options.callback();
    //module.notificationReceived("DOM_CONTENT_UPDATED",null,null))
  }
```

can be used with:
 ```
this.updateDom({options: { speed: 2, callback: () => this.checkInactive()}});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ _This release is scheduled to be released on 2024-07-01._
 
 ### Added
 
+- wrapper ClassName when using njk
+- `DOM_CONTENT_UPDATED` notification when updateDom() complete
+
 ### Removed
 
 ### Updated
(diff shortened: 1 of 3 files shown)


sdetweil:
jeez that sounds like a lot of waste.. notifications on every dom update.. WHY?  what is the use case?

If YOU need to know put a change event on the node

sdetweil:
see the Observer and  Mutation events
https://stackoverflow.com/questions/3219758/detect-changes-in-the-dom

bugsounet:
really ? ... and with njk ?

PR #3426, opened 2024-04-18
check Active (something displayed)

bugsounet:
Added:
 * check if module is active (avoid unwanted `<div></div>` and module spacing)

this need #3425, should I merge with it and make one PR only ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ _This release is scheduled to be released on 2024-07-01._
 
 ### Added
 
+- [updatenotification] check if module is active (avoid unwanted <div></div> and module spacing)
+
 ### Removed
 
 ### Updated
--- a/modules/default/updatenotification/updatenotification.css
+++ b/modules/default/updatenotification/updatenotification.css
@@ -1,3 +1,7 @@
 .module.updatenotification a.difflink {
   text-decoration: none;
 }
+
+.module.updatenotification .inactive {
+  display: none;
+}
--- a/modules/default/updatenotification/updatenotification.js
+++ b/modules/default/updatenotification/updatenotification.js
@@ -38,6 +38,9 @@ Module.register("updatenotification", {
 				this.sendSocketNotification("CONFIG", this.config);
 				this.sendSocketNotification("MODULES", Object.keys(Module.definitions));
 				break;
+			case "DOM_CONTENT_UPDATED":
+				this.checkActive();
+				break;
 			case "SCAN_UPDATES":
 				this.sendSocketNotification("SCAN_UPDATES");
 				break;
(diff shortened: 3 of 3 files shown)


sdetweil:
what do you mean active?   not hidden?    

not active mean display:none;

modules are ALWAYS 'active' , just not displayed..
could be working in the background..
with display:none the content doesn't go away.

if a module is not loaded, then it is not there at all.

bugsounet:
inactive -> nothing displayed ;)
active -> updatenotification display something

Why i add this:
 * If you set a background in this module, you have a strange result 

like this:

image

in top bar: it's the result of `<div></div>` and nothing displayed of `updatenotification`

sdetweil:
what module?

bugsounet:
I don't understand, `what module ?` sorry :/

bugsounet:
euh ... this module is `updatenotification` ;)

sdetweil:
how would updatenotification get a background, it's not coded for one in css or njk

bugsounet:
after.... maybe i'm crazy ... but if you apply  a padding to updatenotification with some effect
you have "color band"

image

image

css:
```
.module-content > *:not(.inactive):not(.hidden) {
    padding: 5px;
    border-radius: 10px;
    box-shadow: -5px -5px 8px #000;
    background: rgba(0, 0, 0, 15%);
    backdrop-filter: blur(5px);
    border: 1px solid rgba(255, 255, 255, 10%);
}
```

bugsounet:
I think, that i'm not ONLY in the same case or ... I have to return to code learning school :) :) :)

image

and ... no "color bar"

sdetweil:
what is the content normally, not hidden, just  the empty div from the njk?

bugsounet:
yes it's the `<div></div>` (empty) when nothing is displayed

bugsounet:
the better way (for me) is add an `display: none` style when nothing is displayed

BUT:
 * updateDom() is not async... so you can't await for add/del an "inactive" class
 * updateDom() have no callback when done

So... 
 * maybe create an callback option with `DOM_CONTENT_UPDATED` notification

something like:
```
updateDom({ options : { callback: true } })
```
  * or again call a function
```
updateDom({ options : { callback: myfonction() } })
```

sdetweil:
there is no getDom() with templated modules

it's getTemplateData()

see weather.js

bugsounet:
already tested.
 -> you can't play with the main njk div content
 -> njk result is inside njk main div content -> `<div> ... njk decoded ... </div>`

Note that why I have added [this](https://github.com/MagicMirrorOrg/MagicMirror/pull/3425/files#diff-50260f11109f55b944da641c51de244dffd718cbd77f7e5ce7b9124ab18fe265R80)

sdetweil:
but why do you need that?  module is over that and you  can get the div below module-content, easy w jquery. and just as easy in css file.  

as we ha e t had a y reports of un showing when hidden.. what's the problem u are trying to solve?

bugsounet:
I was simply trying not to display an empty div which causes a color band when using css `padding` on this div
Only when the module displays the following result `<div></div>`
I report the module name: `updatenotification`

now if you can determine it with css: well done
adding jQuery to `updatenotification` a joke?

In all case, you have right sam, it's noted : no report = no change

I just added `disabled: true` in config and magical there is no color band.

Can I close #3425 too then (because one needs the other) ?
Because no `updateDom()` callback is needed and no `njk classname` div is require

bugsounet:
more seriously, it's solved with this CSS pointer

```
.module-content > *:not(:empty) {
 <...>
}
```

sdetweil:
just for clarity and accuracy, that is called the selector 

of  course it needs the module name class prefix so.it doesn't apply to all .modules

bugsounet:
No no, there is no report about this, so i made no change
It's works since many year like that

PR #3429, opened 2024-04-26
Added Specific ignoreOlderThan value (override) per URL Feature

AartiPhatke:
Hello,
I added the feature request in the newsfeed module for being able to set ignoreOlderThan for each newsfeed rather than have a global variable for it, see: https://github.com/MagicMirrorOrg/MagicMirror/issues/3360

To do this however, instead of adding an ignoreOlderThanOveride parameter, I changed ignoreOldItems and ignoreOlderThan to be arrays so they can hold values for each news feed. I also changed the logic for how these variables are accessed in generateFeed(). I also ran 'npm run test' and some of the unit tests for the calendar module were failing, I don't think this is related to my PR but I wanted to mention it. 
Hope this helps!

### DIFF
--- a/modules/default/newsfeed/newsfeed.js
+++ b/modules/default/newsfeed/newsfeed.js
@@ -24,8 +24,10 @@ Module.register("newsfeed", {
 		updateInterval: 10 * 1000,
 		animationSpeed: 2.5 * 1000,
 		maxNewsItems: 0, // 0 for unlimited
-		ignoreOldItems: false,
-		ignoreOlderThan: 24 * 60 * 60 * 1000, // 1 day
+		// Lets make this an array, so it holds the value for each newsfeed, currently there is only one: New York Times
+		ignoreOldItems: Array.from({length: feeds.length}, () => false),
+		// Let's make this an array, so it holds the values for each newsfeed, currently there is only one: New York Times 
+		ignoreOlderThan:Array.from({length: feeds.length}, () => 24 * 60 * 60 * 1000), // all feeds are set to 1 day 
 		removeStartTags: "",
 		removeEndTags: "",
 		startTags: [],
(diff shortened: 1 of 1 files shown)


sdetweil:
it would be easier to add the variables for each h feed and then use a function that returned the feed value or the default if not specified 

and use that function instead of the direct value access in the genersteFeed function

AartiPhatke:
Hello!
Yes I can make this change and make another PR request

modules/default/newsfeed/newsfeed.js
```diff
@@ -5,7 +5,9 @@ Module.register("newsfeed", {
 			{
 				title: "New York Times",
 				url: "https://rss.nytimes.com/services/xml/rss/nyt/HomePage.xml",
-				encoding: "UTF-8" //ISO-8859-1
+				encoding: "UTF-8", //ISO-8859-1
+				ignoreOldItems: false,
```
AartiPhatke, line 9: Rather than having ignorOldItems and ignorOlderThan be global parameters, I added them inside the feed array and gave then default values of false and 1day.

modules/default/newsfeed/newsfeed.js
```diff
@@ -176,19 +180,23 @@ Module.register("newsfeed", {
 			});
 		}
 	},
-
+	getFeedProperty(feed, property) {
```
AartiPhatke, line 183: I added this function so you can access values inside the objects in the feed array.

modules/default/newsfeed/newsfeed.js
```diff
@@ -176,19 +180,23 @@ Module.register("newsfeed", {
...
-		for (let feed in feeds) {
-			const feedItems = feeds[feed];
-			if (this.subscribedToFeed(feed)) {
+		for (let i = 0; i < feeds.length; i++) {
+			const feedItems = feeds[i];
+			if (this.subscribedToFeed(feedItems)) {
 				for (let item of feedItems) {
-					item.sourceTitle = this.titleForFeed(feed);
-					if (!(this.config.ignoreOldItems && Date.now() - new Date(item.pubdate) > this.config.ignoreOlderThan)) {
+					let ignoreOldItems = getFeedProperty(feedItems, 'ignoreOldItems');
```
AartiPhatke, line 196: I changed how the the variables are accessed when trying to display the feed by using the getFeedProperty()

rejas:
hi @AartiPhatke thanks for your patience. Could you adress the styling issues that github shows in the files tab? And also update your branch to the latest develop so the tests run through again?

khassel:
Due to missing push permissions, I couldn't reuse this PR. Therefore, I'm closing it in favor of https://github.com/MagicMirrorOrg/MagicMirror/pull/3742.

PR #3450, opened 2024-05-14
Check config before starting MM

KristjanESPERANTO:
I think it might be a good idea to check the config at every start.


khassel:
We implemented the possibility to use a `config.js.template` instead of a `config.js` (where the `config.js` is created while startup).

At the moment the check fails if there is no `config.js`:

```bash
[2024-05-14 21:51:17.549] [ERROR] File not found: /opt/magic_mirror/config/config.js
```

So if we want this check the code must be changed that no error is thrown if no `config.js` exists but a `config.js.template`.
If we want to check `config.js.template` in such a case I don't now if the check can handle Variables `${XY}` which can be present in `config.js.template`. There is also a env variable `MM_CONFIG_FILE` which is used to set another config file location. I don't know if the check respects this variable (it should).

Other point: Any idea why the test fails?

bugsounet:
Why not hard coding this check in main core of MM² before loading config ?

sdetweil:
in general I am in favor of not starting if there is a fatal config error . 

 not like now where we start with a built in dummy config, unlike whatever the user expected.

KristjanESPERANTO:
I'd like to take a closer look, but it looks like I will not find the time soon.

KristjanESPERANTO:
> I think loadConfig in app.js must be executed before the check because in this function the config.js is created when using config.js.template.

The config check is now called from inside the `loadConfig` after handling the config file. So I think it should now cover all use cases:

1. normal `config.js`
2. `MM_CONFIG_FILE` and
3. `config.js.template`


> in general I am in favor of not starting if there is a fatal config error

Now if the config check fails, MM will not start:

```bash
kristjan@debian:~/MagicMirror$ npm run start

> magicmirror@2.29.0-develop start
> DISPLAY="${DISPLAY:=:0}" ./node_modules/.bin/electron js/electron.js

[2024-09-18 03:31:25.980] [LOG]   Starting MagicMirror: v2.29.0-develop 
[2024-09-18 03:31:25.999] [LOG]   Loading config ... 
[2024-09-18 03:31:26.001] [LOG]   config template file not exists, no envsubst 
[2024-09-18 03:31:26.319] [INFO]  Checking file...  /home/kristjan/MagicMirror/config/config.js 
[2024-09-18 03:31:26.343] [ERROR] Your configuration file contains syntax errors :( 
[2024-09-18 03:31:26.343] [ERROR] Line 193 column 3: Parsing error: Unexpected token { 
kristjan@debian:~/MagicMirror$ 
```

khassel:
@KristjanESPERANTO can you please solve the merge conflicts, thanks!

khassel (approved):

PR #3457, opened 2024-06-02
Update dependencies

rejas:
Not eslint v9 yet, but I am looking at it :-)


khassel:
see https://github.com/import-js/eslint-plugin-import/issues/2948

khassel:
I created a new PR #3460 with newer dependencies and without "@stylistic/eslint-plugin": "^2.1.0".

This plugin is for eslint v9, so you can update the both together ...

PR #3470, opened 2024-06-17
Update compliments.js

### LINKED ISSUE #3465: Compliments module special days config
I asked this on the forums and was advised to log a bug.

**Platform**: RaspberryPi 5 running raspberryPi OS. Current version of MagicMirror software.

When adding config to enable the "special days" feature of the default compliments module it will seemingly add that days comment to the existing array. In my case that array is considerable so special messages such as "Happy Birthday" or "Merry Christmas" etc. are lost in a sea of compliments.

My expectation of this feature would be that on the configured day it shows only the compliments specified in that area of the config. For example:

`"....-12-25": ["Happy Christmas!"]`

I would expect the only compliment on the 25th December to be "Happy Christmas!". That is not the case though. I am currently adding a line of code in the compliments.js file to empty the array on a special day and only include the configured comments:

`// Add compliments for special days`
`for (let entry in this.config.compliments) {`
`  if (new RegExp(entry).test(date)) {`
`    compliments = []; // Empty the array if it's a special day in config.js`
`    Array.prototype.push.apply(compliments, this.config.compliments[entry]);`
`  }`
`}`

This is not necessarily a "bug" as it may well be the intended behaviour, however I would suggest making this clear in the docs and, if I may be so bold, adding a new config option that allows this behaviour if the user wants it. I am not confident in my coding or GitHub skills to suggest such a change but it seems like it could be quite simple.

WallysWellies:
`Fixes #3465`

Add config option `specialDayUnique` that replaces the compliments array with only the compliments configured for that day. Default setting is `true`. Can be disabled by setting the option to `false` which will add the days compliments to the existing array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [2.27.0] - 2024-04-01
 
-Thanks to: @bugsounet, @crazyscot, @illimarkangur, @jkriegshauser, @khassel, @KristjanESPERANTO, @Paranoid93, @rejas, @sdetweil and @vppencilsharpener.
+Thanks to: @bugsounet, @crazyscot, @illimarkangur, @jkriegshauser, @khassel, @KristjanESPERANTO, @Paranoid93, @rejas, @sdetweil, @vppencilsharpener and @WallysWellies.
 
 This release marks the first release without Michael Teeuw (@michmich). A very special thanks to him for creating MagicMirror and leading the project for so many years.
 
@@ -20,6 +20,7 @@ For more info, please read the following post: [A New Chapter for MagicMirror: T
 - [weather] `showHumidity` config is now a string describing where to show this element. Supported values: "wind", "temp", "feelslike", "below", "none". (#3330)
 - electron-rebuild test suite for electron and 3rd party modules compatibility (#3392)
 - Create MM² icon and attach it to electron process (#3407)
+- [compliments] Added `specialDayUnique` config option which defaults to "true", replacing the existing compliments array with only the compliments that have been configured for that day (#3465)
 
 ### Updated
 
--- a/modules/default/compliments/compliments.js
+++ b/modules/default/compliments/compliments.js
@@ -15,7 +15,8 @@ Module.register("compliments", {
 		morningEndTime: 12,
 		afternoonStartTime: 12,
 		afternoonEndTime: 17,
-		random: true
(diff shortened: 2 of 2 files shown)


sdetweil:
I don't see the changelog file in your commit

WallysWellies:
I saw the message about updating the changelog but I kind of assumed that this change would be included in a future release and the changelog would be managed by the project team. I'm not entirely clear what I am expected to do - should I add a 2.27.1 section ("backwards compatible bug fix" seems appropriate after reading the [versioning notes](https://semver.org/)) with a note of my changes?

WallysWellies:
Ok I think that should be it done now. I added my name after checking to see if other issues all have attribution so I hope that's appropriate.

WallysWellies:
Updated the js file to use tabs for indentation rather than spaces to resolve testing issues.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [2.27.0] - 2024-04-01
 
-Thanks to: @bugsounet, @crazyscot, @illimarkangur, @jkriegshauser, @khassel, @KristjanESPERANTO, @Paranoid93, @rejas, @sdetweil and @vppencilsharpener.
+Thanks to: @bugsounet, @crazyscot, @illimarkangur, @jkriegshauser, @khassel, @KristjanESPERANTO, @Paranoid93, @rejas, @sdetweil, @vppencilsharpener and @WallysWellies.
```
rejas, line 10: Seems you put the changes in the section of the last release, please move them to the top in the Unreleased section

modules/default/compliments/compliments.js
```diff
@@ -15,7 +15,8 @@ Module.register("compliments", {
 		morningEndTime: 12,
 		afternoonStartTime: 12,
 		afternoonEndTime: 17,
-		random: true
+		random: true,
+		specialDayUnique: true
```
rejas, line 19: The default should be false IMO to keep the current behaviour

rejas (changes requested):

WallysWellies:
I think I maybe messed this up from the start by working on the master branch rather than the develop branch. Is it easiest for me to delete this PR and start again? The changes are minimal and I can incorporate the adjustments made by @rejas at the same time (default to `false`).

rejas:
You could rebase your PR if you are comfortable with that, but I don't mind if you start from scratch

PR #3476, opened 2024-06-19
update `config.js.sample` to use openmeteo as weather provider

khassel:
which needs no api key.

I think this is a better choice than the old one because new users which use this config as starting point will now see weather data instead of `loading...`


khassel:
@rejas why have we `[chore]` in the changelog instead of `[core]`?

I can change this with this PR before merging if it is o.k. for you.

rejas (approved):

PR #3483, opened 2024-06-26
Use latest@version of node for automated-tests.yaml

bugsounet:
Maybe it's a good idea to use latest@node version for automated-tests

Actually it's an "random" version

Note: it's already used [there in electronRebuild](https://github.com/MagicMirrorOrg/MagicMirror/blob/develop/.github/workflows/electronRebuild.yaml#L19)


bugsounet:
@khassel : can you rerun failed job ?
`code: 'ETIMEDOUT',` on fetch

khassel (approved):

Issue #3480, opened 2024-06-24
[updatenotification] use of this module creates pm2 daemon process(es)


khassel:
When running mm with updatenotification module (or running tests where this module is active) we get

```bash
node@ec51cfe176dd:/opt/magic_mirror$ npx jest tests/e2e/env_spec.js
 PASS   e2e  tests/e2e/env_spec.js (7.22 s)
  App environment
    ✓ get request from http://localhost:8080 should return 200 (217 ms)
    ✓ get request from http://localhost:8080/nothing should return 404 (27 ms)
    ✓ should show the title MagicMirror² (204 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        7.456 s
Ran all test suites matching /tests\/e2e\/env_spec.js/i.

node@ec51cfe176dd:/opt/magic_mirror$ ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
node         1  0.0  0.0   2472   884 pts/0    Ss   22:52   0:00 /usr/bin/tini -- ./entrypoint.sh bash
node         7  0.0  0.0   4188  3384 pts/0    S    22:52   0:00 bash
node       193  3.6  0.8 1039160 66032 ?       Ssl  22:53   0:00 PM2 v5.4.1: God Daemon (/home/node/.pm2)
node       200  3.1  0.7 1006164 64880 ?       Ssl  22:53   0:00 PM2 v5.4.1: God Daemon (/home/node/.pm2)
node       211  0.0  0.0   8088  3868 pts/0    R+   22:53   0:00 ps aux
```

@bugsounet  I'm not familiar with pm2, so my question is: Can we avoid these `God Daemon` processes?

bugsounet:
sure it's not really logical to have 2 `pm2` process running

I will inspect

bugsounet:
```sh
bugsounet@PC-Bugsounet:~/MagicMirror/modules$ npx jest tests/e2e/env_spec.js && grep pm2
 PASS   e2e  ../tests/e2e/env_spec.js
  App environment
    ✓ get request from http://localhost:8080 should return 200 (37 ms)
    ✓ get request from http://localhost:8080/nothing should return 404 (14 ms)
    ✓ should show the title MagicMirror² (201 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
[...]

PR #3484, opened 2024-06-26
updatenotification: avoid using pm2 when running in docker container

khassel:
related to #3480 

Change module updatenotification so that it can work without `pm2` in a docker container.

It now tests if file `/.dockerenv` exists and if so `require("pm2")` is never called.


khassel:
@bugsounet please check if this change is o.k. for you, thanks.

bugsounet:
@khassel : just test this PR and it's ok

rejas (approved):

PR #3526, opened 2024-08-29
[weather] add humidity to hourly view, fix spacing error when using UV Index, add config option to hide precipitation entries that are zero

OWL4C:
Basically the title. Just some existing weather data included into hourly, some config option ("hideZeroes") to hide precipitation when it is zero (this actually shrinks the entire table, removing columns that are completely empty), and add a spacing column to fix the UV Index column.

I don't understand why CHANGELOG.md is broken, i accidentally did my changes to my master, so i just took the current dev branch changelog and copied it into my master changelog, added my edits, and commited. This shouldn't conflict but it does


sdetweil:
automerge is turned off.. 
you have a conflict in the CHANGELOG.md

OWL4C:
I hope this fixed it. It may have broken my fork but that is irrelevant, the changes proposed seem fine.

rejas:
Thanks for your PR. Could you attach screenshot sof before and after your changes to this PR here?

OWL4C:
Currently:
Screenshot 2024-08-30 at 13 07 35
All functions as mentioned:
Screenshot 2024-08-30 at 13 06 54

However, the need for a spacer is purely the result of a missing css class. I have thus modified the css to include a new (placeholder) class humidity-hourly, as the humidity in the current weather would also be affected by the padding.
```
.weather .precipitation-amount,
.weather .precipitation-prob,
.weather .humidity-hourly,
.weather .uv-index {
  padding-left: 20px;
  padding-right: 0;
}
```
What is the naming convention for these classes? As soon as i know what to name the placeholder i will commit a change to remove the spacer from hourly and add the new css class.

OWL4C:
I just used the placerholder as the actual class name.

modules/default/weather/hourly.njk
```diff
@@ -21,15 +21,27 @@
                         {% endif %}
                     </td>
                 {% endif %}
+                {% if config.showHumidity %}
```
rejas, line 24: please add the new config options to the defaults object at the top of the file

modules/default/weather/hourly.njk
```diff
@@ -21,15 +21,27 @@
...
+                  <td class="align-left bright humidity-hourly">
+                          {{ hour.humidity }}
+                          <i class="wi wi-humidity humidity-icon"></i>
+                  </td>
+                {% endif %}
                 {% if config.showPrecipitationAmount %}
-                    <td class="align-right bright precipitation-amount">
-                        {{ hour.precipitationAmount | unit("precip", hour.precipitationUnits) }}
-                    </td>
+                  {% if config.hideZeroes and hour.precipitationAmount==0 %}
```
rejas, line 31: this empty if seems hard to read.
would prefer a if (!config.hideZeroes or hour.precipitationAmount > 0)

modules/default/weather/hourly.njk
```diff
@@ -21,15 +21,27 @@
...
+                  {% else %}
+                      <td class="align-right bright precipitation-amount">
+                          {{ hour.precipitationAmount | unit("precip", hour.precipitationUnits) }}
+                      </td>
+                  {% endif %}
                 {% endif %}
                 {% if config.showPrecipitationProbability %}
+                  {% if config.hideZeroes and hour.precipitationAmount==0 %}
```
rejas, line 39: same here with the empty if

rejas (changes requested):
sorry for taking so long to review, just some minor things to finish

modules/default/weather/hourly.njk
OWL4C: Do you mean the defaults in the weather.js?

sdetweil: yes, it is a new config property, so it needs to havea default value set, and the defaults:{} object contains those 
as documented in the developers guide for modules

khassel: no progress here, @rejas is the missing `config.showHumidity` the only thing (beside changelog merge conflicts) which is preventing this from merging?

rejas: besides that i had one question above...

modules/default/weather/hourly.njk
```diff
@@ -8,7 +8,7 @@
                 {% if config.fade %}style="opacity: {{ currentStep | opacity(numSteps) }};"{% endif %}>
                 <td class="day">{{ hour.date | formatTime }}</td>
                 <td class="bright weather-icon">
-                    <span class="wi weathericon wi-{{ hour.weatherType }}"></span>
+                    <i class="wi weathericon wi-{{ hour.weatherType }}"></i>
```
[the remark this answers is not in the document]
OWL4C: No idea, apparently i thought icons should be in i but they shouldn't...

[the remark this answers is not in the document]
rejas: The idea is good, but I am not sure if people rely on the span tag for styling...

Maybe cleaning the icons (not just these here) up in another PR where we can discuss that would be the better approach?

modules/default/weather/hourly.njk
OWL4C: showHumidity is a string due to current.njk implementing multiple options. I only ever implemented a binary True/False. Since js' Boolean("none") == True this will not work as intended without changes

OWL4C: I changed the if to check against "none", but other than that it is still undefined what the correct string should be to turn it on, since the current.njk Humidity has multiple placements, which is inapplicable here. Since the current config comment isn't listing possible values right now, i would suggest changing it to something along the lines of ` // current: "none", "wind", "below"; hourly: "none", "on" `. While this wouldn't be clean it would work.
Alternatively it could be split into "ShowHumidity: True/False" + "CurrentHumidityPosition: wind, below" etc; whatever is preferable. Changing the comment is a band aid but works, adding options could become convoluted but has less ambiguity.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ _This release is scheduled to be released on 2024-10-01._
 
 ### Added
 
+- [weather] Added option Humidity to hourly View
```
rejas, line 14: you'll need to move your changelog entries into the latest develop section....

rejas (changes requested):

PR #3527, opened 2024-08-30
update dependencies

khassel:
We cannot upgrade to electron v32 because electron-rebuild is failing with epoll, so staying at v31.


rejas (approved):

bugsounet:
I will inspect

```
/home/bugsounet/.electron-gyp/32.0.1/include/node/v8config.h:13:2: error: #error "C++20 or later required."
   13 | #error "C++20 or later required."
```

PR #3530, opened 2024-09-01
add new env vars MM_MODULES_DIR and MM_CUSTOMCSS_FILE …

khassel:
… for setting these things from outside (and overriding corresponding config.js properties `config.foreignModulesDir` and `customCss`)

- remove elements from index.html when loading script or stylesheet files fails
- removed `config.paths.vendor` (could never work because `vendor` is hardcoded in `index.html`) and renamed `config.paths.modules` to `config.foreignModulesDir`. The `config.paths. ...` properties were implemented in the initial commit in `js/defaults.js` but were never functional.
- fixes `app.js` which didn't respect `config.paths.modules` before
- as `modules/defaults` is directly set in many places in the source code restrict `config.paths.modules` to foreign modules (it has never worked for default modules), now renamed to `config.foreignModulesDir`
- adds new `/env` section in `server.js` for getting the new env vars in the browser
- fixes TODO in `server.js` so test directories are now only published when running tests

These changes allow changing some main paths from outside mm with the new env vars. You now **can** put all user stuff into one directory, e.g. the `config` dir:
- `config.js` as before
- `custom.css`
- foreign modules

This would simplify other setups e.g. the docker setup. At the moment we have to deal with 3 directories where 2 of them (`modules` and `css`) contains mixed stuff, which means mm owned files and user files. This can now simplified and leads to cleaner setups (if wanted).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,8 @@ _This release is scheduled to be released on 2024-10-01._
 ### Added
 
 - [core] Add spelling check (cspell): `npm run test:spelling` and handle spelling issues
-- [core] removed `config.paths.vendor` (could not work because `vendor` is hardcoded in `index.html`), renamed `config.paths.modules` to `config.foreignModulesDir`
-- [core] add variable `MM_CUSTOMCSS_FILE` which - if set - overrides `config.customCss`
-- [core] add variable `MM_MODULES_DIR` which - if set - overrides `config.foreignModulesDir`
+- [core] removed `config.paths.vendor` (could not work because `vendor` is hardcoded in `index.html`), renamed `config.paths.modules` to `config.foreignModulesDir`, added variable `MM_CUSTOMCSS_FILE` which - if set - overrides `config.customCss`, added variable `MM_MODULES_DIR` which - if set - overrides `config.foreignModulesDir`
+- [core] elements are now removed from index.html when loading script or stylesheet files fails
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ _This release is scheduled to be released on 2024-10-01._
 ### Added
 
 - [core] Add spelling check (cspell): `npm run test:spelling` and handle spelling issues
+- [core] removed `config.paths.vendor` (could not work because `vendor` is hardcoded in `index.html`), renamed `config.paths.modules` to `config.foreignModulesDir`
 - [core] add variable `MM_CUSTOMCSS_FILE` which - if set - overrides `config.customCss`
-- [core] add variable `MM_MODULES_DIR` which - if set - overrides `config.paths.modules` for foreign modules (not default modules)
+- [core] add variable `MM_MODULES_DIR` which - if set - overrides `config.foreignModulesDir`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ _This release is scheduled to be released on 2024-10-01._
 ### Added
 
 - [core] Add spelling check (cspell): `npm run test:spelling` and handle spelling issues
+- [core] add variable `MM_CUSTOMCSS_FILE` which - if set - overrides `config.customCss`
+- [core] add variable `MM_MODULES_DIR` which - if set - overrides `config.paths.modules` for foreign modules (not default modules)
 
 ### Removed
 
--- a/js/app.js
+++ b/js/app.js
@@ -9,6 +9,7 @@ const Log = require("logger");
 const Server = require(`${__dirname}/server`);
 const Utils = require(`${__dirname}/utils`);
 const defaultModules = require(`${__dirname}/../modules/default/defaultmodules`);
+const { getEnvVarsAsObj } = require(`${__dirname}/server_functions`);
 
 // Get version number.
 global.version = require(`${__dirname}/../package.json`).version;
@@ -159,7 +160,8 @@ function App () {
 	function loadModule (module) {
 		const elements = module.split("/");
 		const moduleName = elements[elements.length - 1];
-		let moduleFolder = `${__dirname}/../modules/${module}`;
+		const env = getEnvVarsAsObj();
+		let moduleFolder = `${__dirname}/../${env.modulesDir}/${module}`;
 
 		if (defaultModules.includes(moduleName)) {
 			moduleFolder = `${__dirname}/../modules/default/${module}`;
--- a/js/defaults.js
+++ b/js/defaults.js
@@ -19,6 +19,7 @@ const defaults = {
 	units: "metric",
 	zoom: 1,
 	customCss: "css/custom.css",
+	foreignModulesDir: "modules",
 	// httpHeaders used by helmet, see https://helmetjs.github.io/. You can add other/more object values by overriding this in config.js,
 	// e.g. you need to add `frameguard: false` for embedding MagicMirror in another website, see https://github.com/MagicMirrorOrg/MagicMirror/issues/2847
 	httpHeaders: { contentSecurityPolicy: false, crossOriginOpenerPolicy: false, crossOriginEmbedderPolicy: false, crossOriginResourcePolicy: false, originAgentCluster: false },
@@ -72,12 +73,7 @@ const defaults = {
 				text: "www.michaelteeuw.nl"
 			}
 		}
-	],
-
-	paths: {
-		modules: "modules",
-		vendor: "vendor"
-	}
+	]
 };
 
 /*************** DO NOT EDIT THE LINE BELOW ***************/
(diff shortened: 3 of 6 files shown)


sdetweil:
hm.   this will break modules which depend on their location being two directories up from the root and others that depend on folder names.
 like those that need electron-rebuild in the root node_modules folder.  and my mmm-config

will also break my upgrade script

what will be the effect on git pull if the deployed  structure is different?

khassel:
the idea would be 
- to have `modules/default/...` (no change)
- to have `config/MMM-XY` instead of `modules/MMM-XY`
- to have `config/custom.css` instead of `css/custom.css` (which is already now possible by setting in `config.js`)

So can you explain what exactly would break?

> what will be the effect on git pull if the deployed structure is different?

```bash
# Ignore users config file but keep the sample.
/config/*
!/config/config.js.sample
```

no effect because above is in `.gitignore`

khassel:
you can already now change the `modules` dir and `custom.css` location inside `config.js`, this PR only introduces additional env vars for setting this from outside.

sdetweil:
can either/both of these be outside the MagicMirror folder tree?

khassel:
> can either/both of these be outside the MagicMirror folder tree?

I did not test this so far, will do and report.

I think at least `modules` folder must reachable from the browser over `https://<server>:<port>/<modules-dir>` so the choice of the folder is limited.

khassel:
both variables (or corresponding values in `config.js`) are restricted to the published directories in `js/server.js`:

```bash
			let directories = ["/config", "/css", "/fonts", "/modules", "/vendor", "/translations", "/env"];
```

sdetweil:
ok, so one could set modules to config, but not new dir foo, same for config to modules, but not foo

and does it require leading slash?

khassel:
> ok, so one could set modules to config, but not new dir foo, same for config to modules, but not foo

yes.

> and does it require leading slash?

no.

You can define it like

```bash
node@81ac3b4687ab:/opt/magic_mirror$ export MM_CUSTOMCSS_FILE=config/custom.css
node@81ac3b4687ab:/opt/magic_mirror$ export MM_CUSTOMCSS_FILE=/config/custom.css
node@81ac3b4687ab:/opt/magic_mirror$ export MM_CUSTOMCSS_FILE=/opt/magic_mirror/config/custom.css
```

All 3 variants works, the mm root dir is stripped in the code.

sdetweil:
ok,  can cssfile be modules/foo.css

i misread cssfile, thinking config file and folder could be moved

sdetweil:
i always assumed cssfile was name of file, not location

khassel:
The default `config.js` key `customCss` is `"css/custom.css"`. You can overwrite it in `config.js` or with the new env var `MM_CUSTOMCSS_FILE`. `modules/foo.css` works.

sdetweil:
thanks, i never thought one could change the css directory, only the filename

sdetweil:
I haven't checked the code yet,  in all cases at the time  a module is loaded are the config variables set  so I can use those as the true path? vs examining for the env myself..

sdetweil:
in MMM-Config I launch a shell script to collect all the module defaults and generate the form object.. 
I have a bug now, in that I don't check for where the config.js is.. I  'assume' config/config.js and add on modules in modules 
easy fix if the env variables are replicated into the config object

khassel:
at the moment this is still unmerged ...

> I haven't checked the code yet, in all cases at the time a module is loaded are the config variables set so I can use those as the true path? vs examining for the env myself..

> if the env variables are replicated into the config object

The new variables don't touch `config.js`. I don't know where you need the new variables, server or browser?
Depending on this 
- for server you can use `getEnvVarsAsObj` from `server_functions.js`
- for browser you can do something like I implemented in `loader.js`
```bash
		const res = await fetch(`${location.protocol}//${location.host}/env`);
		return JSON.parse(await res.text());
```

> I don't check for where the config.js is.. I 'assume' config/config.js

this is unrelated to this PR but you should check `MM_CONFIG_FILE`

sdetweil:
the config object  after parsing is available to the browser sideof the module, so I could see the config.modules.path I was just wondering  if you used the env to update that (as your PR comments say 'override' )

I don't really need the config object, and I don't care about css really. 
the environment variables get passed to the node_helper in the process.env 
so I can see these vars..

khassel:
> the config object after parsing is available to the browser sideof the module, so I could see the config.modules.path I was just wondering if you used the env to update that (as your PR comments say 'override' )

thanks, didn't thought about this, but I think it would be the better way to change the values in the config object. Will check this and change it if it works.

khassel:
> the config object after parsing is available to the browser sideof the module

the config object is available to the browser because of these lines in `index.html`:

```html
    <script type="text/javascript" src="js/defaults.js"></script>
    <script type="text/javascript" src="#CONFIG_FILE#"></script>
```

(the `#CONFIG_FILE#` is replaced with real `config.js` file/path while loading)

So I stay with the current approach because otherwise I had to manipulate content of `config.js` or do other ugly things.

sdetweil:
understood. i forgot they are loaded from file

khassel:
@rejas from my side this is ready for review

khassel:
> did you change anything in the prior support config.paths.modules?

no

but there was a bug in the code so this couldn't work (because `modules` was hardcoded).

So what are you testing?
- current `develop` branch
- or content of this branch?

sdetweil:
testing 2.29 prior to this PR.. 

>but there was a bug in the code so this couldn't work (because modules was hardcoded).

so, YES, you had to change something to get this PR to work . the original code was broken

trying to test setting modules location thru config.js, before env variable..
(testing my MMM-Config module which needs to know where the modules are..and the config.js name)

khassel:
> so, YES, you had to change something to get this PR to work . the original code was broken

line 162 in `js/app.js`

```js
		let moduleFolder = `${__dirname}/../modules/${module}`;
```

was not using `config.paths.modules` but hardcoded `modules`

khassel:
well, I don't know exactly what you tested, AFAIS you are not testing this PR (the new env vars) but the config.js functionality (which was already there before this PR).

> the paths object doesn't have the same override approach as the rest of config.js
leave out the vendor: setting amd the url loaded says undefined/node_modules....

yes, had this problem too, you must provide a full object under `paths`, otherwise the not provided keys are missing. Maybe objects are copied only on first level. AFAIS the property `config.paths.vendor` is useless (will not work with a value <> `vendor`) because we have `vendor` hardcoded in `index.html`. So we could "fix" this by removing `config.paths.vendor` and replacing `config.paths.modules` with e.g. `config.modulesPath`.

Let me know if I should update this PR with such a fix.

> even tho the modules js wasn't found, it gets injected anyhow..

sorry, don't understand/don't know what you expect from me here

sdetweil:
I tested this pr as a separate branch..

moved one module to the config folder from modules
add the paths:{ modules:"config"} to the config.js 

then started MM 

and looked in the browser development console, elements tab and see this 
Screenshot at 2024-09-11 13-17-52

notice the 3 undefined/node_modules 
for things in vendor

you confirmed its not handle the same way (Object .assign()) like the other config.js parms.. 
and I don't seen any other vendor/  hard coded in config.js 
so it would have worked wiith the object.assign

second, on startup, the config.js references modules NOT found in the  configured 'modules' folder, 
AND correctly reports that the modulename.js cannot be found during startup (the messages above) 

BUT it still considers them 'loaded' as their script link is still added to the in browser index.html.. 
that should not be.. 

I understand you are 'only' proposing env vars, BUT.. the previous (undocumented)  built in support
didn't work, so you have had to fix it.. now you 'own' it for this PR. else this PR would not work either

using the env var MM_MODULES_DIR=config
(config.js paths commented out)  produces the same results for issue 2. 

Screenshot at 2024-09-11 13-28-10

I'm ok with not having vendor movable.. and having only the config.modulesPath

sdetweil:
altho.. I am sure that on all file systems we run on, there is a link capacbility for those that want to move things around physically without us having to provide config options..

khassel:
@sdetweil fixed the things you mentioned above.

Tests are failing at the moment unrelated to this PR (did the tests on `develop` with same result).

sdetweil:
crazy that we get these random failures...

khassel:
> altho.. I am sure that on all file systems we run on, there is a link capacbility for those that want to move things around physically without us having to provide config options..

the basic idea is to use one dir (`config`) for all user stuff so I have only to mount one volume for the docker setup. I could use `ln` in the container to get things in one dir but there would still be the need to mount 3 volumes. Other point could be (not officially supported) windows where are no symlinks out of the box AFAIK.

sdetweil:
windows 11 has symlinks 
https://www.howtogeek.com/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/

sdetweil:
looks good both env and config parm

khassel:
I'm not willing to solve merge conflicts in the changelog over weeks/months, so can I have feedback from @rejas / @sdetweil if this can be merged?

sdetweil:
I have tested it extensively, and I think it works well.. so yes, you can merge

rejas (approved):

PR #3531, opened 2024-09-11
Ignore positions.js

KristjanESPERANTO:
This file is generated when MM is started. As I understand it, it should not be included in the repository.

Should probably have been part of #3518.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ Temporary Items
 *.orig
 *.rej
 *.bak
+
+# Ignore positions file (#3518)
+js/positions.js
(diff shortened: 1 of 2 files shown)


khassel:
Tests are failing at the moment unrelated to this PR ...

sdetweil:
I didn't think I included it..

khassel:
> I didn't think I included it..

you did not include it, you forgot to exclude it from git ...

.gitignore
```diff
@@ -81,3 +81,6 @@ Temporary Items
 *.orig
 *.rej
 *.bak
+
+# Ignore positions file (#3518)
+positions.js
```
khassel, line 86: can you please change to `js/positions.js` because there is another file with the same name in this repo `tests/configs/modules/positions.js` ...

KristjanESPERANTO: Oh, sorry. I just fixed it.

PR #3545, opened 2024-09-18
Review config_check.js

KristjanESPERANTO:
Only details changes. No functional changes.

- remove superfluous colors in Log.error
- invert negative if
- update ESLint env
- use camel case variable name
- optimize Log strings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,9 @@ _This release is scheduled to be released on 2024-10-01._
 ### Added
 
 - [core] Check config at every start of MagicMirror² (#3450)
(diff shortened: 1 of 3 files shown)


rejas (approved):

sdetweil:
why is this the only acceptable answer?
Log.info(`Checking config file ${configFileName} ...`);

I don't like the lack of visibility of the variable use

KristjanESPERANTO:
> why is this the only acceptable answer?

Sorry, I don't understand this question. What do you mean by answer?

> I don't like the lack of visibility of the variable use

What editor do you use? I think even in editors without syntax highlighting the variables in template literals are well recognizable by the `${}.

PR #3535, opened 2024-09-13
Added `DOM_OBJECTS_UPDATED` notification when the DOM is re-rendered via `updateDom`

ryan-d-williams:
- [x] Base your pull requests against the `develop` branch.
- [x] Include these infos in the description:
> - Does the pull request solve a **related** issue?
Yes - solves #3534

> - If so, can you reference the issue like this `Fixes #<issue_number>`?
Fixes #3534 (also mentioned in commit message)

> - What does the pull request accomplish? Use a list if needed.

> > - Adds a new notification (`DOM_OBJECTS_UPDATED`) when the DOM is updated via `updateDom`

- [x] Please run `npm run lint:prettier` before submitting
- [x] Don't forget to add an entry about your changes to the CHANGELOG.md file.

More info can be found in #3534, but as a summary:

The `updateDom` function is not synchronous - there is an undetermined amount of time between when it completes and when the DOM has actually been re-rendered and is ready for interaction. The existing notification (`MODULE_DOM_CREATED`) only fires once on the initial DOM render. This PR solves the issue of subsequent re-renders by adding a new notification that fires whenever the DOM is ready after an update. This notification falls within expected lifecycle notifications (very similar to other libraries that provide DOM lifecycle notifications).

PR #3547, opened 2024-09-19
Updated to new notification name `DOM_OBJECTS_UPDATED` -> `MODULE_DOM_UPDATED`

### LINKED ISSUE #3534: Notification on subsequent DOM updates (not just the first)
Based on the [documentation](https://docs.magicmirror.builders/development/notifications.html) and by taking a look at the source code, right now the `MODULE_DOM_CREATED` notification is fired when the module is first loaded. Subsequent calls to `updateDom` do not fire any event.

I think it would be very beneficial to add a notification for each time the DOM is updated. I am currently accomplishing this in a vary hacky way, but it would be nice to add a notification to know exactly when the DOM update is complete so the module can interact with the new DOM if necessary.

My proposal is to add `MODULE_DOM_UPDATED` as a notification that fires when calls to `updateDom` are made. 

The same code that is used to send the `MODULE_DOM_CREATED` notification is also used during `updateDom`, so the code changes will be minimal.

I'm happy to submit a PR for this change, but wanted to discuss it here first.

ryan-d-williams:
This is an update to #3535. See #3534 for discussion and context. Fixes #3534 (again).


ryan-d-williams:
Set incorrect base branch, closing pull request.

sdetweil:
you can change it!

ryan-d-williams:
Oh sorry I didn't know that! I opened another one before I saw your comment. How do you change it? When I click edit at the top it only lets me edit the title.

PR #3543, opened 2024-09-16
add fix for sliceMultiDayEvents 

sdetweil:
sliceMultiDayEvents occasionally gets the number of events wrong and produces too many rows

Math.ceil() rounds up over 1.04 so we get an abnormal count

then the calcs for the midnight loop control used different moment()  functions, producing different results

fixes #3542

PR #3555, opened 2024-09-24
fix calendar showing previous day when using sliceMultiDayEvents

MarcLandis:
This bug is caused by #3543.

The calculation for midnight adds a day but for endDate we want the day to be subtracted again.


sdetweil:
do you have the test case that shows the prior failure

sdetweil:
ok our test system allows specific day, not today, so  set it to same day as testcase?

MarcLandis:
> ok our test system allows specific day, not today, so set it to same day as testcase?

by today I meant 24th of September :)

sdetweil:
i understand, i have to set the  testcase to always sept 24, 2024

MarcLandis:
For the "1 day repeat" event it shouldn't return it, but for the "2 days repeat" it should and the day in the past was shown too.

sdetweil:
your fix is good.. can u update the changelog

PR #3556, opened 2024-09-24
add tests for minimal node version (currently v20.9.0)

khassel:
Beside testing against node version `v20.x` and `v22.x` we should also test against the minimal node version, currently `v20.9.0`.

This is for seeing changes in dependencies which needs higher node version as with the July-24-release, where we wrote `node >=20` but shipped an `eslint` version which required `node>=20.9.0`.

### DIFF
--- a/.github/workflows/automated-tests.yaml
+++ b/.github/workflows/automated-tests.yaml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [20.x, 22.x]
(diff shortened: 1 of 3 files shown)


sdetweil:
my scripts are already at 20.9.0

khassel:
> my scripts are already at 20.9.0

yes, no change to last release, but I want to know when we have to change this number ...

sdetweil:
i think that is good too.

KristjanESPERANTO (approved):

rejas (approved):

PR #3551, opened 2024-09-22
Eslint

bugsounet:
testing with eslint v9.x

1st step: test:js
 * need to rewrite eslintrc to flat config

2st step: test with global test suite (not reviewed actually)

3st step: add again "eslint-plugin-import" (if possible)

PR #3558, opened 2024-09-24
Switch to ESLint v9 and flat config

KristjanESPERANTO:
Since PR #3551 was not yet complete, I made my own attempt.

1. Update to ESLint v9.
2. Replace deprecated `.eslintrc.json` and `.eslintignore` by flat config `eslint.config.mjs`.
3. Adapt `check_config.js` to use flat config.
4. Since `eslint-plugin-import` still doesn't support ESLint v9 I removed it. We can add it back when it does support v9.
5. Run tests `npm run check:js` and `npm run config:check`.
6. In order not to overload this PR, I have not yet activated more additional rules - there are some useful ones in the new plugin `@eslint/js`.

@bugsounet, please don't take it as an offence that I have created a competing PR. The migration to ESLint v9 has been burning under my nails for some time.


bugsounet:
Don't worry, I just wanted to start coding the tests with eslint v9.
I know that you are more skilled than me on this ;)

khassel (approved):

rejas (approved):

khassel:
@KristjanESPERANTO 

to be honest I have no clue about eslint but with this PR we have

```json
	"dependencies": {
		"eslint": "^9.11.1",
	},
	"devDependencies": {
		"@eslint/js": "^9.11.1",
	},
```

in `package.json`, the versions are the same and both npm packages refer to the same github repo.

Is this correct or would be the one under `dependencies` sufficient?

KristjanESPERANTO:
`eslint` is the real ESLint which we need in `dependencies` because we use it in the config check.

`@eslint/js` is a plugin for ESLint that we use for linting - so `devDependecies` is the right place for it.

So everything is fine in my eyes.

khassel:
Thanks for the explanation, I checked for dependency updates and saw the same update for both ...

KristjanESPERANTO:
The developers seem to synchronize the version number of both. So we will be seeing this even more often in the future.

I understand the doubt :slightly_smiling_face: They should have found a better name for the plugin.

khassel:
hm, I tested now without the `@eslint/js` (see [this commit](https://github.com/MagicMirrorOrg/MagicMirror/compare/develop...khassel:MagicMirror:develop)) and all tests were fine (?)

KristjanESPERANTO:
Yes, it works because the plugin is a dependency of ESLint.

https://github.com/eslint/eslint/blob/main/package.json#L104

Isn't it recommended to put direct dependencies into the package.json? We import the plugin in the ESLint config.

PR #3566, opened 2024-09-28
update dependencies, nail down node-ical version to 0.18.0

khassel:
- node-ical use `0.18.0` instead of `^0.18.0` in `package.json`
- cleanup `package-lock.json`


khassel:
The failed test is caused by the downgrade of `node-ical`, AFAIS we have to live with it (because we cannot upgrade `node-ical` before next release).

bugsounet:
You can tempory add `allow-ghsas: GHSA-8hc4-vh64-cxmj` in dep-review.yaml

```
      ....
      - name: 'Dependency Review'
        uses: actions/dependency-review-action@v3
        with:
          allow-ghsas: GHSA-8hc4-vh64-cxmj
```

We used this kind of exception some time ago

bugsounet:
It's green now :)

khassel:
yes, thanks!

@sdetweil can you please review/merge?

sdetweil:
looks good

Issue #3573, opened 2024-10-01
[tests] electron-rebuild test fails with node v20.9.0


khassel:
Problem is `npm install drivelist`, see e.g. [this job log](https://github.com/MagicMirrorOrg/MagicMirror/actions/runs/11123546494/job/30907210732).

The above command succeeds with latest node versions v20 and v22 (tested this). It is not related to MagicMirror, I can create a new project with `npm install -y` and `npm install drivelist` fails.

@bugsounet any idea?

Options:
- remove this test for node v20.9.0
- remove `drivelist`
- replace `drivelist` with another package

bugsounet:
Hi, i will inspect this evening

PR #3575, opened 2024-10-02
[Electron rebuild] Removed node-pty and drivelist from rebuilded test

bugsounet:
Related to #3573 

I think it's better to keep new library node-libgpiod for testing (library to manage RPI gpio for pir sensor management) and delete others
Only one test is better


bugsounet:
It's strange because 4 days ago we didn't have this problem.
See [there](https://github.com/MagicMirrorOrg/MagicMirror/actions/runs/11086767926/job/30804709045)

khassel:
> It's strange because 4 days ago we didn't have this problem.

yes, and there were no new releases ...

khassel (approved):

PR #3576, opened 2024-10-02
[UpdateNotification] Fix pm2 using detection when pm2 script is in MagicMirror root folder

bugsounet:
I discover this bug:

When pm2 `sh` script is on MagicMirror root folder, updatenotification is not able to detect pm2 using

```
0|MagicMirror  | [2024-10-02 17:23:09.215] [DEBUG]   Version Compare: 2.30.0-develop 2.30.0-develop --> true 
0|MagicMirror  | [2024-10-02 17:23:09.216] [DEBUG]   Status: online 
0|MagicMirror  | [2024-10-02 17:23:09.216] [DEBUG]   PM2 MagicMirror starting  from Path: /home/bugsounet/MagicMirror-dev 
0|MagicMirror  | [2024-10-02 17:23:09.216] [DEBUG]   MagicMirror Path /home/bugsounet/MagicMirror-dev/
0|MagicMirror  | [2024-10-02 17:23:09.216] [DEBUG]   Compare: false
0|MagicMirror  | [2024-10-02 17:23:09.216] [INFO]  updatenotification: [PM2] You are not using pm2 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ _This release is scheduled to be released on 2025-01-01._
 
 ### Fixed
 
+- [updatenotification] Fix pm2 using detection when pm2 script is in MagicMirror root folder
+
 ## [2.29.0] - 2024-10-01
 
 Thanks to: @bugsounet, @dkallen78, @jargordon, @khassel, @KristjanESPERANTO, @MarcLandis, @rejas, @ryan-d-williams, @sdetweil, @skpanagiotis.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ _This release is scheduled to be released on 2025-01-01._
 
 ### Fixed
 
-- [updatenotification] Fix pm2 using detection when pm2 script is in MagicMirror root folder
+- [updatenotification] Fix pm2 using detection when pm2 script is in MagicMirror root folder (#3576)
 
 ## [2.29.0] - 2024-10-01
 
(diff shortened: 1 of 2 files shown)


khassel (approved):

FrankBlackMG:
This change didn't work for me since my cwd in PM2 is different from my MagicMirror path. It didn't work previous to the fix either. The change below works for my setup and appears to be a solid way for a process to determine if it's managed by PM2. My only concern is that I'm not sure when unique_id was implemented in PM2.
```
//if (pm.pm2_env.version === this.version && pm.pm2_env.status === "online" && pm.pm2_env.pm_cwd.includes(`${this.root_path}`)) {

if (process.env.unique_id !== undefined && process.env.unique_id === pm.pm2_env.unique_id) {
```

FrankBlackMG:
unique_id is a GUID injected by PM2. It's in the process of MagicMirror where you are making the call to the PM2 API and it's also in the environment section of the pm object (pm.pm2_env.unique_id) when iterating over the list of PM2 managed processes. That seems definitive unless I'm missing something. I do think a linked file would work nicely, but it's something that has to be done rather than just making a simple code change. I dumped the whole env json from each pm object in the PM2 list and you can see all of the package.json vars e.g. name=magicmirror or the repository.url being "https://github.com/MagicMirrorOrg/MagicMirror" so those could be checked as well if you didn't trust the GUID match check.

bugsounet:
@FrankBlackMG Thanks for this purpose, il will inspect

PR #3577, opened 2024-10-04
reactivated `stale.yaml` as github action

khassel:
The old `stale.yaml` seems not to work anymore, so I set up the same content in a new github workflow.

I think we should use it again to get rid of old issues.


rejas (approved):

khassel:
I thought the job would run this morning, but nothing.

From googling it seems it must on the default branch to run (which means with next release).

Will try later to start it per api to see if it works ...

khassel:
I now managed to run the stale workflow, did it already several times because it is rate limited. I think not all issues needed are marked as stale so I will try later again.

For documentation:

```bash
❯ gh auth login --with-token < gh_token

❯ gh workflow run .github/workflows/stale.yaml --repo MagicMirrorOrg/MagicMirror --ref develop
✓ Created workflow_dispatch event for stale.yaml at develop
```

PR #3578, opened 2024-10-05
[core] add try / catch on mode_helper loading

bugsounet:
When a library is missing on an 3rd party module, MM² stop loading and display a black screen. (I'm sure it's happened to everyone.)

So, I have added a try/catch block and it's avoid black screen, display errors and allow continue loading with next module


sdetweil:
nice!

rejas (approved):

PR #3584, opened 2024-10-11
[core] test electron v32 and electron rebuild

bugsounet:
test deps: nan v2.22.0 and electron v32


bugsounet:
@khassel , it's green ;)

PR #3585, opened 2024-10-11
add address and ipWhitelist to all test configs

khassel:
All test configs have been updated to allow full external access, allowing for easier debugging (especially when running as a container)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ _This release is scheduled to be released on 2025-01-01._
 
 - [repo] reactivated `stale.yaml` as github action to mark issues as stale after 60 days and close them 7 days later (if no activity)
 - [core] Update electron dependency to v32 (test electron rebuild)
+- [tests] All test configs have been updated to allow full external access, allowing for easier debugging (especially when running as a container)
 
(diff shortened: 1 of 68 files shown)


sdetweil:
just added 6 configs today, sorry

khassel:
> just added 6 configs today, sorry

on your fork or here?

sdetweil:
on my fork.  i'll fix it..

sdetweil:
one other thing, relative to test..
I was trying to do individual tests, via npm start with MM_CONFIG_FILE=test/configs/....
but
the test  path isn't enabled unless JEST is running.. (so couldn't load the ics file) 
so had to define that env var (as something) to get the path 
was worried about other side effects

khassel:
> the test path isn't enabled unless JEST is running.. (so couldn't load the ics file)

I added this with the last release and can revert it (with another PR) if you want. I didn't think of your use case

sdetweil:
I can put it back with mine, just wanted to make sure there was no other intent

khassel:
o.k., it was this change:

grafik

rejas (approved):

PR #3586, opened 2024-10-13
Re-add eslint-plugin-import

rejas:
eslint-plugin-import was missing since the switch to [v9](https://github.com/MagicMirrorOrg/MagicMirror/pull/3558). They finally [support](https://github.com/import-js/eslint-plugin-import/pull/2996) it so we can re-add it.


eslint.config.mjs
```diff
@@ -2,9 +2,11 @@ import eslintPluginJest from "eslint-plugin-jest";
 import eslintPluginJs from "@eslint/js";
 import eslintPluginStylistic from "@stylistic/eslint-plugin";
 import globals from "globals";
+import importPlugin from "eslint-plugin-import";
```
KristjanESPERANTO, line 5: Can we keep a consitant variable name pattern for plugins?

I suggest `eslintPluginImport`.

rejas: sure :-)

KristjanESPERANTO (approved):

PR #3588, opened 2024-10-19
Run and test with node 23

KristjanESPERANTO:


.github/CONTRIBUTING.md
```diff
@@ -43,7 +43,7 @@ When submitting a new issue, please supply the following information:
 
 **Platform**: Place your platform here... give us your web browser/Electron version _and_ your hardware (Raspberry Pi 2/3/4, Windows, Mac, Linux, System V UNIX).
 
-**Node Version**: Make sure it's version 18 or later (recommended is 20).
+**Node Version**: Make sure it's version 20 or later (recommended is 22).
```
rejas, line 46: v18 is still supported until april 2025 so why the bump?

KristjanESPERANTO: Since `"node": ">=20.9.0 <21 || 22` was already in the package.json we don't support v18. I didn't investigate why v18 was removed, I just adopted the rest to it.

khassel: The 2 latest releases already need node >= v20.9.0 which is a restriction of eslint.

khassel (approved):

Issue #3582, opened 2024-10-08
Wayland support


goldyfruit:
Hi,

Raspberry Pi OS Bookworm is now defaulting to Wayland and Wayfire and because of this change MagicMirror is not able to start via the standard `npm run start` as its looking for `DISPLAY` to be set which doesn't exist anymore on Wayland.

Maybe adding a `start-wayland` option to `package.json` will be a requirement in order to allow users running on modern distributions to use MagicMirror.

This is how I'm running it  and its working very well so far:

```
WAYLAND_DISPLAY="${WAYLAND_DISPLAY:=wayland-1}" ./node_modules/.bin/electron js/electron.js --enable-features=UseOzonePlatform --ozone-platform=wayland
```

Related to https://gitlab.com/khassel/magicmirror/-/issues/74

KristjanESPERANTO:
I'm running Debian 11 with Wayland but without Wayfire and I have no problems with `npm run start`. So maybe Wayfire is the "problem" and not Wayland?

goldyfruit:
I'm running "pure" Wayland _(not xwayland)_, same issue happens with Sway.

KristjanESPERANTO:
Oh okay, I just checked: Electron is running under xwayland on my system.

I'm just thinking out loud: Would it be possible to put a check into "npm run start" to see if Wayland or X is running and call the right command? This way we could also check if Windows is running (they also need another command).

bugsounet:
hmm, interesting :)

I just write a test script for test it

----
```
pi@raspberrypi:~ $ ./test.sh 
Running: Wayland
pi@raspberrypi:~ $ 
```
----
```
pi@raspberrypi2:~ $ ./test.sh 
Running: X11
pi@raspberrypi2:~ $ 
```
----

now the question is... what language to use for code it ?
bash ? or node ?

* if we use bash we have linux / mac compatibilities BUT not windows
* if we use node I think we have both compatibility

Or ... just add un new start line with something like this: `npm start wayland`, `npm start windows`

khassel:
> just add un new start line with something like this: npm start wayland, npm start windows

I think this is sufficient.

The wayland scenario is an edge case at the moment, in all raspian os images with desktop XWayland is installed which works with X11 as usual.

bugsounet:
@khassel ok ;)

@KristjanESPERANTO , do you want to code it ?  (i'm lazy mouahah)

sdetweil:
[...]

PR #3594, opened 2024-10-23
Add wayland and windows start options

KristjanESPERANTO:
This PR adds start options for Wayland and Windows.

This would solve issue #3582.

**To Do if this PR is merged:**

- [ ] Adjust [Windows section](https://docs.magicmirror.builders/getting-started/installation.html#windows) in documentation
- [ ] Add Wayland section to the documentation


KristjanESPERANTO:
I dropped the attempt to add an windows installation option for this PR.

sdetweil:
@KristjanESPERANTO if you used my shell js app it could be on install-mm

KristjanESPERANTO:
> @KristjanESPERANTO if you used my shell js app it could be on install-mm

I think it's good to handle this in different PRs. This one handles the start options and a new one could handle the installation.

khassel (approved):

khassel:
btw tested this in VirtualBox

```bash
"install-mm:windows": "cmd /c npm install --ignore-scripts --no-audit --no-fund --no-update-notifier --only=prod --omit=dev && echo 'Installing vendor files ...' && cd vendor && npm install --loglevel=error --no-audit --no-fund --no-update-notifier && echo 'Installing fonts ...' && cd fonts && npm install --loglevel=error --no-audit --no-fund --no-update-notifier",`
```

and beside a small error (should be `cd ../fonts`) it worked :)

sdetweil:
i think this could have been done together

my postinstall using my launcher 
```json
"postinstall": "node js/es.js \"npm run install-vendor && npm run install-fonts && echo MagicMirror² installation finished successfully!\n\":npm run install-vendor & npm run install-fonts & echo MagicMirror² installation finished successfully!",
```
then we don't need multiple divergent commands..  (to document and support)

sdetweil:
then we just use run mm-install
if you ran install postinstall would be the same

sdetweil:
we can do the same on start too.. 

js/es.js linux_command : windows_command

KristjanESPERANTO:
> we can do the same on start too..
> 
> js/es.js linux_command : windows_command

We would need something like this: `js/exec_script.js x11_command : wayland_command : windows_command`.

> less confusing for our users..

Yes, but different start options should not overwhelm a tinkerer. Also, when someone uses the additional commands, they should always be reminded that they are doing something that is not the default. This might help them to solve problems themselves and need less support.

Perhaps I will argue in a different direction in the future 😅. But I often have the thought that it might not be a good idea to make it too easy for our users (who are tinkerers). If we automate everything, they understand the system less and need more support. That's why I liked the decision that only the manual installation is the official one.

PR #3595, opened 2024-10-23
Add npm publishing step to release process

rejas:
so that the npm version also stays in sync and https://github.com/MagicMirrorOrg/MagicMirror/issues/2876 can be closed for good


KristjanESPERANTO (approved):

khassel:
we have a new test problem (unrelated to this PR) ... I hate it.

khassel:
@rejas you added a one liner for npm publishing. AFAIS only you can do it and the real doing is not clear for me.

rejas:
> @rejas you added a one liner for npm publishing. AFAIS only you can do it and the real doing is not clear for me.

moving this discussion to discourse...

PR #3596, opened 2024-10-23
Fixtestcase calendar testcase failure

sdetweil:
fix calendar testcase failing after date change (exposes helper failure)


sdetweil:
can someone restart this as I changed the base branch while it was completing

sdetweil:
can someone restart this as I changed the base branch

khassel:
AFAIS the helper is not working (anymore) so every (electron) test running with a specific date does not run with this date but uses today.

So main problem is to fix the helper or find another solution because other tests will fail too in the future.

This is only a hotfix to get the tests running and should be reverted when the main problem is solved.

Will merge this and open an issue for the main problem.

PR #3599, opened 2024-10-24
fix electron tests mocking dates

### LINKED ISSUE #3597: date based electron tests are using today instead of specific date
see https://github.com/MagicMirrorOrg/MagicMirror/pull/3596

The helper is not working (anymore) so every (electron) test running with a specific date does not run with this date but uses today.

Above PR should be reverted if this is solved.

khassel:
fixes #3597 

Changes:
- electron tests: add mocking to `electron.js` for mocking the server side, before only the browser side was mocked
- publish "/tests/configs" and "/tests/mocks" always in `server.js`, this reverts a change done with latest release, we need this for debugging (otherwise you get on the screen that your config has errors but config check is successful ...)
- revert hotfix in `tests/configs/modules/calendar/show-duplicates-in-calendar.js`
- fix `tests/configs/modules/calendar/custom.js` to allow events in the past (~~I don't know how this could work before~~ when testing  css classes `yesterday` and `dayBeforeYesterday` --> it worked before because the server side did not mock and therefore was not in the past)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ _This release is scheduled to be released on 2025-01-01._
 - [updatenotification] Fix pm2 using detection when pm2 script is in MagicMirror root folder (#3576)
 - [core] Fix loading node_helper of modules: avoid black screen, display errors and continue loading with next module (#3578)
 - [weather] changed default value for weatherEndpoint of provider openweathermap to "/onecall" (#3574)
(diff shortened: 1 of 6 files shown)


sdetweil (approved):
looks good, so net have the electron process set the date for  the test run, helper just passes it on thru env. cool

rejas (approved):

PR #3601, opened 2024-10-25
Cleanup testcases that had hard coded Date() values which override the testcase runner

sdetweil:
cleanup testcases with hard coded Date settings after #3597


sdetweil:
i edited and changed to develop, job rerun wont chamge altho other jobs do

khassel:
yes, I think if it was `master` at the beginning this test will always fail. Will merge it now.

PR #3611, opened 2024-10-29
TEST electron v33

bugsounet:
Test:
electron v33.x and electron rebuild


bugsounet:
Link to [nan - Support Node 23](https://github.com/nodejs/nan/pull/979)

test with node23 branch

bugsounet:
Pas de réponse, en attente de la mise a jour de la librairie

PR #3623, opened 2024-11-03
Add spelling check to GitHub workflow

KristjanESPERANTO:
Besides updating cspell and handling spelling issues, the important change is adding the spelling check to the GitHub workflow.

I'm not sure if it will bother us too much when people create PRs. But I wanted to give it a try. Or do you have any other ideas on how we can run the spelling check on a regular basis?


sdetweil:
i saw that the node-ical workflow added coderabbit
which added a codeblock around an example i gave, and improved the text i added in doc

pretty impressive

khassel:
> Or do you have any other ideas on how we can run the spelling check on a regular basis?

maybe as cronjob as we have already in `stale.yaml` on `develop`?

KristjanESPERANTO:
> maybe as cronjob as we have already in `stale.yaml` on `develop`?

That sounds like a better approach! I added a GitHub workflow to run the spell check a few days before each release.

.github/workflows/spellcheck.yaml
```diff
@@ -0,0 +1,32 @@
...
+  contents: read
+
+jobs:
+  spellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: development
```
khassel, line 20: if this is the git ref: `develop` ?

can we remove the following empty line?

KristjanESPERANTO: > if this is the git ref: `develop` ?

Of course :facepalm: -> fixed :heavy_check_mark: 

> can we remove the following empty line?

Removed :heavy_check_mark:

khassel:
AFAIK the cronjobs needs to be on `master` wo work so we have to do the spellcheck manually before next release ...

KristjanESPERANTO:
> AFAIK the cronjobs needs to be on master

That would be suboptimal. Let's see what the job will do on 27th December :slightly_smiling_face:

PR #3605, opened 2024-10-27
[updatenotification] Fix pm2 using detection when pm2 script is inside or outside MagicMirror root folder

bugsounet:
This will fix #3576 

@FrankBlackMG: 

I don't use `*env.unique_id` because some others modules can use pm2 too for starting a service and unique_id is the same (this will make confusion)
So I check `name` and `pm_id` for found it

PR #3626, opened 2024-11-07
PM2 Fix (again): add `pm2_env.unique_id` checking and cleaning

bugsounet:
#3605 : new purpose code with `pm2_env.unique_id` checking


bugsounet:
@khassel : ok, i will reverse it.
@FrankBlackMG : you have right ;)

rejas merges it at lightning speed! (it's my fault I should have drafted it)

PR #3630, opened 2024-11-09
Enhance compliments remote file with refresh support

sdetweil:
add support to refresh the compliments remotefile
add testcases for both without refresh (testcase missing) and with refresh

doc to be updated

Issue #3610, opened 2024-10-29
in test mode is there a way for browser side to know in test?


sdetweil:
in helper side one can test the  process.env.JEST_WORKER_ID
not available in module side

I am helping user with new compliments parm test

khassel:
from `tests/e2e/helpers/global-setup.js`:

```js
exports.getDocument = () => {
	return new Promise((resolve) => {
		const url = `http://${config.address || "localhost"}:${config.port || "8080"}`;
		jsdom.JSDOM.fromURL(url, { resources: "usable", runScripts: "dangerously" }).then((dom) => {
			dom.window.name = "jsdom";
			global.window = dom.window;
			// Following fixes `navigator is not defined` errors in e2e tests, found here
			// https://www.appsloveworld.com/reactjs/100/37/mocha-react-navigator-is-not-defined
			global.navigator = {
				useragent: "node.js"
			};
			dom.window.fetch = fetch;
			dom.window.onload = () => {
				global.document = dom.window.document;
				resolve();
			};
		});
	});
};
```

we are setting here `dom.window.name = "jsdom";` so this could maybe used in the e2e-tests.

sdetweil:
I meant the code running inside MM.. a module

to test the refreshable compliments remote file and have it locally we need to 'change' the configured url on the fly to it will pull a different file.  but i have another way... if the url starts with tests and there is another parm (url2...)

khassel:
> I meant the code running inside MM.. a module

and I meant you maybe can use 

```js
if (global.window.name === "jsdom")` {
  console.log("running in test");
}
```

sdetweil:
ok, thx

sdetweil:
```

**Comment by khassel** [comment]
`In the browser environment, global variables are attached to the `window` object.`

so for code running in browser omit the `global`:

```js
if (window.name === "jsdom")` {
  console.log("running in test");
}
```

sdetweil:
hm.. doesn't fail, but doesn't work 
```js 
		// in test mode only
		if (window.name === "jsdom") {
			// check for (undocumented) remote file2 to change test new load
			if(this.config.remoteFile2!== null && this.config.remoteFileRefreshInterval!==null){
	 			 console.log("running in test");
	 			 this.config.remoteFile=this.config.remoteFile2
			}
		}
```
for changing the remote file to prove that the cycle and fetch work , only wanted to fake a new file in test mode
if I comment out the if jsdom, then testcase works.
[...]

PR #3631, opened 2024-11-11
add support for test mode detection in modulename.js via index.html

sdetweil:
in some cases the modulename.js may need to detect running in test mode (compliments pr #3630)

window.name is not set  web mode

add a new field to the index.html 
window.intest 
and use the server_function to replace the  hard coded string like we do for window.mmversion=#VERSION#
then change the two  test helpers to set the env variable
app.js detects and sets global.intest=true
server func replace with value of global.intest

then module can use   if(window.intest)

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,7 +18,7 @@
 
     <script type="text/javascript">
       window.mmVersion = "#VERSION#";
-      window.intest = "#TESTMODE#";
+      window.mmTestMode = "#TESTMODE#";
     </script>
   </head>
   <body>
--- a/index.html
+++ b/index.html
@@ -18,6 +18,7 @@
 
     <script type="text/javascript">
       window.mmVersion = "#VERSION#";
+      window.intest = "#TESTMODE#";
     </script>
   </head>
   <body>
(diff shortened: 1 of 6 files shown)


sdetweil:
enhancement for #3610

tests/electron/helpers/global-setup.js
```diff
@@ -39,6 +40,7 @@ exports.stopApplication = async () => {
 	global.electronApp = null;
 	global.page = null;
 	process.env.MOCK_DATE = undefined;
+	process.env.jstest = undefined;
```
rejas, line 43: why is it now a "jstest" instead of "intest" ?

index.html
```diff
@@ -18,6 +18,7 @@
 
     <script type="text/javascript">
       window.mmVersion = "#VERSION#";
+      window.intest = "#TESTMODE#";
```
rejas, line 21: regarding naming: what does "intest" mean, reminds me of "intestines" :-) maybe just "mmTestMode" ?

rejas (changes requested):

tests/electron/helpers/global-setup.js
sdetweil: weird, wonder where that came from. will fix

index.html
sdetweil: in_test   

but your suggestion is ok

rejas (approved):

PR #3636, opened 2024-11-17
Remove `@eslint/js` dependency. Already installed with `eslint` in deep

bugsounet:
I lint my modules and just see this:

`@eslint/js` is now not needed.
it's installed by `eslint` it self


rejas (approved):

bugsounet:
pinging: @KristjanESPERANTO do you see this ?

KristjanESPERANTO:
We've had this issue at some point before. I usually prefer to have dependencies that we import in the `package.json`. But in this case we will probably never run into a real problem, so I think the change is okay :slightly_smiling_face:

PR #3644, opened 2024-12-01
Show animations

sdetweil:
I am adding the animateIn/Out support to MMM-Config, but I need the list of animations..
but they are not visible in js/animateCSS.js (from require)

adding an export satisfies that

side issue, these would go in a dropdown list
what value can I use for the default behavior? none/default?
don't want an error produced..
should I add code to check for this value to prevent error?


sdetweil:
no need to add code, if the entry is not found in list it is ignored..

rejas (approved):

bugsounet:
image

rejas:
@sdetweil could you take a look please?

sdetweil:
see #3649

PR #3681, opened 2025-01-04
fix #3267, CORRECTLY, add testcase, add testcase for #3279

### LINKED ISSUE #3267: Weird broadcasted form of multi symbols of calendar module.
I remember it was not so,  but multiple symbols are broadcasted weirdly at some point.

When I set some symbols in a calendar like  this;
```js
symbol: ["file", "calendar"],
```

The expected broadcasted data(And the historical format what I remember) would be like this; (with `defaultSymbolClassName`)
```js
symbol: ["fa fa-fw fa-file", "fa fa-fw fa-calendar"],
// Array(2)
```

Bur really broadcasted would be;
```js
symbol: ["fa fa-fw fa-file, fa fa-fw fa-calendar"],
// Array(1)
```

Is this intentionally designed so? Or just an unexpected accident?

sdetweil:
fixes #3267 AGAIN, correctly, add testcase
add testcase for #3679 , broadcast clipping incorrectly

I added a test module (tests/testNotification) to catch the notification and check the count. (one  way to configure)
i put this module in the tests folder, and added /tests to the server paths. 
(can't have a module in a nested folder, like tests/modules/xxx)

but I have a problem.  i can run the test config (MM_CONFIG_FILE), and the two modules work correctly, 
but in the spec runner, the calendar module times out on the broadcast test..  there is only local ics file access, no outside hosts

I forced my system date to 1/1/24 (same as runner) and again the manual testcase works fine

I added two test config.js,(configs/calendar) one works great (symbols) , one fails (broadcast test)
I added additional delay in the calendarspec runner to try to debug the module, but it still not long enough.. no messages of trouble when I get into the browser.. BUT, this may be because of the log being turned off... (just thought of this)

I created a special ICS (in mocks)  that has 12 events, 1 for each month.. (so I can check clipping and broadcast) the US holidays one is sensitive to the current date, and I couldn't get it to work on 1/1/2024..  

also, in general, is there a mechanism to run test:just_one_runner?  waiting thru the electron test to get to one testcase.. ugh..


sdetweil:
Ok, the problem is here in app.js
```js
	function loadModule (module) {
		const elements = module.split("/");
		const moduleName = elements[elements.length - 1];
		const env = getEnvVarsAsObj();
		let moduleFolder = path.resolve(`${__dirname}/../${env.modulesDir}`, module);

		if (defaultModules.includes(moduleName)) {
			const defaultModuleFolder = path.resolve(`${__dirname}/../modules/default/`, module);
			if (process.env.JEST_WORKER_ID === undefined) {
				moduleFolder = defaultModuleFolder;
			} else {
				// running in Jest, allow defaultModules placed under moduleDir for testing
				if (env.modulesDir === "modules") {
					moduleFolder = defaultModuleFolder;
				}
			}
		}
```

so I am using two modules 
calendars.js (default module) 
test/testNotification foreignModulesDir: "tests"


so moduleFolder never gets set for defaultModuleFolder. 
I don't WANT modules to the the location for this test only module

if I change the test like this 
```js
if (env.modulesDir === "modules" || env.modulesDir === "tests") {
```
then it works with the  foreignModulesDir:"tests"

is this acceptable?

khassel:
I'm not happy with these changes.

You could move the special `testNotification` stuff beside the electron test and then copy it into the needed folder before doing the test, similar to [this in neewsfeed test](https://github.com/MagicMirrorOrg/MagicMirror/blob/master/tests/e2e/modules/newsfeed_spec.js#L88).

This would avoid the changes in `app.js` and `server.js`.

sdetweil:
I think we are talking about different problems..

the testcase needs to receive a notification, sent by calendar.   and the testcase needs to examine the data sent, 
but it is not visible to the testcase 

also, the same module cannot receive notifications it sends... so I can't put code IN calendar to access that data

so I need s NEW module that can receive the notification, and present some web content that the testcase runner can examine

ok, now I have this new module, where to put it ...

I originally put it in defaults, but its not a general purpose module, its ONLY for use during test.

so, put it in tests, but there is no server path to tests, so it can't be loaded. 
fix that (server.js) but then the defaults path  is broken (app.js)

are you proposing I copy the module from tests folder to modules folder?  ok, but then this testcase cannot be run on its own without the user KNOWing they have to copy the folder first..

khassel:
> but then this testcase cannot be run on its own without the user KNOWing they have to copy the folder first..

you do the copy inside the test case with e.g. `fs.cpSync` as already done in newsfeed test, see link in my comment above

sdetweil:
yes, i have that working 

can you explain what your concern is w having the addition test assets IN  the tests folder.  and exposing that?  (we do that for all the other test assets, you added the config and mocks to be accessible outside test)

this is different than the newsfeed test , where you NEED to the module to be in a different location to test the foreignModuleDir property

sdetweil:
> you do the copy inside the test case with e.g. fs.cpSync as already done in newsfeed test, see link in my comment above

yes, so you can do 
export MM_CONFIG_FILE=the_testcase_config
npm start and it will work
that will NOT work for this test as the copy is done (under the covers) in the runner

khassel:
other possibility (instead of the copy) would to move the `testNotification` into an already published `tests` folder.

this makes no sense:

```bash
let directories = ["/config", "/css", "/fonts", "/modules", "/vendor", "/translations", "/tests/configs", "/tests/mocks", "/tests"];
```

if we really need to publish the whole `tests` folder the 2 subfolders should be removed.

sdetweil:
yes, testNotification module folder is ALREADY in tests in the PR as submitted

I admit that I screwed up.. and should have removed the explicit sub folder shares.

khassel:
I would move `testNotification` under `tests/mocks`

sdetweil:
Ok, moved, and we are back to this hack in app.js
```
		let moduleFolder = path.resolve(`${__dirname}/../${env.modulesDir}`, module);  ///assume module is in 'modules' folder

		if (defaultModules.includes(moduleName)) {  // if its a default module
			const defaultModuleFolder = path.resolve(`${__dirname}/../modules/default/`, module); // calc new path
			if (process.env.JEST_WORKER_ID === undefined) {  // if not in test mode
				moduleFolder = defaultModuleFolder;   // use the defaults folder
			} else {
				// running in Jest, allow defaultModules placed under moduleDir for testing
				if (env.modulesDir === "modules" || env.modulesDir === "tests/mocks") { //<--

					moduleFolder = defaultModuleFolder;
				}
			}
		}
```


//<-- have to add, else non-default modules are ONLY allowed in wherever 'modules' folder is set, so we would be back to copy
the config works ok outside test without this
this testcase is failing on the default calendar module not being found (as its not in 'modules' folder, set as default)

but this exposes using a default module for 3rd party module location test..
  
this is set in the testcase config, as module search is not recursive..  

```js
foreignModuleDir:"tests/mocks"
```

js/server.js
```diff
@@ -72,7 +72,7 @@ function Server (config) {
 			app.use(helmet(config.httpHeaders));
 			app.use("/js", express.static(__dirname));
 
-			let directories = ["/config", "/css", "/fonts", "/modules", "/vendor", "/translations", "/tests/configs", "/tests/mocks"];
+			let directories = ["/config", "/css", "/fonts", "/modules", "/vendor", "/translations", "/tests"];
```
khassel, line 75: why is this still needed when moved to `/tests/mocks`?

tests/electron/modules/calendar_spec.js
```diff
@@ -12,10 +12,11 @@ describe("Calendar module", () => {
 		await expect(elem.isVisible()).resolves.toBe(true);
 		return true;
 	};
+	//doTestTableContent(".testNotification tr", "td.elementCount", "12", first)).resolves.toBe(true);
```
khassel, line 15: can this be removed?

js/server.js
sdetweil: because in test mode you leave the path as modules, but the calendar module i use is in modules/default

so it is not found , read my post before the push

tests/electron/modules/calendar_spec.js
sdetweil: yes,will do, copied to verify the call parms matched  

there was a day where it worked and then suddenly didn't. i had the test module in defaults (had to change other files to allow that., wrong)

then moved module to tests and nothing worked
cause of the path problem in app.js, cause of foreign setting

sdetweil: yes

js/server.js
khassel: tested your branch with `let directories = ["/config", "/css", "/fonts", "/modules", "/vendor", "/translations", "/tests/configs", "/tests/mocks"];` and it works

sdetweil: oh, tests/mocks back in, yes

sdetweil: have grandson til 3pm now

sdetweil:
after sync and pull  (I changed a file directly and it had only spaces..)

anyhow

tests don't run 

platform
```text
Linux sams 5.15.0-127-generic #137-Ubuntu SMP Fri Nov 8 15:21:01 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
(base) sam@sams:~/MagicMirror.old$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy
```
output
```text
    electron.launch: Process failed to launch!

      16 | 	electronParams.unshift("js/electron.js", "--enable-features=UseOzonePlatform", "--ozone-platform=wayland");
      17 |
    > 18 | 	global.electronApp = await electron.launch({ args: electronParams });
         | 	                     ^
      19 |
      20 | 	await global.electronApp.firstWindow();
      21 |
```
suite summary 
```text
Test Suites: 4 failed, 4 total
Tests:       46 failed, 46 total
Snapshots:   0 total
Time:        37.717 s
```

sdetweil:
I edited on the web directly so I wouldn't drag in the new runner problem..

khassel (approved):

PR #3690, opened 2025-01-13
chore: update newsfeed.js

eltociear:
Therefor -> Therefore

PR #3692, opened 2025-01-13
chore: update newsfeed.js

eltociear:
Therefor -> Therefore

refs: #3690


khassel:
@rejas well, this fixes only a typo in a comment so I think we can merge this without a CHANGELOG entry (I don't really want to read such stuff in a CHANGELOG ...)

PR #3695, opened 2025-01-14
Use different issue templates

KristjanESPERANTO:
This PR will introduce different issue templates for bug reports, feature requests and so on on GitHub. There is still room for fine-tuning, but it's reached a state to show it to you and get feedback.

I think that this can lead to better bug reports.

You can see the result in my repo: https://github.com/KristjanESPERANTO/MagicMirror/issues/new/choose

Feel free to create new issues for testing.

What do you think? Do we want to pursue this further?


rejas (approved):
There is no way as far as I know to test this locally, so I am fine with this being tested "live" What do you think @khassel @sdetweil ?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,141 @@
...
+    attributes:
+      label: Which start option are you using?
+      description: |
+        Please keep in mind that some problems are specific to certain start options.
+      options:
+        - "npm run start"
+        - "npm run start:wayland"
+        - "npm run start:windows"
+        - "npm run start:x11"
+    validations:
```
khassel, line 40: looks like `npm run server` is missing

khassel:
> There is no way as far as I know to test this locally

 did some testing using the fork of @KristjanESPERANTO 

> so I am fine with this being tested "live" What do you think @khassel @sdetweil ?

@rejas  I'm fine with live testing. Currently the PR is based on `develop` so "live" testing would start in April ...

So from my side after merging on `develop` we could cherry pick the commit (without changelog stuff) and push it on `master` too. So we would have real live testing. Update notification is no problem, it reports only new versions on `master` if a new tag is applied.

.github/ISSUE_TEMPLATE/bug_report.yml
khassel: and start of clientonly is missing too ...

sdetweil:
i agree, no local testing, don't see why not pushing to
master as soon as ready.

.github/ISSUE_TEMPLATE/bug_report.yml
KristjanESPERANTO: I just added both :+1:

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,141 @@
...
+      value: |
+        MagicMirror² version:
+          (It can be found in the `package.json` file. Please use the latest version before reporting a bug.)
+        Node version: 
+          (Run `node -v` to find out. Make sure it's version 20 or later (recommended is 22))
+        npm version: 
+          (Run `npm -v` to find out.)
+        Platform: 
+          (Raspberry Pi 2/3/4/5, Windows, Mac, Linux, System V UNIX, ...)
+    validations:
```
khassel, line 27: maybe we should ask here for the systeminformation from the logs (as replacement for the last 3 values), this one:

```bash
[2025-01-14 20:05:03.529] [INFO]  System information:
### SYSTEM:   manufacturer: Microsoft; model: Docker Container; virtual: true
### OS:       platform: linux; distro: Debian GNU/Linux; release: 12; arch: x64; kernel: 5.15.167.4-microsoft-standard-WSL2
### VERSIONS: electron: undefined; used node: 23.5.0; installed node: 23.5.0; npm: 10.9.2; pm2:
### OTHER:    timeZone: Europe/Berlin; ELECTRON_ENABLE_GPU: undefined
```

rejas: > I just added it. Is that okay?

@khassel you're fine with it?

khassel: sure, forgot to resolve ...

khassel:
@rejas are you fine with cherry picking this on `master`?

rejas:
> @rejas are you fine with cherry picking this on `master`?

Absolutely

khassel:
o.k., will do it ...

khassel:
it's live now ...

rejas:
Looks neat. Lets see if people can read this better :-)

sdetweil:
might have to add text under 'logs' for the system info.. most newish users still don't know what/where to find that 
(and pm2  involved very often  too)

PR #3706, opened 2025-01-26
Fix frozen Yr weather-module

### LINKED ISSUE #3296: Weather module (yr) stops after a while
Running the latest MM v2.25.00 with the YR weather fix from https://github.com/MichMich/MagicMirror/issues/3227
I have 2 weather modules, 1 which shows the current weather and 1 that shows the forecast with the following config:

`// Weather now
{
	module: "weather",
	position: "top_right",
	config: {
		weatherProvider: "yr",
		type: "current",
		initialLoadDelay: 1000,
		updateInterval: 60000,
		onlyTemp: false,
		showWindDirection: false,
		showSun: false,
		lat: xx.xxxx,
		lon: xx.xxxx,
		altitude: 111
	}
},
// Daily forecast
{
	module: "weather",
	position: "top_right",
	config: {
		weatherProvider: "yr",
		type: "daily",
		initialLoadDelay: 3000,
		updateInterval: 120630,
		maxNumberOfDays: 7,
		fade: true,
		fadepoint: 0.25,
		lat: xx.xxxx,
		lon: xx.xxxx,
		altitude: 111
	},
},
`

After the module starts it stalls after a while, unsure how long after but within the same day.
I looked at the dev_tools, and this is the error log:

image

For some reason it gets an error response from YR, which causes the module to stop updating.

MagMar94:
Fixes #3296 

The problem was that the fetch-methods threw errors when something went wrong instead of calling `updateAvailable()`.  `updateAvailable()` must be called in order to schedule the next update.

I added some filtering for the hourly forecast that removes hours in the past. If the API call fails we use the cached data, but we should only display hours in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ planned for 2025-04-01
 - [core] Optimize systeminformation calls and output (#3689)
 - [core] Add issue templates for feature requests and bug reports (#3695)
 - [core] Adapt `start:x11:dev` script
+- [weather/yr] The Yr weather provider now enforces a minimum `updateInterval` of 600 000 ms (10 minutes) to comply with the terms of service. If a lower value is set, it will be automatically increased to this minimum.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ planned for 2025-04-01
 - [weather] Fix wrong weatherCondition name in openmeteo provider which lead to n/a icon (#3691)
 - [core] Fix wrong port in log message when starting server only (#3696)
 - [calendar] NewYork event processed on system in Central timezone shows wrong time #3701
+- [weather/yr] The Yr weather provider is now able to recover from bad API resposes instead of freezing (#3296)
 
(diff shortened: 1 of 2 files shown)


khassel:
>  I tried setting the updateInterval in the start-method with this.config.updateInterval = 600000 and this.setConfig({ ...config, updateInterval: 600000 }), but the setInterval still used the original value.

can confirm this, so we live with the error message instead. Or may @rejas has an idea how to change `config.updateInterval` within a provider?

khassel (approved):

khassel:
after some digging in the code I found in `weatherprovider.js` this part

```js
WeatherProvider.initialize = function (providerIdentifier, delegate) {
	const pi = providerIdentifier.toLowerCase();

	const provider = new WeatherProvider.providers[pi]();
	const config = Object.assign({}, provider.defaults, delegate.config);

	provider.delegate = delegate;
	provider.setConfig(config);
...
```

which explains the behavior.

So I tested `this.delegate.config.updateInterval = 123456;` inside the provider (in `setConfig`) which works. So if you are willing to change it again for better user experience ...

Issue #3708, opened 2025-01-30
[Bug] if a node_helper has an init function, the 'this' pointer is not set


sdetweil:
### Environment

System information:
### Which start option are you using?

npm run start

### Are you using PM2?

No

### Module

None

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

```
        init: function(){
                Log.log(this.name + " is in init!");
        },
```

from the doc 
```text
init()
This method is called when a module gets instantiated. In most cases you do not need to subclass this method.
```

### What did you expect to happen?

I expected a log message (using my brithdaylist module)

birthdaylist is in init

### What actually happened?

undefined is in init

### Participation

- [x] I am willing to submit a pull request for this change.

khassel:
Had same problem in a module last week. Found a solution and can provide a PR (but it's o.k. for me if you want to implement this) ...

sdetweil:
no, if you got it, go

sdetweil:
i tried to open this issue from my phone, but it wouldnt let me NOT fill in the system info section. so i had to get out my laptop, run MagicMirror again to get it

KristjanESPERANTO:
> i tried to open this issue from my phone, but it wouldnt let me NOT fill in the system info section. so i had to get out my laptop, run MagicMirror again to get it

Which browser did you use? I just checked with Chrome on Android and the system info section was editable.

sdetweil:
I did not HAVE  the info on my phone, and my computer was put away due to travel. 

I figured I could edit it later to paste in the info, which is not relevant  to this problem.

PR #3710, opened 2025-01-31
set name and path in module init

khassel:
This is a solution for #3708 

I set it to draft because I'm not sure if it should be merged.

With this change the `this.name` will now work in the `init` function but for get this working in a module you have to copy the content of the node_helper `init` into your module. This is nothing new, you have to do the same for e.g. `start`.

AFAIS the used construction in `js/class.js` misses inheritence for this.


khassel:
test fails will be fixed with https://github.com/MagicMirrorOrg/MagicMirror/pull/3709

sdetweil:
yeh, I dont think we should go that way

khassel:
o.k.

PR #3726, opened 2025-02-24
Fix empty part-of-day logic

rejas:


sdetweil:
where is the code change? i see tests but no change to compliments.js

rejas:
Just a draft because its still work in progress. I want to create some tests first

rejas:
now its ready @sdetweil

PR #3733, opened 2025-03-01
update dependencies incl. electron to v35

khassel:


sdetweil:
can we include @electron/rebuild as part of the base as an aide for module developers?

khassel:
> as an aide for module developers?

a module developer should be able to install it if he needs it.

I'm not for including because it will not work everywhere, e.g. it will not work in the normal docker images (maybe in the `fat` one) because using it needs compiler stuff etc.

sdetweil:
i disagree, its dependent on the electron we ship, not their fault, AND needs to be installed in the MM  node_modules folder. ive made postinstall scripts to help them out, but now the module name changed , again not their fault. 
if we had shipped it the module would have worked

every quarter we whack them AND the users 
stability matters

khassel:
please open an issue for this to discuss, I will not change this MR, this should be merged asap as is.

sdetweil:
will, do. didn't intend it to be added here, but whats the rush? we have a month

rejas (approved):
Still a month till next release indeed, but lets see if all is well with electron v35 in the meantime

PR #3737, opened 2025-03-06
add support for digital clock time color

sdetweil:
see https://forum.magicmirror.builders/topic/19440/digital-clock-minutes-darker

changelog added

question.. 

we have a config parm for seconds color, but not hour/minute
I have used css here.. is that too inconsistent?


modules/default/clock/clock.js
```diff
@@ -146,6 +146,8 @@ Module.register("clock", {
 		}
 
 		if (this.config.displayType !== "analog" && this.config.showTime) {
+			let ts = timeString.split(":");
+			timeString = `<span class="clock_hour">${ts[0]}</span>:<span class="clock_minute">${ts[1]}</span>`;
```
rejas, line 150: Yes, I personally would stay consistent and use a config param... Makes it easier to add to the docs and wouldnt confuse new users...

sdetweil: got it. will fix

sdetweil: well, those are used for the analog view ONLY.. AND there are classes added for hour and minute and seconds already
```
			clockHour.id = "clock-hour";
			clockHour.style.transform = `rotate(${hour}deg)`;
			clockHour.className = "clock-hour";
```
seconds in analog
```
			if (this.config.displaySeconds) {
				const clockSecond = document.createElement("div");
				clockSecond.id = "clock-second";
				clockSecond.style.transform = `rotate(${second}deg)`;
				clockSecond.className = "clock-second";  //<-   classname
				clockSecond.style.backgroundColor = this.config.secondsColor;
				clockFace.appendChild(clockSecond);
			}
```

none of the classes  are used for digital view 

so I will add usage of them for digital view AND add parms for config value an use in both views..

khassel:
> weather test failed

this one weather test seems to fail very often, don't know why ... (restarted it)

sdetweil:
will update doc, to add 2 parms, hoursColor, minutesColor (like secondsColor) 
only used if changed,
added css example, (commented out)

KristjanESPERANTO:
Just a thought: Instead of adding new options to the config for the hour and minute color. What do you think about marking `secondsColor` as deprecated and just moving the color (for hours, minutes and seconds) into the css file? If someone wants a different color, they can change the custom.css. Or am I missing some advantage of defining this in the config?

I hope I'm not annoying you as I often ask such questions 😅

sdetweil:
not annoying at all. more eyes and questions improve the results

deprecated as in not documented, replaced by css
(need still defined for mmm-config type config options, else lost on save) i am ok with

i had css only before this push, but asked for guidance  on consistency.

sdetweil:
weather failed again

khassel:
> weather failed again

you have the rights to restart (failed) tests

sdetweil:
yes i know. waiting til we come to consensus on which way to go

KristjanESPERANTO:
> not annoying at all. more eyes and questions improve the results

Nice :smiley: 

> deprecated as in not documented, replaced by css

Yes. And maybe if this option is set, a deprecation warning so we could after some releases may remove this option at all.

> i had css only before this push, but asked for guidance on consistency.

Yes, it should be consistent, either everything via config options or only via css.

sdetweil:
>Yes, it should be consistent, either via config options or via css.

its both before my changes now

sdetweil:
we could only do the deprecation message with lint as the module is in browser only. so
messages go to browser log which is hardly ever reviewed

KristjanESPERANTO:
> we could only do the deprecation message with lint as the module is in browser only.

We could also do it in the `check_config.js` :slightly_smiling_face:

KristjanESPERANTO:
I just had a closer look. It turned out that in `check_config.js` it wouldn't be so easy, but I noticed that we already have a `checkDeprecatedOptions` function in the `app.js` (to check deprecated options in the core) :bulb:. So I just extended it to check for deprecated options in the modules.

Checkout this commit in my testing branch:
https://github.com/KristjanESPERANTO/MagicMirror/commit/f34f2b8df76c75742da6459650275cc1c0efe9cd

What do you think?

sdetweil:
@rejas @khassel  what do you think, 
i like the idea of moving to css

rejas:
Both solutions have its dis/advanteges. 
A config option doesnt require much knowledge of the user in css
A css styling option is much cleaner in the code 

So I dont mind going the css route, depreacting the config options and adding the css styling as a new option into the docs of the module

sdetweil:
@KristjanESPERANTO   I want to add an entry to the clock_styles.css for the three classes..but empty

eslint rejects that.. 

what is a good approach to bypass this error?

the reason for adding is that there are already classes for the analog clock arms, 
and these are different names.. 
analog
clock-minute,hour, second

there is already a set of classes set on the div around the time
timeWrapper.className = "time bright large light";

digital
```
.module.clock .clock-hour-digital {
  
}
.module.clock .clock-minute-digital {

} 
.module.clock .clock-second-digital {

}
```

KristjanESPERANTO:
@sdetweil I took a look at it and made a few changes. I've set a PR on your branch so that we don't have to talk in theory for too long. What do you think of it? https://github.com/sdetweil/MagicMirror/pull/57

sdetweil:
now, can we write a testcase to prove the two spans instead of the time string. i hadnt made it there yet

sdetweil:
i submitted a pr to  documentation for this deprecated proprty and added info on css

modules/default/clock/clock.js
```diff
@@ -23,7 +23,7 @@ Module.register("clock", {
 		analogFace: "simple", // options: 'none', 'simple', 'face-###' (where ### is 001 to 012 inclusive)
 		analogPlacement: "bottom", // options: 'top', 'bottom', 'left', 'right'
 		analogShowDate: "top", // OBSOLETE, can be replaced with analogPlacement and showTime, options: false, 'top', or 'bottom'
-		secondsColor: "#888888",
+		secondsColor: "#888888", // DEPRECATED, use CSS instead. Class "clock-second-digital" for digital clock, "clock-second" for analog clock.
```
rejas, line 26: seems it is not only deprecated but not usable anymore?

sdetweil: will fix, Kristian moved to css only

KristjanESPERANTO: Sorry for the messup, thanks Sam for reactivating.

CHANGELOG.md
```diff
@@ -15,7 +15,10 @@ planned for 2025-04-01
 
 ### Added
 
+- Added Arabic translation.
```
rejas, line 18:
```suggestion
```

rejas (changes requested):

PR #3747, opened 2025-03-20
update required node version and dependencies

khassel:
see discussion in https://github.com/MagicMirrorOrg/MagicMirror/pull/3746

As [electron v35 requires node js v22.14.0](https://www.electronjs.org/blog/electron-35-0) we should update this here too.


sdetweil:
yep, will update scripts

rejas (approved):

Issue #3749, opened 2025-03-22
[Feature Request] Introduction of a graphical backend


KristjanESPERANTO:
I would like to suggest a feature that could improve the accessibility and usability of MagicMirror²: the introduction of a **graphical backend**. I would like to hear your opinions on this.

My suggestion is that we try to list the advantages and disadvantages of such a backend here. To then weigh up whether we want to go down this route.

Perhaps we could also implement a module system for the backend, just like in the frontend. Then most important features like module management and configuration would be in the core - and the rest could be done by 3rd party modules.

Of course, all of this would be a huge project. That's why I could also imagine that we first agree on the introduction of module management (via a graphical backend). A feature that should definitely be part of the core.

Once this is implemented, we could then discuss whether we should extend the backend further.

Here's a list of arguments and features I've spontaneously collected. Please let me know if there are any important points missing. But we don't have to go into too much detail yet, I'm mainly interested in summarising the discussion (which has already flared up in various places) and its arguments in one place.

We could certainly also be inspired by other good approaches/features from MMM-GoogleAssistant, MMM-Remote-Control, mmpm or MMM-Config.

## Disadvantages

- Reducing the learning effect for users
- Increased complexity
- Increased maintenance effort (requires additional maintenance, such as ensuring compatibility, updating features, fixing bugs and writing docs ...)

## Advantages

- Accessibility: A graphical backend would make the platform more user-friendly, especially for non-coding users who might struggle with manual file editing or terminal commands.
- Reduced user errors: By automating certain processes through a graphical backend, common mistakes made by users during manual configuration could be eliminated. This would lead to fewer support requests and issue reports.
- Efficiency: Tasks like module installation, configuration, and system diagnostics would become faster and more streamlined.
- Remote administration: A graphical backend would simplify the ability to manage MagicMirror² remotely over a LAN connection.

### Must have features

- Module Management (install, uninstall and configure modules ...)
- System Control (restart or shut down MagicMirror² or the entire system ...)

### Other features

- Display system information (such as versions, RAM usage, CPU load, and disk space ...)
- Access and review MagicMirror² logs for troubleshooting.
- Open a terminal directly from the graphical backend for advanced operations.

khassel:
I'm not sure if we should start such a project. You already mentioned some disadvantages and some projects which are doing similar things.

Right out of my head:
[...]

PR #3764, opened 2025-04-13
Use "node --run" instead of "npm run"

KristjanESPERANTO:
This has the advantage that the package manager is no longer involved after the installation process.

However, previous start commands such as `npm run start` continue to work. So we don't even have to adapt the documentation.


sdetweil:
what advantage is that for the user?

they just want the thing to run

khassel:
I see it as cleanup and restriction to the real use cases:
- npm only for install/update dependencies
- node for running scripts

There is no advantage or disadvantage for the user.

sdetweil:
so why add more confusion over what to do when?

KristjanESPERANTO:
I see the arguments against it and it's okay if we decide against it for now. 

But for the user I see also these advantages:

1. it is more intuitive for users. Why should I use a package manager to start a script?

2. it reduces complexity. Which is helpful for understanding and also for troubleshooting, as the package manager no longer plays a role after installation. The user may therefore be able to solve their problem themselves more quickly or receive help more quickly.

sdetweil:
the users dont know what a package manager is,
we have successfully graduated to they dont care about the technicalities of the base. witness the success  of the docker deployment(almost due to update issues for modules). they just want it to run!

they want to create an info panel. 90% of that content isn't ours at all. 

lets focus on making it run with as few crazy options as possible.  my objective is to make manual install,  
upgrade  and configuration including modules, go away

KristjanESPERANTO:
>  the users

I think we have a very different picture of who “the users” are and what they want.

> the users dont know what a package manager is,

It doesn't take magic to learn it.

> they just want it to run!

But simple consumers are not our main target group. As I interpret our manifesto, we want to inspire people to learn, try things out, break stuff, create things and get involved 💥

If we confuse users when they come across both options, so much the better, then they will question it and learn. It's not as if the change would be a breaking change.

And if a user just wants to be a simple consumer, he can use a third party installation script and a module manager such as mmpm or even buy a system on Amazon.

In any case, a simple user won't care whether we use “npm run” or node --run".

> my objective is to make manual install, upgrade and configuration including modules, go away

The question is whether we need to do this in the core? The discussion about a graphical backend in the core (#3749) did not go in that direction, for example.

-----

I think it would be good if we looked at the change mainly from a technical point of view. And as I understand it, there are fewer arguments against it than for it. But it is like a polish, the car will run without it and if there are others who think we shouldn't do it, I can live without this "polish" 🙂

rejas (approved):

rejas:
I like this polish (although I can also understand the "dont change if it aint broke" mentality) and it seems to be the way forward. 
Still would like to get either @khassel or @sdetweil approval before merging it.

sdetweil:
i disagree with this pr. we have all kinds of other problems to solve without this unnecessary 'polish'

bugsounet:
For once, i agree with @sdetweil

khassel:
As I already stated above:

> There is no advantage or disadvantage for the user.

I think we should use `node` wherever possible. For me, it's cleaner not to use a package manager to launch scripts.

khassel (approved):

rejas:
Ok, so I will merge it. Sams arguments are noted, but I agree with the other two more. lets hope node doesnt prove us wrong.

KristjanESPERANTO:
Thanks guys! And if it turns out to be a wrong/bad decision, I will be happy to help reverse it. Nothing is set in stone.

PR #3774, opened 2025-05-08
update dependencies

khassel:
- tried electron v36 but [electron-rebuild-tests are failing](https://github.com/MagicMirrorOrg/MagicMirror/actions/runs/14916012897/job/41901667578)
- is dependabot still working?


khassel:
after clicking the "check for updates" button we have now PR's again ...

khassel:
that is an error in my update process (I remove pm2 from the docker images so I'm always missing the updates). I'm to lazy to fix this, I think this can be merged and then we can merge the pm2-dependabot-PR for pm2 ...

khassel:
as @rejas is not responding please merge @KristjanESPERANTO

PR #3783, opened 2025-05-10
[linter] Review linter setup

KristjanESPERANTO:
- [2b395b9f] Fix command to lint markdown in `CONTRIBUTING.md`
  - [fed4c86c] Re-activate JSDoc ESLint plugin and fix linting issues (As far as I remember, we deactivated it when we upgraded to ESLint 9 because it was not compatible with it. But its now.)
  - [a3d2064b] Refactor ESLint config to use `defineConfig` and `globalIgnores` ([these are like new defaults](https://eslint.org/blog/2025/03/flat-config-extends-define-config-global-ignores/))
  - [a3d2064b] Replace `eslint-plugin-import` with `eslint-plugin-import-x` (https://github.com/es-tooling/module-replacements/blob/main/docs/modules/eslint-plugin-import.md)
  - [86a185b6] Switch Stylelint config to flat format and simplify Stylelint scripts (like we already did for ESLint and prettier)
  - [f5a2c541] Fix some typos


bughaver:
@KristjanESPERANTO would this be a good pr to add the eol setting to `prettier.config.mjs`? Without this setting on windows it defaults to `lf`.

setting to add would be: `endOfLine: "crlf",`

khassel (approved):

PR #3787, opened 2025-05-18
fix `roboto.css` to avoid error message

khassel:
`Unknown descriptor 'var(' in @font-face rule.` in firefox console

Issue was found by @KristjanESPERANTO 

I updated the `roboto.css` ~ 2 years ago with https://github.com/MagicMirrorOrg/MagicMirror/pull/3121 by using their css contents. Meanwhile they changed their contents again so we should follow.


KristjanESPERANTO:
The warning message no longer appears. Thank you!

PR #3791, opened 2025-05-26
add error handling to weather fetch functions, including cors

### LINKED ISSUE #3687: Weather modules don't update/grow stale

**Platform**: Raspberry Pi 4.

**Node Version**: 22.

**MagicMirror² Version**: latest

**Description**: Weather forecast (pirate weather) doesn't update without relaunching or reloading the page. Today, Tomorrow and the next day are all wrong after one day (today is yesterday, tomorrow is today). I don't think the values are updating either. Believe the current weather / temperature is wrong after it ages as well as it didn't update for a day.

**Steps to Reproduce**: Use Pirate Weather with this config:

```
	{
			module: "weather",
			position: "top_right",
			config: {
				weatherProvider: "pirateweather",
				type: "current",
				lat: 39.132721,
				lon: -84.628098,
				apiKey: "---",
			}
		},
		{
			module: "weather",
			position: "top_left",
			header: "Weather Forecast",
			config: {
				weatherProvider: "pirateweather",
				type: "forecast",
				lat: 39.132721,
				lon: -84.628098,
				apiKey: "---",
			}
		}
```

**Expected Results**: Live data.

**Actual Results**: Stale data

**Configuration**: See above

**Additional Notes**: Recently switched to Pirate Weather. Don't see any errors in the console.

sdetweil:
add error detection and handling for weather and cors fetch 

fixes #3687

### DIFF
--- a/modules/default/utils.js
+++ b/modules/default/utils.js
@@ -17,20 +17,26 @@ async function performWebRequest (url, type = "json", useCorsProxy = false, requ
 		requestUrl = url;
 		request.headers = getHeadersToSend(requestHeaders);
 	}
-	const response = await fetch(requestUrl, request);
-	const data = await response.text();
 
-	if (type === "xml") {
-		return new DOMParser().parseFromString(data, "text/html");
-	} else {
-		if (!data || !data.length > 0) return undefined;
+	let response = await fetch(requestUrl, request);
+	if (response.ok) {
+		const data = await response.text();
 
-		const dataResponse = JSON.parse(data);
-		if (!dataResponse.headers) {
-			dataResponse.headers = getHeadersFromResponse(expectedResponseHeaders, response);
+		if (type === "xml") {
+			return new DOMParser().parseFromString(data, "text/html");
+		} else {
+			if (!data || !data.length > 0) return "null"; //undefined;
+
+			const dataResponse = JSON.parse(data);
+			if (!dataResponse.headers) {
+				dataResponse.headers = getHeadersFromResponse(expectedResponseHeaders, response);
+			}
+			return dataResponse;
 		}
-		return dataResponse;
+	} else {
+		throw new Error(`Response status: ${response.status}`);
 	}
+
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ planned for 2025-07-01
 - [calendar] fix fullday event rrule until with timezone offset (#3781)
 - [feat] Add rule `no-undef` in config file validation to fix #3785 (#3786)
 - [fonts] Fix `roboto.css` to avoid error message `Unknown descriptor 'var(' in @font-face rule.` in firefox console (#3787)
+- [weather] add error handling to fetch functions including cors
 
 ### Updated
 
--- a/js/server_functions.js
+++ b/js/server_functions.js
@@ -41,25 +41,29 @@ async function cors (req, res) {
 			url = `invalid url: ${req.url}`;
 			Log.error(url);
 			res.send(url);
-		} else {
-			url = match[1];
+			return;
+		}
+		url = match[1];
 
-			const headersToSend = getHeadersToSend(req.url);
-			const expectedReceivedHeaders = geExpectedReceivedHeaders(req.url);
+		const headersToSend = getHeadersToSend(req.url);
+		const expectedReceivedHeaders = geExpectedReceivedHeaders(req.url);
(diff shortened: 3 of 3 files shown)


modules/default/utils.js
```diff
@@ -17,20 +17,26 @@ async function performWebRequest (url, type = "json", useCorsProxy = false, requ
...
 		request.headers = getHeadersToSend(requestHeaders);
 	}
-	const response = await fetch(requestUrl, request);
-	const data = await response.text();
 
-	if (type === "xml") {
-		return new DOMParser().parseFromString(data, "text/html");
-	} else {
-		if (!data || !data.length > 0) return undefined;
+	let response = await fetch(requestUrl, request);
```
rejas, line 21: can still be a const, right?

```suggestion
	const response = await fetch(requestUrl, request);
```

sdetweil:
all tests run locally, electron and e2e

khassel:
> all tests run locally, electron and e2e

the 2 failed tests are unit tests - and they fail if I run them locally ...

jbaboval:
Running this patch locally with success thus far.... Thanks!

modules/default/utils.js
```diff
@@ -17,20 +17,26 @@ async function performWebRequest (url, type = "json", useCorsProxy = false, requ
...
+		const data = await response.text();
 
-	if (type === "xml") {
-		return new DOMParser().parseFromString(data, "text/html");
-	} else {
-		if (!data || !data.length > 0) return undefined;
+		if (type === "xml") {
+			return new DOMParser().parseFromString(data, "text/html");
+		} else {
+			if (!data || !data.length > 0) return "null"; //undefined;
```
KristjanESPERANTO: Return `"null"` as a string looks unusual. Can you explain why you did this and why there is a comment `// undefined`?

sdetweil: Looking at suggested error recovery, the undefined prior result (comment) is faulty
Leading to a secondary error masking the first (no data) 

And null in quotes is the recommended change

I don't understand that (or undefined) either, as none of these are JSON , or error (throw)

sdetweil: and looking at the provider that calls the performWebRequest (here looking at pirateweather) 
```
	fetchWeatherForecast () {
		this.fetchData(this.getUrl())
			.then((data) => {
				if (!data || !data.daily || !data.daily.data.length) {  // undefined or null (no quotes) 
					// No usable data?
					return;  // <----- I think this bypasses the finally, so the provider is waiting forever, hung
				}

				const forecast = this.generateWeatherObjectsFromForecast(data.daily.data);
				this.setWeatherForecast(forecast);
			})
			.catch(function (request) {   // <--- maybe throw is a better choice
				Log.error("Could not load data ... ", request);
			})
			.finally(() => this.updateAvailable());
```

sdetweil: this is the fetchData in the providers layer
```
	async fetchData (url, type = "json", requestHeaders = undefined, expectedResponseHeaders = undefined) {
		const mockData = this.config.mockData;
		if (mockData) {
			const data = mockData.substring(1, mockData.length - 1);
			return JSON.parse(data);
		}
		const useCorsProxy = typeof this.config.useCorsProxy !== "undefined" && this.config.useCorsProxy;
		return performWebRequest(url, type, useCorsProxy, requestHeaders, expectedResponseHeaders, config.basePath);
	}
```

sdetweil:
Every other provider had to be fixed as well

KristjanESPERANTO:
@sdetweil I worked on a solution and pushed it directly into your branch. I hope that's okay(?) Unfortunately, I was not able to provoke the error before and after this PR, but the tests are now running again. Can you check if the error occurs with it?

sdetweil:
I saw. Will try to test today. I had forced the error path
Was trying not to change the providers

KristjanESPERANTO:
@sdetweil What do you think? Can we merge this?

sdetweil:
let me do some testing, been pretty busy lately

KristjanESPERANTO:
No problem. Take your time 🙂

KristjanESPERANTO (approved):

KristjanESPERANTO:
I have rebased this PR because I still consider it a beneficial change :slightly_smiling_face: However, since I have added changes myself, I do not like to merge it myself. Could someone else please take another look at it?

KristjanESPERANTO:
Thanks @khassel for your approval. Let's go with that. Thanks @sdetweil for that improvement! :smiley:

PR #3792, opened 2025-05-26
[tests] Review and refactor translation tests

KristjanESPERANTO:
I have refactored the translations tests, they should now be clearer and easier to understand. There should be no functional impact.

I have discarded the original approach of also replacing `XMLHttpRequest` with `fetch` in the file `js/translator.js`. I had managed to get it to work functionally, but I couldn't get the tests to work.


KristjanESPERANTO:
The function looks okay, but the tests just didn't work. I've gone a bit deep in the rabbit hole and have refactored a lot of the translation tests. The unit tests are now green locally, but I still need to validate that they are still checking the right things. The e2e are still an issue.

That's why this PR is set to draft for now.

khassel:
sorry, missed your "mark as ready" ...

PR #3795, opened 2025-05-29
simplify install and maintaining dependencies

khassel:
I was always unhappy when maintaining dependency updates to have 3 `package.json` files.

This PR moves all deps into the main `package.json` and removes the folders `fonts` and `vendor`.

If accepted I will update the docs too.


KristjanESPERANTO (approved):

rejas:
I wonder what the original Intention of this  seperation was. Maybe @MichMich can explain?

khassel:
> I wonder what the original Intention of this seperation was. Maybe @MichMich can explain?

well, I wouldn't remember...

Digging in the old sources the directories `fonts` and `vendor` were first filled with needed files but without `package.json`. So I think they decided later to use npm to get the files and put a `package.json` in the existing folders.

Other point is that only the directories `fonts` and `vendor` were published by the web server. In my first approach here I published the whole `node_modules` dir which seems not be good practice.

With the last commit I changed this by publishing only the needed subfolders of `node_modules` so now nothing more is published as before this PR.

khassel:
@rejas any objections to merge this?

We have 4 weeks to test until release and can fix/revert if needed.

Not merging in time will produce merge conflicts here which I would like to avoid.

KristjanESPERANTO:
Maybe removing this from `.gitignore` was a bit to early.

```
fonts/node_modules/**/*
vendor/node_modules/**/*
```

I just updated my system working on develop branch, and with this commit git didn't ignores the files in these two directories anymore. So I had over 3000 unstaged files 😅. With the next release the users which will update would have the same problem. With a new install their is no issue. Either we should re-add these two lines for a while or somehow manage to delete the directories during update.

khassel:
maybe best to add them back. Other idea would be a postinstall with `rm -rf vendor && rm -rf fonts`.

khassel:
> maybe best to add them back

would not be sufficient, we would have to add

```bash
fonts/
vendor/
```

I'm for deleting in postinstall so this is cleaned up on the user side.

What do you prefer @KristjanESPERANTO @rejas @sdetweil ?

sdetweil:
why didn't delete and git rm remove the folder altogether?  it worked for installers

KristjanESPERANTO:
I am okay with both options. Removing the directories would be the cleaner option, but two more lines in `.gitignore` wouldn't hurt us either.

sdetweil:
why wouldnt we remove folders

KristjanESPERANTO:
> why wouldnt we remove folders

This is how I understand it: When git pulls the commit, it should usually delete the two directories `fonts` and `vendor` because they aren't in the GitHub repo, but since locally there is a subdirectory (`node_modules`), it can not delete them.

khassel:
opened https://github.com/MagicMirrorOrg/MagicMirror/pull/3805 , we should continue the discussion there...

PR #3804, opened 2025-06-01
update dependencies

khassel:
and add now required "type" to `package.json`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ planned for 2025-07-01
 
 ### Updated
 
-- [core] Update dependencies incl. electron to v36 (#3774, #3788)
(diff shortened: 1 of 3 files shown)


rejas (approved):

bugsounet:
```
▲ [WARNING] "" is not a valid value for the "type" field [package.json]

    ../../package.json:25:9:
      25 │   "type": "",
         ╵           ~~

  The "type" field must be set to either "commonjs" or "module".
```

I just see this on my installer
related to  `"type": "",` of `package.json` file

Maybe `commonjs` is needed in our case ?

khassel:
> Maybe commonjs is needed in our case ?

I think so because "module" did not work, will try to fix it with next dependency update ...

PR #3805, opened 2025-06-01
remove existing folders fonts and vendor from local installations

khassel:
see discussions [here](https://github.com/MagicMirrorOrg/MagicMirror/pull/3795#issuecomment-2927804743).

The used command should work under windows too. I'm open to better solutions.


sdetweil:
why do you need a postinstall?

just remove folders 
and git rm folders

they should be gone after pull

KristjanESPERANTO (approved):
Looks good to me. I just successfully tested it on my system (Debian 12).

khassel:
> they should be gone after pull

all files under git control are gone but not the ignored files in `node_modules`. You can test it locally.

khassel:
are there any objections not to merge this?

rejas:
No objections from my side :-)

PR #3806, opened 2025-06-03
Refactored calendarfetcherutils to fix many of the timezone and DST related issues and make debugging way easier

plebcity:
Refactored calendarfetcherutils to remove as many of the date conversions as possible and use moment tz when calculating recurring events, this will make debugging a lot easier and fixes problems from the past with offsets and DST not being handled properly. Also added some tests to test the behavior of the refactored methodes to make sure the correct event dates are returned.

Refactored calendar.js aswell to make sure the unix UTC start and end date of events are properly converted to a local timezone and displayed correctly for the user.

This PR relates to: https://github.com/MagicMirrorOrg/MagicMirror/issues/3797


modules/default/calendar/calendar.js
```diff
@@ -1,5 +1,7 @@
 /* global CalendarUtils */
 
+const moment = require("moment-timezone");
```
KristjanESPERANTO: Can we use `require` here? `calender.js` is loaded by the browser and not by node, right?

plebcity: Ah this must have been my IDE doing this automatically, I will remove it

sdetweil: no, you have to load the script w getscripts() response to use its global methods . 

first version loaded wins

plebcity:
There is something broken, I will try to fix it tomorrow. I now have an ics calendar to test all cases locally so that should make it easier to find the bug. I took a quick look and it looks like some events might not have a timezone in the start field (probably full day events)

sdetweil:
correct, full day events are always considered to be local time

today, wherever you are

there can be local time events as well without timezone

all of the test cases use ics files from the mock folder with specific clock dates and times set in the testcase runner

sdetweil:
have you tested w modules that use the calendar broadcast

MMM-CalendarExt3 (there are multiple modules in the family)
MMM-MonthlyCalendar

i think there are more.
if the data changed, they are broken 

im guessing they will be broken as you changed to calendar ui side too

plebcity:
> have you tested w modules that use the calendar broadcast
> 
> MMM-CalendarExt3 (there are multiple modules in the family) MMM-MonthlyCalendar
> 
> i think there are more. if the data changed, they are broken
> 
> im guessing they will be broken as you changed to calendar ui side too

No I haven't tested those yet. They could be broken because the timestamps of startDate and endDate are now always in UTC timezone and they use the proper unix timestamp instead of the format('x') which adds 3 extra 0's at the end for milliseconds. Locally everything works fine with my own calendar's now but can't seem te figure out why electron shows errors.

plebcity:
> have you tested w modules that use the calendar broadcast
> 
> MMM-CalendarExt3 (there are multiple modules in the family) MMM-MonthlyCalendar
> 
> i think there are more. if the data changed, they are broken
> 
> im guessing they will be broken as you changed to calendar ui side too

Do you think it's important to stay backwards compatible with millisecond timestamps instead of using the second based unix timestamp? If so then I will change the return timestamps back to milliseconds and change the calendar.js to first convert it back to unix.

plebcity:
@sdetweil All tests have passed and I have made the necessary changes to make this backwards compatible with the old event broadcast, so timestamps are milliseconds and strings again. This will hopefully make it much easier to work with and more robust to edge cases in timezones. 
Also after this PR is merged I will create an issue to migrate magicmirror to Luxon since moment and moment-timezone is in maintenance mode right now. I thought it was too much of change to introduce Luxon to magicmirror right now since moment and moment-timezone were already being used.

tests/unit/modules/default/calendar/calendar_fetcher_utils_spec.js
```diff
@@ -49,5 +52,90 @@ describe("Calendar fetcher utils test", () => {
...
+			const moments = CalendarFetcherUtils.getMomentsFromRecurringEvent(data["67e65a1d-b889-4451-8cab-5518cecb9c66"], moment(), moment().add(365, "days"));
+
+			const januaryFirst = moments.filter((m) => m.format("MM-DD") === "01-01");
+			const julyFirst = moments.filter((m) => m.format("MM-DD") === "07-01");
+
+			expect(januaryFirst[0].toISOString(true)).toContain("09:00:00.000+01:00");
+			expect(julyFirst[0].toISOString(true)).toContain("09:00:00.000+02:00");
+		});
+
+		it("debug", () => {
```
KristjanESPERANTO: ESLint complains that this test has no assertion. Shall we allow this and disable the ESLint rule for this test?

sdetweil: I think it should be changed to make the assertion

plebcity: Ah sorry, this was left there by accident

plebcity:
> Good work 👍
> I recommend opening a general issue for replacing 'moment' because Luxon is not the only option. We should talk about all options before we decide.

Thanks! I created a new issue for it with a broader scope: https://github.com/MagicMirrorOrg/MagicMirror/issues/3808

tests/electron/modules/calendar_spec.js
```diff
@@ -153,19 +166,6 @@ describe("Calendar module", () => {
          * RRULE TESTS:
          * Add any tests that check rrule functionality here.
          */
-	describe("sliceMultiDayEvents", () => {
```
rejas, line 156: why is that removed?

modules/default/calendar/calendarfetcherutils.js
```diff
@@ -161,218 +176,47 @@ const CalendarFetcherUtils = {
...
-					return;
-				}
-
 				const location = event.location || false;
 				const geo = event.geo || false;
 				const description = event.description || false;
-				let d1;
-				let d2;
 
+				// TODO This should be a seperate function.
```
rejas, line 206: so why not do it in this PR?

modules/default/calendar/calendarfetcherutils.js
```diff
@@ -382,52 +226,52 @@ const CalendarFetcherUtils = {
...
+									recurringEventEndMoment = moment(curEvent.end).tz(CalendarFetcherUtils.getLocalTimezone());
+								}
 							} else {
 								Log.debug("recurrence key ", dateKey, " doesn't match");
 							}
 						}
 						// If there's no recurrence override, check for an exception date.  Exception dates represent exceptions to the rule.
 						if (curEvent.exdate !== undefined) {
-							Log.debug("have datekey=", dateKey, " exdates=", curEvent.exdate);
+							console.log("have datekey=", dateKey, " exdates=", curEvent.exdate);
```
rejas: should use Log.

tests/electron/modules/calendar_spec.js
plebcity: Because as you can see below this test is the exact same test, so it was duplicate and therefor I removed it

modules/default/calendar/calendarfetcherutils.js
plebcity: I felt like this was too much of a change right now so I left a TODO for somebody else to fix in the future or whenever I have some more time to fix this myself. What I meant with this TODO is change the entire flow of the createEventList method to seperate recurring events from single events. Then you probably also want to extract the push to newEvents aswell.

modules/default/calendar/calendarfetcherutils.js
plebcity: Oops, well spotted. I will check if I left any other console.log statements or other debug lines in there

rejas (approved):

modules/default/calendar/calendarfetcherutils.js
KristjanESPERANTO: With PR #3810 ESLint will complain about the use of `console` in future PRs :-)

PR #3794, opened 2025-05-28
[l10n] Complete translations

KristjanESPERANTO:
**Short description**: I completed the translations with the help of translation tools.

**Long description**: I'm currently looking at the translation-tests and I noticed the e2e-test `${language} should contain all base keys, ()`. It is supposed to check whether all keys are present in each translation file, but it doesn't actually work that way. When I fix it, a lot of missing translations are reported. I have completed these translations with the help of translation tools and packed them into this PR. I have left out the modified test so that we can focus on the question if we accept such amount of automatic translations or not.

rejas has already expressed in another PR (https://github.com/MagicMirrorOrg/MagicMirror/pull/3775#discussion_r2083099252) that he prefers human translators. I basically do too, but I don't see how we can manage to have all translations completed by humans. And even if a few translations are not correct, hopefully a user will get in touch.

If we decide against the translations, we should at least remove the test. If we go for the tranlsations, I'll add the test.

What do you think?

PR #3809, opened 2025-06-07
[tests] Fix and refactor e2e test "Same keys" in "translations_spec.js"

KristjanESPERANTO:
While working on the translations (in #3792 and #3794) I realised that the e2e "Same keys" tests were not working entirely. There was also a TODO entry for this in the test, as well as a try-catch-block workaround. I therefore fixed and refactored it.

I also sorted the translations in `translations/translations.js`, so that the test outputs are alphabetically.


khassel (approved):

KristjanESPERANTO:
@khassel thanks for approving! Would you like to merge it or do we need another reviewer? A merge would be helpful for me to continue on #3792.

PR #3813, opened 2025-06-20
add missing gitignore part (fonts/vendor node_modules)

bugsounet:
just add some missing gitignore path
 * /fonts/node_modules/
 * /vendor/node_modules/

Change log is not needed for this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,7 @@ js/positions.js
 # Ignore lock files other than package-lock.json
 pnpm-lock.yaml
 yarn.lock
+
+# Ignore fonts and vendor node_modules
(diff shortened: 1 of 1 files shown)


khassel:
this should be already solved with https://github.com/MagicMirrorOrg/MagicMirror/pull/3805

bugsounet:
I have this on my client:

image

khassel:
when you e.g. switch from `master` to `develop` you have to run `npm run install-mm` or `npm install`, the folders are removed in the `postinstall`.

Tested this again and it works on my side.

bugsounet:
yes, you have right :)

PR #3817, opened 2025-06-23
fixes e2e tests

khassel:
- fix newsfeed and calendar e2e tests so they don't need `--forceExit` anymore
- `--forceExit` should stay in test runs to avoid running until test limit in case of errors
- remove mocking console, not needed anymore
- configFactory-stuff should not run in browser (otherwise we get errors `[ReferenceError: exports is not defined]`)


modules/default/calendar/calendarfetcher.js
```diff
@@ -81,10 +81,13 @@ const CalendarFetcher = function (url, reloadInterval, excludedEvents, maximumEn
 	 * Schedule the timer for the next update.
 	 */
 	const scheduleTimer = function () {
-		clearTimeout(reloadTimer);
-		reloadTimer = setTimeout(function () {
-			fetchCalendar();
-		}, reloadInterval);
+		if (process.env.JEST_WORKER_ID === undefined) {
```
rejas, line 84: could we refactor this (and the similar code in newsfeedfetcher) into a method in modules/utils.js?

khassel: I think "we" should not do this in this PR (and better not before next release).

`modules/default/utils.js` is not the right place, the code inside this file is only used in browser, the above fetchers are used on server side.

khassel: I opened https://github.com/MagicMirrorOrg/MagicMirror/issues/3819

KristjanESPERANTO (approved):

rejas (approved):

PR #3832, opened 2025-07-02
Fix calendar js timestamp conversion

plebcity:
This fixes a bug in the calendar.js method where the local timezone unix timestamp is converted to a moment. And it also fixes the broken unittest on 01-01 and 01-07.

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -586,12 +586,13 @@ Module.register("calendar", {
 	},
 
 	/**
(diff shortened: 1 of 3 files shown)


plebcity:
@khassel @rejas I hope we can get this fix released as a hotfix for the july 1st release since this has now broken the recurring events for me.

plebcity:
Ah wait a second, my local files haven't all updated yet. I will check if this fix is actually needed in calendar.js. Otherwise I will remove that fix and just fix the unit test

Issue #3819, opened 2025-06-24
[Change Request] refactor default modules


khassel:
### What problem do you want to solve with this change?

As discussed at https://github.com/MagicMirrorOrg/MagicMirror/pull/3817, there is similar code in the newsfeed and calendar fetchers that needs to be cleaned up (setTimeout).

We need a place where we can consolidate the code used in both.

There are certainly other things that need to be improved.

### What do you think is the correct solution?

to be discussed

### Participation

- [ ] I am willing to submit a pull request for this change.

khassel:
closing this, if someone has ideas for more improvements we can reopen.

PR #3837, opened 2025-07-08
refactor default modules: move scheduleTimer to one place

khassel:
see #3819

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Thanks to: @dathbe.
 ### Changed
 
 - [clock] Add CSS to prevent line breaking of sunset/sunrise time display (#3816)
+- [core] Enhance system information logging format and include additional env and RAM details (#3839)
 - [refactor] Add new file `js/module_functions.js` to move code used in several modules to one place (#3837)
 
 ### Updated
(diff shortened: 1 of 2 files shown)


KristjanESPERANTO:
I would like to test something before merging, but don't have the time now.

khassel:
> I would like to test something before merging, but don't have the time now.

o.k. for me, nothing urgent in this PR ...

KristjanESPERANTO:
Oh, I actually wanted to push in my repo :facepalm: Sorry! But with the last commit you can see my approach. I would like to avoid the additional use of `module-alias` and even drop it in the medium term. It prevents us from running MM under other runtimes like deno or bun (see #2934). If you don't like my approach, we can undo the commit :slightly_smiling_face:

KristjanESPERANTO (approved):

khassel:
> Oh, I actually wanted to push in my repo 🤦 Sorry! 

ups ... no problem ... I'm always afraid of doing this, so I added a separate (fake) push url for the original mm repo ...

> But with the last commit you can see my approach. I would like to avoid the additional use of module-alias and even drop it in the medium term. It prevents us from running MM under other runtimes like deno or bun (see https://github.com/MagicMirrorOrg/MagicMirror/pull/2934). If you don't like my approach, we can undo the commit 🙂

LGTM (but it's over my js knowledge ...)

So as we both worked on this PR @rejas should merge it.

KristjanESPERANTO:
Thanks for being understanding.

> I added a separate (fake) push url for the original mm repo

That's a cool idea. I just did this on my setup :slightly_smiling_face:

rejas (approved):

PR #3862, opened 2025-08-27
fix for #3380, socket.io timeout closure

### LINKED ISSUE #3380: default calendar losing events between backend and front end
see

https://forum.magicmirror.builders/topic/18325/calendar-not-updating/85?_=1706893005707

I have added  traces to just before the socketsend from node helper and traces to the front end, and the event shows before send, and does not show after socket receive

the event is NOT at the start or end of the data

how else can I debug this?

we just did a new npm install, on 2.26, with node-ical@0.16.1

sdetweil:
socket.io times out and closes the client side socket without any callback
sendSocntNotification() from the server side data is lost as the socket is closed.  but the client doesn't know

increase the timeout 

fixes #3380

### DIFF
--- a/js/server.js
+++ b/js/server.js
@@ -42,7 +42,9 @@ function Server (config) {
 					origin: /.*$/,
 					credentials: true
 				},
-				allowEIO3: true
+				allowEIO3: true,
+				pingInterval: 120000, // server → client ping every 2 mins   //add
+				pingTimeout: 120000 // wait up to 2 mins for client pong   //add
 			});
 
 			server.on("connection", (socket) => {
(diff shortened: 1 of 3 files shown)


js/server.js
```diff
@@ -42,7 +42,9 @@ function Server (config) {
 					origin: /.*$/,
 					credentials: true
 				},
-				allowEIO3: true
+				allowEIO3: true,
+				pingInterval: 120000, // server → client ping every 2 mins   //add
+				pingTimeout: 120000 // wait up to 2 mins for client pong   //add
```
KristjanESPERANTO: What is `//add` for?

sdetweil: removed, was from form instructions to users

PR #3869, opened 2025-09-01
Update feels_like temperature formula

rejas:
The current logic never showed any different temperature than the current one, so I looked into a more "explained" formula and found one in the HomeAssistant forums.

### DIFF
--- a/modules/default/weather/providers/openmeteo.js
+++ b/modules/default/weather/providers/openmeteo.js
@@ -370,8 +370,6 @@ WeatherProvider.register("openmeteo", {
 		 */
 		const h = moment().hour();
 		const currentWeather = new WeatherObject();
-		Log.info(weather);
-		Log.info(h);
 
 		currentWeather.date = weather.current_weather.time;
 		currentWeather.windSpeed = weather.current_weather.windspeed;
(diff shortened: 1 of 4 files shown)


modules/default/weather/providers/openmeteo.js
```diff
@@ -370,6 +370,8 @@ WeatherProvider.register("openmeteo", {
 		 */
 		const h = moment().hour();
 		const currentWeather = new WeatherObject();
+		Log.info(weather);
+		Log.info(h);
```
KristjanESPERANTO: Do we want these logs here?

modules/default/weather/providers/openmeteo.js
```diff
@@ -381,13 +383,15 @@ WeatherProvider.register("openmeteo", {
 		currentWeather.maxTemperature = parseFloat(weather.daily[0].temperature_2m_max);
 		currentWeather.weatherType = this.convertWeatherType(weather.current_weather.weathercode, currentWeather.isDayTime());
 		currentWeather.humidity = parseFloat(weather.hourly[h].relativehumidity_2m);
-		currentWeather.feelsLikeTemp = parseFloat(weather.hourly[h].apparent_temperature);
+		//currentWeather.feelsLikeTemp = parseFloat(weather.hourly[h].apparent_temperature);
```
KristjanESPERANTO: Can we remove this line?

modules/default/weather/providers/openmeteo.js
rejas: no :-) accidentally added them

modules/default/weather/providers/openmeteo.js
rejas: no, but we uncomment it :-)

KristjanESPERANTO:
I haven't checked the math, but it looks good. Thanks for the improvement! 🙂

PR #3886, opened 2025-09-10
add test for config:check

khassel:


KristjanESPERANTO:
Very good :slightly_smiling_face:

rejas:
should this test fail when the config is erroneous?

KristjanESPERANTO:
Currently, the test only checks if the config check runs without issues, regardless of whether the config is faulty or not. I'm pretty sure that was the original intention after this correction: https://github.com/MagicMirrorOrg/MagicMirror/pull/3885.

But yes, testing the config check with a good and a faulty config would certainly be even better.

PR #3887, opened 2025-09-11
Add Prettier plugin for Nunjuck templates

rejas:
While looking into the weather module I got errors when trying to commit:

```
✖ prettier --write:
[error] No parser could be inferred for file "/Users/rejas/Projekte/MagicMirror/modules/default/weather/current.njk".
modules/default/weather/weather.css 15ms (unchanged)
```

So I looked into a plugin for prettier that allows formatting njk files and here is the PR for it.

Some noise due to never having them formatted with prettier but in the long run this should be helpful.


KristjanESPERANTO:
Nice :+1:

PR #3888, opened 2025-09-11
chore: use `prettier --write --ignore-unknown` in `lint-staged` to avoid errors on unsupported files

KristjanESPERANTO:
This prevents `prettier` from failing when `lint-staged` passes unknown/binary files, making the pre-commit hook more robust.

In concrete terms this could happen, when we, for example, add a new PNG file. Since we rarely do this, it has not been noticed so far.

This is recommended when using asterisk: https://github.com/lint-staged/lint-staged#automatically-fix-code-style-with-prettier-for-any-format-prettier-supports

## before

```bash
$ npx lint-staged <-- after staging a new PNG file
✔ Backed up original state in git stash (c3247d4b)
✔ Hiding unstaged changes to partially staged files...
⚠ Running tasks for staged files...
  ❯ package.json — 2 files
    ❯ * — 2 files
      ✖ prettier --write [FAILED]
    ↓ *.js — no files
    ↓ *.css — no files
↓ Skipped because of errors from tasks.
↓ Skipped because of errors from tasks.
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ prettier --write:
[error] No parser could be inferred for file "~/MagicMirror/test.png".
...
```

## after

```bash
$ npx lint-staged <-- after staging a new PNG file
✔ Backed up original state in git stash (0c3fe285)
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Cleaning up temporary files...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ Thanks to: @dathbe.
   - Avoid potential port conflicts by using port 3001 for translator unit tests
   - Improve test reliability and maintainability
 - [tests] add alert module tests for different welcome_message configurations (#3867)
-- [lint-staged] use `prettier --write --ignore-unknown` in `lint-staged` to avoid errors on unsupported files
+- [lint-staged] use `prettier --write --ignore-unknown` in `lint-staged` to avoid errors on unsupported files (#3888)
 
 ### Updated
 


KristjanESPERANTO:
By the way: It's actually somewhat good that we didn't have it before. Otherwise we wouldn't be formatting the `njk` files now – see #3887 :smiley:

rejas:
> By the way: It's actually somewhat good that we didn't have it before. Otherwise we wouldn't be formatting the `njk` files now – see #3887 😃

thats the only downside of this PR: we would need to manually run prettier to find these un-linted file extension

PR #3891, opened 2025-09-15
chore: update dependencies

KristjanESPERANTO:
Since receiving a security warning about `axios` in `node-ical` (https://github.com/MagicMirrorOrg/MagicMirror/security/dependabot/70), I've created [a pull request to remove `axios` from `node-ical`](https://github.com/jens-maus/node-ical/pull/397) — it was accepted and we now have a new version :slightly_smiling_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ Thanks to: @dathbe.
 ### Updated
 
 - [calendar] Update defaultSymbol name and also the link to the icon search site (#3879)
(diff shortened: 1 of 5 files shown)


KristjanESPERANTO:
I think the tests are failing because of the `jsdom` update. I'm searching for a solution :slightly_smiling_face:

KristjanESPERANTO:
`jsdom` v27 broke animation-name assertions in `tests/e2e/animateCSS_spec.js`. I replace them with class polling. And additionally added a negative assertion. Test green again :smiley:

khassel (approved):

PR #3892, opened 2025-09-16
[weather] better null value handling for weather type

rejas:
As mentioned [here](https://github.com/MagicMirrorOrg/MagicMirror/pull/3878#issuecomment-3275344406) our weather module needs a bit better handling for null values in the type field.

This pr adds this and cleans up the layout a little.

### DIFF
--- a/modules/default/weather/current.njk
+++ b/modules/default/weather/current.njk
@@ -59,7 +59,9 @@
         {% endif %}
       </span>
     {% endif %}
-    <span class="light wi weathericon wi-{{ current.weatherType }}"></span>
+    {% if config.weatherType %}
+      <span class="light wi weathericon wi-{{ current.weatherType }}"></span>
+    {% endif %}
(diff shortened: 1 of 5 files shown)


modules/default/weather/current.njk
```diff
@@ -59,7 +59,9 @@
         {% endif %}
       </span>
     {% endif %}
-    <span class="light wi weathericon wi-{{ current.weatherType }}"></span>
+    {% if config.weatherType %}
```
KristjanESPERANTO: Is `config.weatherType` right? Should it be `current.weatherType`?

rejas: good catch, fixed and added a test for it. should go back to bed, not in a state for coding....

KristjanESPERANTO: Yes, being tired and coding is not the best combination. I know that too 😪

PR #3893, opened 2025-09-16
refactor: replace `module-alias` dependency with internal alias resolver

KristjanESPERANTO:
- removes the external unmaintained `module-alias` dependency -> reducing complexity and risk
- introduces a small internal alias mechanism for `logger` and `node_helper`
- preserves backward compatibility for existing 3rd‑party modules
- should simplify a future ESM migration of MagicMirror

I'm confident that it shouldn't cause any problems, but we could also consider including it in the release after next. What do you think?

This PR is inspired by PR #2934 - so thanks to @thesebas! :bow: :smiley:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ planned for 2026-01-01
 
 ### Changed
 
+- [core] refactor: replace `module-alias` dependency with internal alias resolver (#3893)
+
 ### Fixed
 
 ### Updated
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,7 @@
+const aliasMapper = {
+	logger: "<rootDir>/js/logger.js"
+};
+
 const config = {
 	verbose: true,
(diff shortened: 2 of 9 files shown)


khassel:
> I'm confident that it shouldn't cause any problems, but we could also consider including it in the release after next. What do you think?

cool, I will do some tests on my side and let you know ...

khassel:
LGTM!

rejas:
> I'm confident that it shouldn't cause any problems, but we could also consider including it in the release after next. What do you think?

I am very much in favor of merging this after the next release...

KristjanESPERANTO:
> I am very much in favor of merging this after the next release...

Okay, that's perfectly fine for me :slightly_smiling_face:

khassel:
so I will merge this when conflicts are resolved ...

KristjanESPERANTO:
I'm on it.

khassel (approved):

PR #3900, opened 2025-09-26
refactor(calendar): improve recurring event handling clarity and robustness

KristjanESPERANTO:
- enhance comments for better understanding of timezone conversions and recurrence logic
- replace `arrow functions` with `named functions` for improved readability
- compact `dtstart` normalization code to reduce duplication
- add conditional `tzid` reset to prevent errors on missing options
- snap all-day recurrences to event timezone midnight using `startOf("day")` to fix date drift

The trigger for all this is that I am currently rebuilding node-ical to [get rid of the dependency on moment-timezone](https://github.com/jens-maus/node-ical/pull/404), and I wanted to better understand how we use it here.


sdetweil (commented):
Remember to test with modules that receive our notifications

MMM-CalendarExt3xxxx
For sure

KristjanESPERANTO:
@sdetweil Thanks :slightly_smiling_face:  I'll keep it in mind and test whether this would break `MMM-CalendarExt3xxxx`.

But I think I messed up a bit here - seems I wanted to fix things that doesn't need to be fixed :sweat_smile: I need to take deeper look at it.

sdetweil:
Yeh, it’s easy to say, I wouldn’t do it like this 
And then end up busting everything

KristjanESPERANTO:
Okay. I'll drop it for now.

Issue #3912, opened 2025-10-10
[Bug] Default Weather Module - envcanada provider is no longer pulling data from Environment Canada


Crazylegstoo:
### Environment

MagicMirror² version: 2.33.0
Node version: 22.18.0
npm version: 10.9.3
Platform: Raspberry Pi 4 - Bookworm

### Which start option are you using?

node --run start

### Are you using PM2?

Yes

### Module

weather

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

Validated there are no issues in logs and that Config.js is correct. The problem is definitely specific to the envcanada provider in the Weather module

### What did you expect to happen?

Verify that Environment Canada provider is not working properly

### What actually happened?

Verified that Environment Canada data is not providing data to Weather module

### Additional comments

The envcanada provider in the default Weather module was fixed in MM v2.33.0 to use a new URL hierarchy that Environment Canada implemented to access weather data for Canadian locations. Subsequent to this provider update, Environment Canada has implemented one further update to their URL hierarchy to make it easier to access 'current day' weather data. Here are the old and new URL formats with the updated in **bold**:

_Old:_ https://dd.weather.gc.ca/citypage_weather/...

_New:_ https://dd.weather.gc.ca/**today/**citypage_weather/...

As a result of this additional URL update, the envcanada provider is, once again, not able to pull weather data. The net effect is that the Weather module will display outdated information or will display 'loading' in place of weather data.

Solution: The envcanada provider code will need to be updated (again) to incorporate the URL format change. 

### Participation

- [x] I am willing to submit a pull request for this change.

KristjanESPERANTO:
@Crazylegstoo I don't see anyone working on the problem. Are you preparing a PR? 🙂

sdetweil:
We have 70 days til next release and they have broken the api  two sep times. I vote for patience

Crazylegstoo:
> [@Crazylegstoo](https://github.com/Crazylegstoo) I don't see anyone working on the problem. Are you preparing a PR? 🙂

Yup! Got a bit sidetracked this week but I'm about to prep a PR. The coding change is minor and has been running on my MM for a week or so without issue. I've also bundled in a small update to catch a windspeed value of 'calm' and translate it to a numeric value.

PR #3919, opened 2025-10-18
Fix for envcanada Provider to use updated Env Canada URL

Crazylegstoo:
The envcanada provider in the default Weather module was fixed in MM v2.33.0 to use a new URL hierarchy that Environment Canada implemented to access weather data for Canadian locations. Subsequent to this provider update, Environment Canada has implemented one further update to their URL hierarchy to make it easier to access 'current day' weather data. Tis change was raised in Issue #3912  as a Bug, which is addressed in this Provider update. There are no Magic Mirror UI changes from this update.

The fix is to add one additional element to the URL used to access Environment Canada XML-based weather data.

This PR is also taking the opportunity to make one further small fix to how windspeed is handled in this Provider. Most of the time, Env Canada provides an expected numeric value. There are instances, however, where the value provided is 'calm', which the Weather module does not expect. The Provider code has been changed to detect a 'calm' windspeed and convert it to '0' for the purposes of the Weather module. Note that in the world of weather/climate analysis, a windspeed of 'calm' is used as a synonym for a windspeed of 0.

Note that a ChangeLog entry is included in this PR.

### DIFF
--- a/modules/default/weather/providers/envcanada.js
+++ b/modules/default/weather/providers/envcanada.js
@@ -243,7 +243,7 @@ WeatherProvider.register("envcanada", {
 		} else {
 			currentWeather.temperature = this.cacheCurrentTemp;
 		}
-		
+
 		if (ECdoc.querySelector("siteData currentConditions wind speed").textContent === "calm") {
 			currentWeather.windSpeed = "0";
 		} else {


modules/default/weather/providers/envcanada.js
```diff
@@ -244,7 +244,12 @@ WeatherProvider.register("envcanada", {
 			currentWeather.temperature = this.cacheCurrentTemp;
 		}
 
-		currentWeather.windSpeed = WeatherUtils.convertWindToMs(ECdoc.querySelector("siteData currentConditions wind speed").textContent);
+		if (ECdoc.querySelector("siteData currentConditions wind speed").textContent === "calm") {
```
rejas, line 247: i wouldnt be suprised if there were more windspeeds like this. is there something in the docs of the env canada api about this?

Crazylegstoo: The doc is saying to expect a numeric windspeed value. I reached out to Env Canada and they are 'investigating', so I'm kind of expecting the 'calm' value is a bug. 

FWIW, 'calm' is an official description for a windspeed that is (effectively) 0 on the Beaufort Scale. As Beaufort values increase, each increase has it's own textual description that describes a windspeed range (whereas Beaufort 0 is not a range). I've done some checking, and I cannot find any other instances where the Env Canada provides a textual windspeed for any winds above 0 - i.e. they provide a numeric value in those instances. 

I would keep the 'calm' check as-is at this point. If Env Canada decides to fix/change their system to eliminate 'calm' in favour of a numeric value, it will likely be months down the road before that would make it into production (and I should get a notification if they plan to make a change). In that scenario, we could make a further change to the Provider code to remove the extraneous 'calm' check.

KristjanESPERANTO (approved):

rejas (approved):

PR #3920, opened 2025-10-19
feat(core): add `server:watch` script with automatic restart on file changes

jboucly:
## Description

This PR adds a new `server:watch` script that runs MagicMirror² in server-only mode with automatic restart and browser reload capabilities.

Particularly helpful for:
- **Developers** who need to see changes immediately without manual restarts.
- **Users setting up their mirror** who make many changes to `config.js` or `custom.css` and need quick feedback.

### What it does

When you run `npm run server:watch`, the watcher monitors files you specify in `config.watchTargets`. Whenever a monitored file changes:

1. The server automatically restarts
2. Waits for the port to become available
3. Sends a reload notification to all connected browsers via Socket.io
4. Browsers automatically refresh to show the changes

This creates a seamless development experience where you can edit code, save, and see the results within seconds.

### Implementation highlights

**Zero dependencies:** Uses only Node.js built-ins (`fs.watch`, `child_process.spawn`, `net`, `http`) - no nodemon or external watchers needed.

**Smart file watching:** Monitors parent directories instead of files directly to handle atomic writes from modern editors (VSCode, etc.) that create temporary files during save operations.

**Port management:** Waits for the old server instance to fully release the port before starting a new one, preventing "port already in use" errors.

### Configuration

Users explicitly define which files to monitor in their `config.js`:

```js
let config = {
  watchTargets: [
    "config/config.js",
    "css/custom.css",
    "modules/MMM-MyModule/MMM-MyModule.js",
    "modules/MMM-MyModule/node_helper.js"
  ],
  // ... rest of config
};
```

This explicit approach keeps the implementation simple (~260 lines) while giving users full control over what triggers restarts. If `watchTargets` is empty or undefined, the watcher starts but monitors nothing, logging a clear warning message.

---

**Note:** This PR description has been updated to reflect the final implementation. During the review process, we refined the approach multiple times based on feedback.

### DIFF
--- a/serveronly/watcher.js
+++ b/serveronly/watcher.js
@@ -211,46 +211,6 @@ function watchFile (file) {
 
 startServer();
 
-// Watch the config file (might be in custom location)
-// Priority: MM_CONFIG_FILE env var, then global.configuration_file, then default
-const configFile = getConfigFilePath();
-watchFile(configFile);
-
-/**
- * Resolve the active custom CSS path based on config or environment overrides
- * @param {object} config The loaded MagicMirror config
- * @returns {string} Absolute path to the CSS file
- */
-function resolveCustomCssPath (config = {}) {
-	const cssFromEnv = process.env.MM_CUSTOMCSS_FILE;
-	let cssPath = cssFromEnv || config.customCss || "css/custom.css";
-
-	if (!cssPath || typeof cssPath !== "string") {
-		cssPath = "css/custom.css";
-	}
-
-	return path.isAbsolute(cssPath) ? cssPath : path.join(rootDir, cssPath);
-}
-
-/**
- * Determine fallback watch targets when no explicit watchTargets are provided
- * @param {object} config The loaded MagicMirror config (may be partial)
- * @returns {string[]} Array of absolute paths to watch
- */
-function getFallbackWatchTargets (config = {}) {
-	const targets = new Set();
-	if (configFile) {
-		targets.add(configFile);
-	}
-
-	const cssPath = resolveCustomCssPath(config);
-	if (cssPath) {
-		targets.add(cssPath);
-	}
-
-	return Array.from(targets);
-}
-
 // Setup file watching based on config
 try {
 	const configPath = getConfigFilePath();
@@ -260,9 +220,10 @@ try {
 	let watchTargets = [];
 	if (Array.isArray(config.watchTargets) && config.watchTargets.length > 0) {
 		watchTargets = config.watchTargets.filter((target) => typeof target === "string" && target.trim() !== "");
-	} else {
-		watchTargets = getFallbackWatchTargets(config);
-		Log.log("Watch targets not specified. Using active config and custom CSS as fallback.");
+	}
+
+	if (watchTargets.length === 0) {
+		Log.warn("Watch mode is enabled but no watchTargets are configured. No files will be monitored. Set the watchTargets array in your config.js to enable file watching.");
 	}
 
 	Log.log(`Watch mode enabled. Watching ${watchTargets.length} file(s)`);
@@ -289,17 +250,7 @@ try {
 	}
 } catch (err) {
 	// Config file might not exist or be invalid, use fallback targets
-	Log.warn("Could not load watchTargets from config, watching active config/custom CSS instead:", err.message);
-
-	for (const target of getFallbackWatchTargets()) {
-		if (!fs.existsSync(target)) {
-			Log.warn(`Fallback watch target does not exist: ${target}`);
-			continue;
-		}
-
-		watchFile(target);
-		Log.log(`Watching fallback file: ${target}`);
-	}
+	Log.warn("Could not load watchTargets from config.");
 }
 
 process.on("SIGINT", () => {
--- a/js/server_functions.js
+++ b/js/server_functions.js
@@ -181,6 +181,16 @@ function getEnvVars (req, res) {
  * @returns {string} The absolute config file path
  */
 function getConfigFilePath () {
+	// Ensure root_path is set (for standalone contexts like watcher)
+	if (!global.root_path) {
+		global.root_path = path.resolve(`${__dirname}/../`);
+	}
+
+	// Check environment variable if global not set
+	if (!global.configuration_file && process.env.MM_CONFIG_FILE) {
+		global.configuration_file = process.env.MM_CONFIG_FILE;
+	}
+
 	return path.resolve(global.configuration_file || `${global.root_path}/config/config.js`);
 }
 


sdetweil:
You can reload the browser side anytime, ctl-r, or f5

it only loads the web side. You have to stop server and start to reload node_helpers

sdetweil:
What does server dev do? dev was only for web side

if you open your own browser it doesn’t get that parm

sdetweil:
The scripted installer handles node install  plus other things

jboucly:
Yes, but stopping the server and then restarting it quickly becomes redundant when we make a lot of modifications and tests on the module side in development....

jboucly:
> You can reload the browser side anytime, ctl-r, or f5
> 
> it only loads the web side. You have to stop server and start to reload node_helpers

Yes, but stopping the server and then restarting it quickly becomes redundant when we make a lot of modifications and tests on the module side in development....


> What does server dev do? dev was only for web side
> 
> if you open your own browser it doesn’t get that parm

I don't understand

sdetweil:
So again, change module, ctrl-r if cursor is over web side, f5 if over dev window side

what does server:dev do?
The dev parm is only used when electron is the browser in full mode, npm run start:dev

in run server mode
You have to start your own browser. Chrome/firefox/surf/safari pointing at the MagicMirror address and port 

dev mode curl-shift-I , toggles the dev window open or closed

.nvmrc
```diff
@@ -0,0 +1 @@
+v22.20.0
```
KristjanESPERANTO: I like the intention behind this change, but I see problems for contributors. Pinning NVM to v22.20.0 conflicts with the broader engines range (>=22.20.0 <23 || >=24). Contributors who already run Node 24 or even 25 via nvm will silently be switched back to 22.20.0, which is confusing and makes testing against newer versions harder.

Do you have a solution for this?

package.json
```diff
@@ -43,6 +43,7 @@
 		"lint:prettier": "prettier . --write",
 		"prepare": "[ -f node_modules/.bin/husky ] && husky || echo no husky installed.",
 		"server": "node ./serveronly",
+		"server:dev": "nodemon",
```
KristjanESPERANTO: I like this intention too. But did you considered a built-in flow like `node --watch ./serveronly` instead of nodemon? It should automatically track whichever config or module directories folks actually load (including custom paths set via `MM_CONFIG_FILE` or `MM_MODULES_DIR`), so the auto-reload keeps working without hard-coding watch lists. It also avoids bringing in nodemon and its dependency tree while still supporting hot reload.

Watch mode (via `nodemon` or built in `--watch`) could even be used as the default for the server mode or rename `server:dev` to `server:watch`  - that should also be practical for normal users.

KristjanESPERANTO:
> what does server:dev do?

To start MagicMirror in server mode + Nodemon keeps an eye on server-side files, and whenever a change is detected the server restarts automatically.

sdetweil:
Ah

.nvmrc
jboucly: Ok i see a problem, i remove this ! 
Unfortunately nvmrc does not support a version range...

package.json
jboucly: Oh yes i see ! Nodemon actually has quite a few dependencies... I try to create `--watch` flag for `node --watch ./serveronly` 👌 

I'll ping you as soon as it's pushed.

.nvmrc
jboucly: done

package.json
jboucly: Done and PR description is changed

KristjanESPERANTO:
Thanks for the changes :slightly_smiling_face:  I would like to test something relevant. Give me some time for the review.

KristjanESPERANTO:
I have re-based the branch.

During my tests, I encountered various issues. For example, the restart was triggered by files that my editor created temporarily when saving. Furthermore, the restart did not wait until the old instance released the port, so the new instance could not use it. I have adjusted this and other details in one commit.

Please feel free to give feedback on my changes.

From my point of view, everything looks good now :smiley: However, I would wait for the opinion of at least one other contributor on the current status.

Edit: I removed the manual restart option since it would be another feature. One feature per PR is easier to handle.

jboucly:
> I have re-based the branch.
> 
> During my tests, I encountered various issues. For example, the restart was triggered by files that my editor created temporarily when saving. Furthermore, the restart did not wait until the old instance released the port, so the new instance could not use it. I have adjusted this and other details in one commit.
> 
> Please feel free to give feedback on my changes.
> 
> From my point of view, everything looks good now 😃 However, I would wait for the opinion of at least one other contributor on the current status.
> 
> Edit: I removed the manual restart option since it would be another feature. One feature per PR is easier to handle.

Nice jobs ! Thanks for the changes ! I see that I don't have the whole history of the project when I see all the changes you made 🫣

Ok for the removal of manual reload and sorry for my english in `changelog` file x')

But i don't understand why on your local you have this bug with my version of watcher 🤔 

Edit: I change the description of the PR

serveronly/watcher.js
```diff
@@ -0,0 +1,239 @@
...
+// Watch the config file (might be in custom location)
+// Priority: MM_CONFIG_FILE env var, then global.configuration_file, then default
+const configFile = getConfigFilePath();
+watchFile(configFile);
+
+// Watch core directories (modules, js and serveronly)
+// We watch specific directories instead of the whole project root to avoid
+// watching unnecessary files like node_modules (even though we filter it),
+// tests, translations, css, fonts, vendor, etc.
+watchDir(path.join(__dirname, "..", "modules"));
```
khassel: @KristjanESPERANTO 

Nitpicking stuff from my side ...

AFAIR only `modules/default` is fix, the modules path for 3rd-party can be changed in `config.js` (see `foreignModulesDir` defined in `defaults.js`). There is an environment variable `MM_MODULES_DIR` too. You can search for `modulesDir` in the project to see the construction ...

Other point: You excluded css in the comment above, but what about `custom.css`? If this file should be watched: Same construction as for modules, `customCss` in `defaults.js`, env var `MM_CUSTOMCSS_FILE` ...

So better approach here would be (untested):

```js
const { getEnvVarsAsObj } = require("#server_functions");
const env = getEnvVarsAsObj();
watchDir(path.join(__dirname, "..", env.modulesDir));
watchDir(path.join(__dirname, "..", env.customCss));
```

KristjanESPERANTO: > You excluded css in the comment above, but what about custom.css?

Hm... If the CSS changes and the watcher restarts the server, the frontend would not restart/reload. That's why I removed the CSS monitoring, as it would not make sense with the current logic.

> AFAIR only modules/default is fix, the modules path for 3rd-party can be changed in config.js (see foreignModulesDir defined in defaults.js). ...

Good point. That should be manageable.

----

I just had an idea for how to create a common watcher that would work in both server and electron mode and can reload the browser/electron. I'll give it a try, but that needs some refactoring.

I think a watch mode also for electron would definitely be nice. Don't you think? I'll try to look into that over the next few days.

khassel: I think the frontend will restart after server restart (with a little delay), AFAIR this was implemented some times ago. I use this in my setup (Browser gets refreshed when docker container was restarted).

Watching/Restarting is also possible with pm2 ...

KristjanESPERANTO: > I think the frontend will restart after server restart

Okay, I'll test it. My previous tests were probably too short.

> Watching/Restarting is also possible with pm2 ...

Or do you think a watch mode in the core is unnecessary? I find it kind of interesting to be able to do this without nodemon or pm2. But of course, we don't want the core to complex.

khassel: pm2 is already in this project (not happy with this because I remove this when building the container images) so was only meant as hint before reinventing the wheel ...

jboucly: Hence the fact that I proposed Nodemon to have everything already built without reinventing the wheel.
Nodemon already handles these reloading aspects for CSS, js, etc.

What I'm afraid of is that the `watcher.js` will become complex... Nodemon can be implemented in `devDependencies` and therefore not installed during a Prod installation?

KristjanESPERANTO: > You excluded css in the comment above, but what about custom.css? 

With the latest commit, CSS files can now be monitored too.

> I think the frontend will restart after server restart (with a little delay)

Previously, the watcher ran as a separate process outside server.js and had no access to Socket.io, so automatic browser reloads weren't possible. The latest commit changes that: the watcher now sends an HTTP request to `/reload`, which triggers the server to emit a `RELOAD` event via Socket.io to all connected clients, causing them to execute `window.location.reload(true)`.

> pm2 is already in this project (not happy with this because I remove this when building the container images) 

Good point - that's actually a strong argument *against* using pm2 for watch mode. Watch mode should also be helpful in containers, I think.

> Hence the fact that I proposed Nodemon to have everything already built without reinventing the wheel.

I understand the concern about reinventing the wheel, but I think this is one case where a small, purpose-built solution is better.

- **Zero additional dependencies** - uses only Node.js built-ins (fs, child_process, net, http)
- **MagicMirror-aware** - understands `MM_CONFIG_FILE`, `MM_CUSTOMCSS_FILE`, and custom paths - useful for defaults/fallback
- **Integrated browser reload** - leverages existing Socket.io to trigger client reloads
- **Small & maintainable** - ~300 lines of code

----

With the latest commit, I introduced a new config option `watchTargets` that allows explicit file monitoring:

```js
let config = {
  customCss: "css/custom_test.css",
  watchTargets: ["css/custom_test.css", "config/config_test.js", "modules/MMM-PublicTransportHafas/MMM-PublicTransportHafas.js"],
  address: "127.0.0.1",
  // ...
```

The previous approach of monitoring everything (including third-party modules, core files, while excluding node_modules, etc.) would have made the solution quite complex. This way, it stays compact and users can explicitly decide what they want to monitor/watch.

If `watchTargets` is not specified, the fallback monitors the active config file and active CSS file automatically.

What do you think about that?

@jboucly Sorry for bending your PR so much.

----

Even though I'm already proposing a very specific solution here, I still see us in the brainstorming stage. So if I'm heading in the wrong direction, please let me know.

jboucly: > What do you think about that?

I actually find that much better! And likewise, the aspect of reloading the front end directly during a change adds the touch that automatic end-to-end Hot Reload was missing!

I don't see any objections to any of the changes you suggested in my MR. I actually find the points relevant to the project ! And given that I don't have all the functional components of the project, your comments were more than relevant !

serveronly/watcher.js
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Get the server port from config
+ * @returns {number} The port number
+ */
+function getServerPort () {
+	if (serverPort) return serverPort;
+
+	try {
+		// Try to read the config file to get the port
+		const configPath = path.join(__dirname, "..", "config", "config.js");
```
khassel: hardcoded `config.js path`? Maybe [this part](https://github.com/MagicMirrorOrg/MagicMirror/blob/develop/js/app.js#L30) of `js/app.js` should be moved to `js/server_functions,js` and used in both places?

serveronly/watcher.js
```diff
@@ -0,0 +1,318 @@
...
+		delete require.cache[require.resolve(configPath)];
+		const config = require(configPath);
+		serverPort = global.mmPort || config.port || 8080;
```
khassel: copy/paste from `app.js` too ...

serveronly/watcher.js
```diff
@@ -0,0 +1,318 @@
...
+		server.once("error", () => {
+			resolve(false);
+		});
+
+		server.once("listening", () => {
+			server.close();
+			resolve(true);
+		});
+
+		server.listen(port, "127.0.0.1");
```
khassel: should this not be `config.address`? Above will not work in a container ...

serveronly/watcher.js
```diff
@@ -0,0 +1,318 @@
...
+	if (process.env.MM_CONFIG_FILE) {
+		return process.env.MM_CONFIG_FILE;
+	}
+
+	if (global.configuration_file && global.root_path) {
+		return path.resolve(global.root_path, global.configuration_file);
+	}
+
+	return path.join(__dirname, "..", "config", "config.js");
+}
```
khassel: see comment above, logic here copy/pasted from `app.js`, should be implemented in one place if possible.

Global question: Do we need to implement all edge cases for "watching"?

Another thing is the `config.js.template` thing, where the `config.js` is produced when starting mm, see [this code](https://github.com/MagicMirrorOrg/MagicMirror/blob/develop/js/app.js#L76)

serveronly/watcher.js
KristjanESPERANTO: Applied - `isPortAvailable()` now binds to `config.address` instead of 127.0.0.1. Since `config` isn't in scope, I cache it via `getServerConfig()` to align with the main server's bind address.

serveronly/watcher.js
KristjanESPERANTO: > should be moved to js/server_functions,js and used in both places?

Great idea! Done.

> hardcoded config.js path?

Fixed.

> Maybe this part of js/app.js should be moved to js/server_functions.js and used in both places?

As I understand it: The linked part sets `global.configuration_file` from the `MM_CONFIG_FILE` env var and must run early in `app.js` before anything else uses it. `getConfigFilePath()` already reads this global, so moving the initialization to `server_functions.js` would create circular dependencies. The watcher doesn't need this initialization logic since it just consumes the already-configured path.

serveronly/watcher.js
KristjanESPERANTO: Oh well, thank you for your thoroughness! I changed it.

serveronly/watcher.js
KristjanESPERANTO: > see comment above, logic here copy/pasted from app.js, should be implemented in one place if possible.

Fixed - centralized `getConfigFilePath()` in `server_functions.js`.

> Global question: Do we need to implement all edge cases for "watching"?

After centralizing the config logic in `server_functions.js` this is no question any more, right?

> Another thing is the config.js.template thing ...

I don't understand the statement right now, sorry. Are you suggesting we should also watch `config.js.template`  and `config.env` to trigger restarts?

khassel: don't know if we should implement watching of template. Problem I see is that no `config.js` exists (or an old one) when using the watcher and using `config.js.template` because the `config.js` is produced in `app.js` in this case. I can live with this ...

serveronly/watcher.js
```diff
@@ -0,0 +1,310 @@
...
+	});
+}
+
+/**
+ * Send reload notification to all connected clients
+ */
+function notifyClientsToReload () {
+	const { port } = getServerConfig();
+	const options = {
+		hostname: "localhost",
```
khassel: serverConfig.address instead of localhost?

serveronly/watcher.js
```diff
@@ -0,0 +1,310 @@
...
+const configFile = getConfigFilePath();
+watchFile(configFile);
+
+/**
+ * Resolve the active custom CSS path based on config or environment overrides
+ * @param {object} config The loaded MagicMirror config
+ * @returns {string} Absolute path to the CSS file
+ */
+function resolveCustomCssPath (config = {}) {
+	const cssFromEnv = process.env.MM_CUSTOMCSS_FILE;
```
khassel: what about using `getEnvVarsAsObj` from `server_functions.js` here?

serveronly/watcher.js
KristjanESPERANTO: Ah! In that case users could set the template files to be watched instead of the config.js. Llike this:

```js
watchTargets: ["config/config.js.template", "config/config.env"]
```

KristjanESPERANTO: Or we don't watch anything at all as a fallback/default. Users must explicitly define what is being watched themselves, which would make the code a lot simpler :slightly_smiling_face:

KristjanESPERANTO: I like the idea of "Dropping the fallback". Less code to maintain. We could give good examples in the docs instead.

khassel: my point is
- the watcher uses `config.js` and starts mm
- while starting mm creates a new `config.js` from the template

so in some cases these configs could differ!

But as the watcher only uses address and port from config I think this is no real problem.

serveronly/watcher.js
KristjanESPERANTO: I tried swapping in `getEnvVarsAsObj`, but without success - it seems there is no simple way to make it work. To avoid the crash, we would have to preload the entire config just for the watcher, add guards for the global state, and so on. That would require far more lines of code than this single direct `process.env` lookup.

serveronly/watcher.js
KristjanESPERANTO: Changed :slightly_smiling_face:

khassel:
From my side this can be merged. Will wait until tomorrow to give others a chance to comment/review.

serveronly/watcher.js
KristjanESPERANTO: > my point is ...

Oh, I was quite on the wrong track... 

> I think this is no real problem.

Yes, you're right, that shouldn't be a problem.

> while starting mm creates a new config.js from the template

But if mm changes `config.js` and it is in the `watchTargets`, we would end up in an endless restart loop. So template users have to adapt the `watchTargets`.

KristjanESPERANTO: @khassel And what do you think of "Dropping the fallback* to simplify the code"?

*fallback = watching the config and css files in use

khassel: removing fallback would be o.k. for me. I see this watcher as a function for developers and they should be able to set the watchTargets themselves ...

KristjanESPERANTO: Okay, this saves us almost 50 lines in the watcher and is also easier to document. I implemented this and discovered and solved another problem in the process: `getConfigFilePath()` was called before globals were initialized, causing it to return invalid paths (`undefined/config/config.js`). This worked previously because the fallback logic called `getConfigFilePath()` a second time inside the try-block where globals were already set by the running server. Fixed by making `getConfigFilePath()` self-initializing when globals are missing.

khassel (approved):

PR #3921, opened 2025-10-19
chore: update dependencies

KristjanESPERANTO:
Normally, I wouldn't update the dependencies again so soon, but `node-ical` underwent some major changes (see https://github.com/jens-maus/node-ical/pull/404) with the last release, and I'd like to use it here as early as possible to see if there are any problems with it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1537,64 +1537,6 @@
 				"semver": "bin/semver.js"
 			}
 		},
-		"node_modules/@emnapi/core": {
-			"version": "1.5.0",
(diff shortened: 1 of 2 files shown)


khassel:
The new patch version of jsdom does not work anymore with all tests using jsdom, I saw this already in my tests today, maybe you have a solution for this.

Edit: Problem seems to be the new major release v8 of [parse5](https://github.com/inikulin/parse5/releases/tag/v8.0.0)

KristjanESPERANTO:
I'll take a look at it too.

KristjanESPERANTO:
To move forward, I have pinned the old version of `parse5` for now. I haven't found another elegant solution yet, but we should resolve this sooner or later.

PR #3923, opened 2025-10-20
[logger] Add prefixes to most Log messages

rejas:


khassel:
@rejas I'm not good in nodejs so I'm asking if there maybe is a mechanism to automatically provide the filename into the log message instead of hard coding this everywhere?

sdetweil:
on the web side 
document.currentScript.src
and on the node side 
__filename

khassel:
yes, but I was thinking about something where the `logger.js` could get the caller filename ...

sdetweil:
yeh I have looked at that a number of times in different places, we need a macro that can capture that info in the calling code and pass it to the worker function.

rejas:
Having it done autoamtically would of course be nice, but maybe we just merge this, create a ticket with "PR welcome" for that idea and see who takes it?

PR #3925, opened 2025-10-20
[core] Update deps and pin jsdom to v27.0.0

rejas:


KristjanESPERANTO:
We already pinned `parse5` (#3921). It would certainly be sufficient to pin one (either `jsdom` or `parse5`), but it doesn't hurt either. We just have to remember to remove the pinning in both places when there is a solution :slightly_smiling_face:

PR #3926, opened 2025-10-21
logger: add calling filename as prefix on server side

khassel:
follow up to discussions in #3923 

This small change adds the calling filename (path und suffix are stripped) to the log messages coming from server side, example screenshot:

grafik

I choosed grey as color for better readability.

Let me know if this change is wanted and if the duplicates in the default module log messages should stay or removed.

Links:
- [console-stamp: create custom token](https://github.com/starak/node-console-stamp?tab=readme-ov-file#create-a-custom-token)

### DIFF
--- a/js/animateCSS.js
+++ b/js/animateCSS.js
@@ -132,7 +132,7 @@ function addAnimateCSS (element, animation, animationTime) {
 	const node = document.getElementById(element);
 	if (!node) {
 		// don't execute animate: we don't find div
-		Log.warn("[animateCSS] node not found for adding", element);
+		Log.warn("node not found for adding", element);
 		return;
 	}
 	node.style.setProperty("--animate-duration", `${animationTime}s`);
@@ -149,7 +149,7 @@ function removeAnimateCSS (element, animation) {
 	const node = document.getElementById(element);
 	if (!node) {
 		// don't execute animate: we don't find div
-		Log.warn("[animateCSS] node not found for removing", element);
+		Log.warn("node not found for removing", element);
 		return;
 	}
 	node.classList.remove("animate__animated", animationName);
--- a/js/check_config.js
+++ b/js/check_config.js
@@ -44,7 +44,7 @@ function checkConfigFile () {
 	}
 
 	// Validate syntax of the configuration file.
-	Log.info(`[checkconfig] Checking config file ${configFileName} ...`);
+	Log.info(`Checking config file ${configFileName} ...`);
 
 	// I'm not sure if all ever is utf-8
 	const configFile = fs.readFileSync(configFileName, "utf-8");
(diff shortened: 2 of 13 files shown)


sdetweil:
Not sure the MagicMirror code prefix (app,server…) is useful, and certainly not double  
99% are not as convoluted as calendar and newsfeed

khassel:
> Not sure the MagicMirror code prefix (app,server…) is useful

but do they hurt? I prefer to stay with this instead of implementing filters for removing.

> and certainly not double 99% are not as convoluted as calendar and newsfeed

yes, they should be removed. We could also think of renaming some files so it is clearer to what they belong, e.g. `git-helper`

sdetweil:
Doesn’t hurt.

KristjanESPERANTO:
> Let me know if this change is wanted 

I like it :+1: 

> and if the duplicates in the default module log messages should stay or removed.

You mean entries like `... [node_helper]  [nodehelper] ...`?

Removed.

----

What effect does that have on 3rd-party-modules?

khassel:
> You mean entries like ... [node_helper]  [nodehelper] ...?

yes, and e.g. `[update_helper] [updatenotification]`, `[git_helper] [updatenotification]`, `[newsfeedfetcher] [newsfeed]`, ...

> What effect does that have on 3rd-party-modules?

searching for a good module to test this. Testing with mine gave most time `node_helper` which is clear by construction.

Should have no negative effect, either a filename is displayed or otherwise `unknown` if an exception occurs.

khassel:
we could also extract the module name (from the path) for 3rd-party-modules instead of dislaying the filename which will be 95% `node_helper` for 3rd-party ...

sdetweil:
>we could also extract the module name, yes modulename filename (no path)

>yes, and e.g. [update_helper] [updatenotification], [git_helper] [updatenotification], [newsfeedfetcher] [newsfeed], ...
I haven't seen any 3rd party constructed like newsfeed or calendar, node_helper calls a lib, which calls a lib, which puts out log info

in that case the module name and current file name is really all thats important.. don't need to call tree
and I'd rather have it modulename, filename, then all the modulename lines could be together after the timestamp
sort on that column in the log file

KristjanESPERANTO:
Prefixes for modules would be a nice win. We'll have duplicates there too at first (because now some modules add own prefixes), but they'll gradually disappear.

khassel:
so this is my current state:

- get last directory level before filename
- if = `js` then use stripped filename, else use directory name
- 3rd-party-mods looking good, see MMM-Remote-Control and MMM-Flights in Screenshot

We can change this logic if we find things which are ugly or not working properly.

If this is accepted I will update this PR accordingly and remove all the hardcoded duplicates introduced with the last PR (all the stuff in `[...]` which is not grey).

grafik

KristjanESPERANTO:
Looks good to me! :slightly_smiling_face:

khassel:
final screenshot with removed duplicates:

grafik

Issue #3930, opened 2025-10-25
[Bug] weather provider openmeteo fails handling summer/winter time


khassel:
### Environment

MagicMirror² version: develop

### Which start option are you using?

node --run server

### Are you using PM2?

No

### Module

None

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

<details>
<summary>Configuration</summary>

```
                {
                        module: "weather",
                        header: "openmeteo forecast",
                        position: "top_right",
                        config: {
                                weatherProvider: "openmeteo",
                                type: "forecast",
                                appendLocationNameToHeader: false,
                                lat: 50.17659,
                                lon: 8.62685,
                                useCorsProxy: true,
                                roundTemp: true,
                                showPrecipitationProbability: true,
                                windUnits: "kmh",
                                fade: false,
                                forecastDateFormat: "dd HH",
                                absoluteDates: true
                        },
                },
                {
                        module: "weather",
                        header: "yr forecast",
                        position: "top_right",
                        config: {
                                weatherProvider: "yr",
                                type: "forecast",
                                appendLocationNameToHeader: false,
                                lat: 50.17659,
                                lon: 8.62685,
                                roundTemp: true,
                                windUnits: "kmh",
                                fade: false,
                                forecastDateFormat: "dd HH",
                                absoluteDates: true
                        },
                },
```
</details>

<img width="454" height="588" alt="Image" src="https://github.com/user-attachments/assets/87365565-6f6d-482e-a70e-8066e1e063fc" />

### What did you expect to happen?

openmeteo should display

SA 00
SU 00
MO 00
TU 00
WE 00

### What actually happened?

openmeteo displays

SA 00
SU 00
SU 23
MO 23
TU 23

### Additional comments

I added the same with yr provider which works. Because they use other hours they are not impacted by the openmeteo problem.

openmeteo data are linux timestamps:

<img width="495" height="240" alt="Image" src="https://github.com/user-attachments/assets/f7980374-8aa0-48f7-bb7a-d3fdb6fad195" />

At the moment it is not clear if this is a problem of openmeteo or a general problem of the weather module.

### Participation

- [x] I am willing to submit a pull request for this change.

PR #3931, opened 2025-10-25
fixes problems with daylight-saving-time in weather provider `openmeteo`

khassel:
fix for #3930


rejas:
how come this isnt a problem for the other providers?

khassel:
I don't know, I'm using only this one. It will depend how data is presented by the provider and how it is converted by us which is done in an own file for every provider.

khassel:
so can this be merged? If someone wants to look into the other providers he can create a ticket with "PR welcome" for this and see who takes it?  😜

PR #3940, opened 2025-11-02
[tests] migrate from `jest` to `vitest`

KristjanESPERANTO:
This is a big change, but I think it's a good move, as `vitest` is much more modern than `jest`.

I'm excited about the UI watch feature (run `npm run test:ui`), for example - it's really helpful and saves time when debugging tests. I had to adjust a few tests because they had time related issues, but basically we are now testing the same things - even a bit better and less flaky (I hope).

What do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
 		"eslint-plugin-vitest": "^0.5.4",
 		"express-basic-auth": "^1.2.1",
 		"husky": "^9.1.7",
-		"jsdom": "27.1.0",
+		"jsdom": "^27.1.0",
 		"lint-staged": "^16.2.6",
 		"markdownlint-cli2": "^0.18.1",
 		"playwright": "^1.56.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
 				"eslint-plugin-vitest": "^0.5.4",
 				"express-basic-auth": "^1.2.1",
 				"husky": "^9.1.7",
-				"jsdom": "27.1.0",
+				"jsdom": "^27.1.0",
 				"lint-staged": "^16.2.6",
 				"markdownlint-cli2": "^0.18.1",
 				"playwright": "^1.56.1",


khassel:
Give me some time to review, it's to late now ... 

I did the switch to jest some time ago, but I'm not bothered about which testing system we use here. Another advantage is that vitest can handle the current jsdom version ...

package.json
```diff
@@ -98,23 +100,25 @@
...
 		"eslint-plugin-import-x": "^4.16.1",
-		"eslint-plugin-jest": "^29.0.1",
 		"eslint-plugin-jsdoc": "^61.1.11",
 		"eslint-plugin-package-json": "^0.59.1",
+		"eslint-plugin-vitest": "^0.5.4",
 		"express-basic-auth": "^1.2.1",
 		"husky": "^9.1.7",
-		"jest": "^30.2.0",
-		"jsdom": "27.0.0",
+		"jsdom": "27.1.0",
```
khassel: can we change this back to `"^27.1.0"`? (here and in `package-lock.json`)

KristjanESPERANTO: Done :slightly_smiling_face:

khassel (approved):

KristjanESPERANTO:
@khassel Thanks for checking and accepting!

I'm curious to see how/if it will affect things in the future.

khassel:
Looking into the tests I see a lot of error and debug messages, any idea to get rid of them?

KristjanESPERANTO:
The verbose output is caused by this:

1.  **Vitest behavior** - Unlike Jest, Vitest doesn't automatically suppress console output by default
2. **Intentional test errors** - Many tests deliberately trigger error conditions (missing configs, invalid files, etc.) to verify error handling works correctly

I've created a PR (#3941) to improve the situation by setting `LOG_LEVEL=ERROR` in CI environments (in `tests/utils/vitest-setup.js`). This should eliminate most of the noise while keeping the intentional error messages visible.

PR #3941, opened 2025-11-03
[tests] suppress debug logs in CI environment + improve calendar symbol test stability

KristjanESPERANTO:
## CI Log Suppression

**Two-level approach for clean test output:**

1. **Suppress debug/info logs**: Call `logger.setLogLevel("ERROR")` in CI to hide verbose logging
2. **Suppress intentional error logs**: Set `mmTestMode` flag and check it before logging errors that are part of test assertions (e.g., testing error handling in `git_helper.js` and `server_functions.js`)

This keeps CI output clean and makes genuine failures immediately visible, while preserving full logging for local development.

**Before:** 1348 log lines with verbose debug/info output  
**After:** 168 log clean lines with only test results

## Calendar Symbol Test Stability

Convert the calendar symbol test from external URL (`calendarlabs.com`) to existing local mock file (`12_events.ics`). This eliminates CI timeouts caused by external dependencies and improves test reliability.

The test still validates the same symbol array feature but now runs faster and deterministically without network dependencies.

### DIFF
--- a/js/server_functions.js
+++ b/js/server_functions.js
@@ -58,7 +58,10 @@ async function cors (req, res) {
 			res.send(data);
 		}
 	} catch (error) {
-		Log.error(error);
+		// Only log errors in non-test environments to keep test output clean
+		if (process.env.mmTestMode !== "true") {
+			Log.error(error);
+		}
 		res.send(error);
 	}
 }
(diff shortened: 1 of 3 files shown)


KristjanESPERANTO:
The failing test looks like a Timing-Flake 🤔  I'm looking into it.

KristjanESPERANTO:
Test works now -> PR description updated -> still a lot of logging -> I'm investigating

KristjanESPERANTO:
Hold on. I'll take a look at the expected error messages so that we might get a clean output after all 🤓

KristjanESPERANTO:
We now have clean output - the expected error messages are no longer displayed in test mode :rocket:

khassel (approved):

PR #3942, opened 2025-11-03
test: respect GitHub ACTIONS_STEP_DEBUG for verbose test logging

KristjanESPERANTO:
When tests failed, I clicked “Enable debug logging” several times when rerunning them and was always disappointed that I couldn't find any more logs. This should fix this.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ planned for 2026-01-01
 - [calendar] test: remove "Recurring event per timezone" test (#3929)
 - [calendar] chore: remove `requiresVersion: "2.1.0"` (#3932)
 - [tests] migrate from `jest` to `vitest` (#3940, #3941)
+- [test] respect GitHub ACTIONS_STEP_DEBUG for verbose test logging (#3942)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


KristjanESPERANTO:
I triggered a re-run and it doesn't work as expected :thinking: I'll investigate.

KristjanESPERANTO:
I'm closing this PR because the approach doesn't work as expected.

**The Problem:**
The "Enable debug logging" feature in GitHub Actions sets `ACTIONS_STEP_DEBUG` and `ACTIONS_RUNNER_DEBUG` as **secrets for the GitHub Actions runner infrastructure**, but these variables are **not exposed as environment variables** to the shell processes running our tests. They only control GitHub's internal debug output (the `##[debug]` lines), not our application code.

**Alternative Approach:**
We could use a [repository variable](https://github.com/MagicMirrorOrg/MagicMirror/settings/variables/actions) like `ENABLE_VERBOSE_LOGGING` that can be set manually and accessed via `${{ vars.ENABLE_VERBOSE_LOGGING }}` in the workflow. However, this would require:
1. Manually setting the variable before re-running tests (not as elegant as a simple checkbox)
2. Remembering to rest it afterwards

Since this isn't as elegant as the intended "re-run with debug logging" checkbox solution, I'm abandoning this approach.

PR #3949, opened 2025-11-05
[calendar] Show `repeatingCountTitle` only if `yearDiff` > 0

jarnoml:
Current implementation shows 0 for non-repeating calendar events. This would fix it.

Hello and thank you for wanting to contribute to the MagicMirror² project!

before fix

after fix


KristjanESPERANTO:
Thanks for this PR! The change looks good to me :+1: 

Could you please add a CHANGELOG entry for this? 

And would you like to add a test?

jarnoml:
I added a CHANGELOG entry. Adding the test would require more time, as the test framework being used is unfamiliar to me.

KristjanESPERANTO:
Okay, we'll go with that. Thanks for this improvement! :smiley:

PR #3950, opened 2025-11-05
[core] refactor: replace `XMLHttpRequest` with `fetch` and migrate e2e tests to Playwright

KristjanESPERANTO:
#### Changes
- **translator.js**: Use `fetch` with `async/await` instead of XHR callbacks
- **loader.js**: Align URL handling and add error handling (follow-up to fetch migration)
- **Tests**: Update infrastructure for `fetch` compatibility

#### Benefits
- Modern standard API
- Cleaner, more readable code
- Better error handling and fallback mechanisms

### 2. Migrate e2e tests to Playwright

This wasn't originally planned for this PR, but is related. While investigating suspicious log entries which surfaced after the fetch migration I kept running into JSDOM’s limitations. That pushed me to migrate the E2E suite to Playwright instead.

#### Changes
- switch e2e harness to Playwright (`tests/e2e/helpers/global-setup.js`)
- rewrite specs to use Playwright locators + shared `expectTextContent`
- install Chromium via `npx playwright install --with-deps` in CI

#### Benefits
- much closer to real browser behaviour
- and no more fighting JSDOM’s quirks

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
 		"config:check": "node js/check_config.js",
 		"postinstall": "git clean -df fonts vendor",
 		"install-mm": "npm install --no-audit --no-fund --no-update-notifier --only=prod --omit=dev",
-		"install-mm:dev": "npm install --no-audit --no-fund --no-update-notifier",
+		"install-mm:dev": "npm install --no-audit --no-fund --no-update-notifier && npx playwright install chromium",
 		"lint:css": "stylelint 'css/main.css' 'css/roboto.css' 'css/font-awesome.css' 'modules/default/**/*.css' --fix",
 		"lint:js": "eslint --fix",
 		"lint:markdown": "markdownlint-cli2 . --fix",


tests/e2e/helpers/global-setup.js
```diff
@@ -64,40 +64,48 @@ exports.startApplication = async (configFilename, exec) => {
...
 			dom.window.name = "jsdom";
 			global.window = dom.window;
+			global.document = dom.window.document;
 			// Following fixes `navigator is not defined` errors in e2e tests, found here
 			// https://www.appsloveworld.com/reactjs/100/37/mocha-react-navigator-is-not-defined
 			global.navigator = {
 				useragent: "node.js"
 			};
 			dom.window.fetch = fetch;
 			dom.window.onload = () => {
```
khassel: I think the whole `dom.window.onload` funktion can be removed (?)

KristjanESPERANTO: Hmm... Good eye! I'll test it.

KristjanESPERANTO: Without waiting for the `load` event, tests fail because modules aren't fully initialized.

While investigating I used this as an opportunity to refactor the `getDocument()` method to use `async/await` and `once(dom.window, "load")`.

KristjanESPERANTO: But now there are additional suspicious log entries. I need to look into this again - in the meantime, I've set the PR to draft.
```
stderr | tests/e2e/modules/clock_spec.js > Clock module > with showSun/MoonTime enabled
Could not load script: "http://localhost:8080/modules/default/clock/clock.js"

stderr | tests/e2e/modules/clock_spec.js > Clock module > with showSun/MoonTime enabled
Could not load link: "http://localhost:8080/css/custom.css"
```

khassel: well, I remember how long I needed getting the jest tests "stable" (the e2e tests were never really stable) ...

KristjanESPERANTO: Respect for all this work! I'm really learning a lot from it.

KristjanESPERANTO: While investigating the suspicious log entries which surfaced after the fetch migration I kept running into JSDOM’s limitations. That pushed me to migrate the E2E suite to Playwright instead - much closer to real browser behaviour and no more fighting JSDOM’s quirks. Of course, this has caused many changes...

@khassel, what do you think?

tests/e2e/modules/clock_spec.js
```diff
@@ -134,7 +134,7 @@ describe("Clock module", () => {
 			const weekToShow = `Week ${currentWeekNumber}`;
 			const elem = await helpers.waitForElement(".clock .week");
 			expect(elem).not.toBeNull();
-			expect(elem.textContent).toBe(weekToShow);
+			await expect(helpers.expectTextContent(elem, { equals: weekToShow })).resolves.toBe(true);
```
KristjanESPERANTO: Oh, we can make that easier. I'll take another look at it soon.

tests/e2e/helpers/global-setup.js
khassel: from my side we can go this way

tests/e2e/modules/clock_spec.js
KristjanESPERANTO: I tried introducing `expectTextContent` as a helper function, but it didn't make the tests easier to understand. I just removed the new helper function - this adds a few more lines per test, but makes the test logic transparent and future failures should be easier to diagnose.

khassel: let me know when ready for review

KristjanESPERANTO: Just some minutes. I take another quick look at it now.

KristjanESPERANTO: I think I've discovered some potential for optimization. Playwright has a few features that might make some of our complex helper functions obsolete. Let's see how deep I'll get into that rabbit hole... :smiley:

khassel: take your time ...

KristjanESPERANTO: Okay, I'm done now! 🤯 Sorry for so many changes in one PR, it's hard to review.

KristjanESPERANTO: Should we give Copilot a chance to do a review? :slightly_smiling_face:

.github/workflows/automated-tests.yaml
```diff
@@ -59,6 +59,9 @@ jobs:
       - name: "Install MagicMirror²"
         run: |
           node --run install-mm:dev
+      - name: "Install Playwright browsers"
+        run: |
+          npx playwright install --with-deps chromium
```
khassel, line 64: one question concerning this command: If I want to run e2e tests locally I have to execute this before? If so, could this be done inside `node --run install-mm:dev`?

KristjanESPERANTO: Good idea! Playwright already checks if browsers are installed and skips if up to date. 

So this should work:

```json
"install-mm:dev": "npm install --no-audit --no-fund --no-update-notifier && npx playwright install chromium",
```

What do you think?

khassel: the `--with-deps` is missing in the new statement. Do you know **where** the stuff is installed? In `node_modules` or somewhere in the os?

KristjanESPERANTO: > Do you know where the stuff is installed?

In a Playwright cache directory: `~/.cache/ms-playwright/chromium-<version>/`

Not in `node_modules` - that's why they persist across `npm install` and other repos.

> the `--with-deps` is missing in the new statement.

**With `--with-deps`:**
- ✅ Installs system dependencies (libnss3, etc.) - works out of the box
- ❌ Requires sudo

**Without `--with-deps`:**
- ❌ User might need to install system deps manually once
- ✅ No sudo needed

I'm not so sure what the best option is. But **without** may suffice for the script since deps are usually already installed on dev machines.

khassel: thanks, so I would go without.

KristjanESPERANTO: Okay. I'm preparing a commit for that :slightly_smiling_face:

khassel (approved):

KristjanESPERANTO:
That was huge! Thanks for reviewing and accepting, @khassel! :smiley:

PR #3967, opened 2025-11-19
[compliments] fix: duplicate query param "?" in compliments module refresh url

xsorifc28:
Discovered a bug for compliments module  "remote file config" fetching. If the remote file contains a `?`, like `http://remoteconfig.com?type=all`, the url becomes `http://remoteconfig.com?type=all?dummy={timestamp}`, which is incorrect.

This fix:
- Checks if `this.config.remoteFile.includes` already contains a `?`
  - If it does, uses `&` to append the dummy parameter
  - If it doesn't, uses `?` to start a new query string


KristjanESPERANTO:
Nice catch and good PR! Thanks for this fix! :smiley:

PR #3969, opened 2025-11-19
[compliments] refactor: optimize `loadComplimentFile` method and add unit tests

KristjanESPERANTO:
## Changes

- Replace `indexOf()` with `startsWith()` for cleaner protocol detection
- Use `URL` API for robust cache-busting parameter handling
- Add HTTP response validation and improved error logging
- Add JSDoc type annotations for better documentation
- Remove unused `urlSuffix` instance variable
- Add unit tests
- Fix `.gitignore` pattern

## Motivation

After merging #3967, I noticed some potential for improving reliability and user experience related to the method `loadComplimentFile`. With these changes the method now validates URLs upfront to catch configuration errors early, checks HTTP status codes to detect server issues (404/500), and provides specific error messages that help users troubleshoot problems.

The complexity of the code does not really increase with the changes. On the contrary, the method should now be more intuitive to understand.

## Testing

Added unit tests for `loadComplimentFile()` to validate the improvements:
- HTTP error handling
- Cache-busting

Since E2E tests already cover the happy path, these unit tests focus on error cases and edge cases.

## Additional Fix

While adding the test file, I discovered that the `.gitignore` pattern `modules` was incorrectly matching `tests/unit/modules/`, preventing test files from being tracked. Changed to `/modules` to only match the root directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ planned for 2026-01-01
 - [tests] migrate e2e tests to Playwright (#3950)
 - [calendar] refactor: migrate CalendarFetcher to ES6 class and improve error handling (#3958)
 - [gitignore] cleanup/simplify .gitignore (#3952, #3954, #3968, #3969)
-- refactor(compliments): optimize `loadComplimentFile` method (#3969)
+- refactor(compliments): optimize `loadComplimentFile` method and add unit tests(#3969)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


rejas (approved):

KristjanESPERANTO:
@rejas Thanks for reviewing and accepting! :slightly_smiling_face:

PR #3970, opened 2025-11-20
fix: set compliments remote file minimum delay to 15 minutes

xsorifc28:
fix: set compliments remote file minimum delay to 15 minutes..extra *60 in there was making it 15 hours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ planned for 2026-01-01
 - [newsfeed] fixed header layout issue introduced with prettier njk linting (#3946)
 - [weather] fixed windy icon not showing up in pirateweather (#3957)
 - [compliments] fixed duplicate query param "?" when constructing refresh url (#3967)
-- [compliments] fixed compliments remote file minimum delay to be 15 minutes (#3969)
+- [compliments] fixed compliments remote file minimum delay to be 15 minutes (#3970)
 
 ### Updated
 


sdetweil:
Was set long intentionally. But configurable if need to be changed. Most won’t change every 15 minutes. Maybe daily. The default should not be fast

rejas:
I agree with @sdetweil 

Maybe we just set the default value to 1 day and also adjust the comment there accordingly? 15 hours seems kinda strange indeed

sdetweil:
Yeh, I made it hours instead of minutes, but didn’t change 15 to 12

xsorifc28:
The comment along the code said 15 minutes. But default value can be one day, agreed, as long as the comment matches and the value is configurable.

xsorifc28:
To clarify, the `remoteFileRefreshInterval` is configurable and by default 0 (off).
The `refreshMinimumDelay` is 15 hours, so `remoteFileRefreshInterval` at minimum can be 15 hours.

I don't see why `refreshMinimumDelay` needs to be so high since the default value is 0. It should be up to the user, why not be able to refresh every minute? Again, out-of-the-box the default is 0.

My use case is: the remote file is updated by multiple sources throughout the day, and I'd like the mirror to pull these changes multiple times a day.

sdetweil:
Ok, I see your point
Default is no refresh(0)
If the value is NOT 0, then the refresh code is executed on that time cycle 

Minimum 15 minutes seems ok. Which is what I intended but pasted an extra * 60

xsorifc28:
> Minimum 15 minutes seems ok. Which is what I intended but pasted an extra * 60

Yes, I think the logic was correct, just a typo 😄

KristjanESPERANTO:
@sdetweil So we can merge this, right? 🙂

sdetweil:
Yes

PR #3974, opened 2025-11-29
[core] chore: simplify Wayland start script

KristjanESPERANTO:
Remove `--enable-features=UseOzonePlatform` and `--ozone-platform=wayland` flags as Electron 38+ changed the default `--ozone-platform` to `auto`, which automatically detects and uses Wayland when running in a Wayland session.

Source: https://www.electronjs.org/blog/electron-38-0#removed-electron_ozone_platform_hint-environment-variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,8 @@ planned for 2026-01-01
 - [tests] migrate e2e tests to Playwright (#3950)
 - [calendar] refactor: migrate CalendarFetcher to ES6 class and improve error handling (#3958)
 - [gitignore] cleanup/simplify .gitignore (#3952, #3954, #3968, #3969)
-- refactor(compliments): optimize `loadComplimentFile` method and add unit tests(#3969)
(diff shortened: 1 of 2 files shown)


khassel:
@KristjanESPERANTO this is not working on my side (inside docker container)

I can remove `--enable-features=UseOzonePlatform` but not `--ozone-platform=wayland`. The linked documentation is not clear, they mention that the default of `--ozone-plaftform` is now `auto` but when I try this option I get 

```bash
node@magicmirroros:/opt/magic_mirror$ ./node_modules/.bin/electron js/electron.js --ozone-platform=auto
[1591:1214/132827.200843:FATAL:ui/ozone/platform_selection.cc:46] Invalid ozone platform: auto

node@magicmirroros:/opt/magic_mirror$ npx electron --version
v39.2.7
```

So AFAIS we should bring back the `--ozone-platform=wayland` flag before next release.

Other ideas?

KristjanESPERANTO:
Okay, interesting. I don't have any other ideas at the moment, so I've already created a PR :slightly_smiling_face:

PR #3975, opened 2025-11-29
Draft: Added extra regions to index.html

jkriegshauser:
I added these regions a while ago locally so that my config could make use of them. And I'm tired of merging them in with each update, so I figured I'd try a pull request 😄

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -234,6 +234,17 @@ sup {
   text-align: left;
 }
 
+.region.bottom.left1 {
+  left: 0;
+  bottom: 100%;
+}
+
+.region.bottom.right1 {
+  left: 25%;
+  bottom: 100%;
+  width: 75%;
+}
+
 .region table {
   width: 100%;
   border-spacing: 0;


sdetweil:
Where is the css for those?

jkriegshauser:
> Where is the css for those?

Oops, left that out. I just pushed what I used.

Though I'm also open to any other suggestions about adding regions that don't require modifying the index.html

sdetweil:
The detection code reads index.html to get the names 
There is no way without moding index.html

sdetweil:
The css could be in custom.css

jkriegshauser:
> The detection code reads index.html to get the names There is no way without moding index.html

That's unfortunate. Where is the detection code?

> The css could be in custom.css

Yeah that's where I had it prior to creating the PR.

sdetweil:
js/utils.js getModulePositions

we validate all the positions used against the discovered list, cause before we might fail, or silently not show the module.

we need the html to connect the styles to for proper positioning 

others want left/right in middle_center

but you can also move a module right w css without a position

jkriegshauser:
> but you can also move a module right w css without a position

This is a good point that i hadn't really considered. Let me see if there are any regions that I'm not using that i could just move into place rather than adding new ones

sdetweil:
you could easily have applied your right1 css to the module in bottom_left
wouldn't need to specify bottom as its already specified

you don't need to move a position.. the whole page is open to css.. so you can put your module anywhere you like.
the regions are just to generalize the layout...
you can use one of the positions as a starting point if it helps.. (like bottom left), and set the module left:25%; to move it

jkriegshauser:
Thanks, that actually did the trick. Closing this PR

Issue #3971, opened 2025-11-20
[Bug] CALENDAR_EVENTS are being resend all the time with no updates


michaelarnauts:
### Environment

MagicMirror² version: v2.33.0
Node version: v22.21.1
npm version:   10.9.4
Platform: Docker

### Which start option are you using?

node --run server

### Are you using PM2?

I don't know

### Module

calendar

### Have you tried disabling other modules?

- [ ] Yes
- [x] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

<details>
<summary>Configuration</summary>

```
var config = {
    address: '0.0.0.0',
    port: 8080,
    ipWhitelist: [],
    language: 'nl',
    timeFormat: 24,
    units: 'metric',
    logLevel: ["LOG", "WARN", "ERROR"],
    modules: [
        {
            module: 'alert',
        },
        {
            module: 'clock',
            position: 'top_right',
            config: {
                displaySeconds: false,
                dateFormat: 'dddd LL'
            }
        },
        {
            module: 'calendar',
            position: 'bottom_left',
            config: {
                fetchInterval: 600000,
                maximumNumberOfDays: 2,
                maxTitleLength: 50,
                fade: false,
                getRelative: 0,
                urgency: 0,
                timeFormat: 'dateheaders',
                dateFormat: 'dddd D MMMM',
                fullDayEventDateFormat: 'ddd D MMMM',
                broadcastEvents: false,
                sliceMultiDayEvents: true,
                colored: true,
                defaultSymbolClassName: '',
                calendars: [
                    {
                        name: 'xxx',
                        url: 'https://calendar.google.com/calendar/ical/xxx/private-yyy/basic.ics',
                    },
                    {
                        name: 'xxx',
                        url: 'https://calendar.google.com/calendar/ical/xxx/private-yyy/basic.ics',
                    },
                    {
                        name: 'xxx',
                        url: 'https://calendar.google.com/calendar/ical/xxx/private-yyy/basic.ics',
                    },
                    {
                        name: 'xxx',
                        url: 'https://calendar.google.com/calendar/ical/xxx/private-yyy/basic.ics',
                    },
                    {
                        name: 'xxx',
                        url: 'https://calendar.google.com/calendar/ical/xxx/private-yyy/basic.ics',
                    },
                    {
                        name: 'xxx',
[...]

PR #3976, opened 2025-11-29
[calendar] fix: prevent excessive fetching on client reload and refactor `calendarfetcherutils.js`

KristjanESPERANTO:
The bottom line of this PR is, that it fixes an issue and simplifies the code by dealing with the TODOs in the code.

For review, I suggest looking at each commit individually. If there are too many changes for a PR, let me know and I'll split it up :slightly_smiling_face:

## 1. [fix(calendar): prevent excessive fetching with smart refresh strategy](https://github.com/MagicMirrorOrg/MagicMirror/pull/3976/commits/8892cd3d5aa3aca45ccc97e1eaff4685c98b3fcd)

- Add lastFetch timestamp tracking to CalendarFetcher
- Add shouldRefetch() method with configurable minimum interval (default: 3 minutes)
- When reusing existing fetcher: fetch if data is stale (>3 min), otherwise broadcast cached events
- Prevents double broadcasts to consuming modules while maintaining fresh data
- Balances rate limit prevention (Issue https://github.com/MagicMirrorOrg/MagicMirror/issues/3971) with data freshness on user reload
- Prevents excessive fetching during rapid reloads (e.g., Fully Kiosk screensaver use case)
- Allows fresh calendar data when enough time has passed since last fetch

## 2. [refactor(calendar): simplify event exclusion logic](https://github.com/MagicMirrorOrg/MagicMirror/commit/d507aba82d68bb0feb0c3b45a321f9407f95d311)

- Extract filtering logic from `shouldEventBeExcluded` into new helper `checkEventAgainstFilter`
- Simplify the main loop in `shouldEventBeExcluded

It resolves a TODO from the comments in the code:
* `This seems like an overly complicated way to exclude events based on the title.`

## 3. [refactor(calendar): extract recurring event expansion logic](https://github.com/MagicMirrorOrg/MagicMirror/commit/d510160bd214d1440cb8c2f7598cf48ec0c19c36)

This change separates the expansion of recurring events from the main filtering loop into a new helper function 'expandRecurringEvent'.

It resolves two TODOs from the comments in the code:
- `This should be a separate function`
- `This should create an event per moment so we can change anything we want`

This improves code readability, reduces complexity in 'filterEvents', and allows for cleaner handling of individual recurrence instances.

## 4. [refactor(calendar): simplify recurring event handling](https://github.com/MagicMirrorOrg/MagicMirror/commit/b04f716cc0f74883bbcf8d812d76ee56ebca8fe5)

- Simplify 'getMomentsFromRecurringEvent' using modern syntax
- Improve handling of full-day events across different timezones

## 5. [test(calendar): fix UNTIL date in fullday_until.ics fixture](https://github.com/MagicMirrorOrg/MagicMirror/pull/3976/commits/1d762b2ade2b24e157e66b83774cdef4ccc95c52)

The issue was with the UNTIL date being May 4th while DTSTART was May 5th. This created an invalid recurrence rule where the end date came before the start date.

The fix only adjusts the UNTIL date from May 4th to May 5th, so it matches the start date.

### DIFF
--- a/modules/default/calendar/calendarfetcher.js
+++ b/modules/default/calendar/calendarfetcher.js
@@ -182,15 +182,16 @@ class CalendarFetcher {
 	}
 
 	/**
-	 * Check if enough time has passed since the last fetch to warrant a new one
-	 * @param {number} minInterval - Minimum milliseconds between fetches (default: 3 minutes)
+	 * Check if enough time has passed since the last fetch to warrant a new one.
+	 * Uses reloadInterval as the threshold to respect user's configured fetchInterval.
 	 * @returns {boolean} True if a new fetch should be performed
 	 */
-	shouldRefetch (minInterval = 3 * 60 * 1000) {
+	shouldRefetch () {
 		if (!this.lastFetch) {
 			return true;
 		}
-		return (Date.now() - this.lastFetch) >= minInterval;
+		const timeSinceLastFetch = Date.now() - this.lastFetch;
+		return timeSinceLastFetch >= this.reloadInterval;
 	}
 
 	/**


sdetweil:
But 1 is how we get calendar updates. Resending the cached events doesn’t keep the calendar up to date

KristjanESPERANTO:
> But 1 is how we get calendar updates. Resending the cached events doesn’t keep the calendar up to date

Very good catch! Many thanks! In the meantime I've updated the implementation in commit 182d3dd to address exactly that concern.

Now when the calendar is reloaded, it always broadcasts the cached events immediately (for fast UI response), and then checks if the data is older than 3 minutes - if so, it fetches fresh data in the background. This way you get instant display with cached data, plus automatic updates when needed, while still preventing rate limit issues from rapid reloads. I couldn't think of a simpler approach.

----

I'm thinking about that I should split up this PR since it touches several points. Then we could look at each individual point more easily and decide whether we merge them or not. What do you think?

sdetweil:
Do you know what the impact of the multiple broadcasts does to the using modules?  Ext3, calendarMonthly

ext3 has a 10 min (default) refresh cycle

KristjanESPERANTO:
> Do you know what the impact of the multiple broadcasts does to the using modules? Ext3, calendarMonthly

Again a good point! I've just made a small adjustment to optimize this. As it stands now, it shouldn't have any noticeable effect on ext3 and similar modules.

## Calendar Fetch Strategy Comparison

### Key Differences

| Aspect | Before | After (< 3 min) | After (> 3 min) |
|--------|----------|---------------|---------------|
| **Fetch on reload** | ✅ Always | ❌ No | ✅ Yes |
| **Broadcasts per reload** | 1 | 1 | 1 |
| **Rate limit risk** | ❌ High | ✅ None | ✅ Very low |
| **Data freshness** | ✅ Always fresh | ⚠️ Cached | ✅ Fresh |
| **Ext3 processing** | 1x | 1x | 1x |
| **Network requests** | Every reload | None | When needed |

### Impact on Ext3 and Other Modules

**Before:**
- Every page reload → 1 fetch → 1 broadcast → 1x processing (but too many fetches)

**After:**
- Every page reload → 0 or 1 fetch → 1 broadcast → 1x processing
- Ext3's 10-minute refresh cycle remains independent and unaffected

KristjanESPERANTO:
Since calendars don't change that often, we could even set the default cache limit time higher. Maybe 5 or 10 minutes. And perhaps even offer it as an option that users can adjust.

modules/default/calendar/calendarfetcherutils.js
sdetweil, line 1: On this
if (isFullDayEvent && rule.options?.until) {
			rule.options.until = moment(rule.options.until).endOf("day").toDate();
		}
I used start of day  vs end of, cause end of could appear as tomorrow in some timezones. Start of  day will never be wrong.

modules/default/calendar/calendarfetcher.js
sdetweil, line 1: What’s the usecase for caching?
Currently on strict setTimeout , only one fetcher per url

If ANOTHER browser (.0001% of the time) 
Finds existing fetcher , must be same parms if connecting to this node helper 

If we saved last broadcast, send it again, sendSocketNotification goes to all clients( sockets, same room) . I think this happens now ( refresh ui, events show up immediately) 

Then
Timeout expires, we fetch again, save broadcast, send broadcast to all clients. We are synced, only fetch on cycle started by first request from same config 
No ‘extra’ fetches

modules/default/calendar/calendarfetcherutils.js
KristjanESPERANTO: Good point about timezones. But I think `endOf("day")` shouldn't cause issues in this case because it operates on the same calendar day - it just moves the time to 23:59:59 of that same day, never into the next day.

So even if someone is in a timezone like UTC+12, `endOf("day")` on May 5th stays May 5th 23:59:59 - it can't "appear as tomorrow" because we're not adding time, just extending to the end of the current day.

That said, this is admittedly a workaround for a deeper issue in rrule.js itself. The library has known problems with UNTIL and timezone handling (see [rrule#596](https://github.com/jkbrzt/rrule/issues/596), open since Aug 2023), and unfortunately rrule.js appears to be unmaintained at this point (~180 open issues).

I'm working on replacing rrule.js with [rrule-temporal](https://github.com/ggaabe/rrule-temporal) in node-ical to solve these problems at the root. rrule-temporal is built on the Temporal API which handles timezones correctly by design. Since this is a major change, it will take a while to implement it, so unfortunately I think we will need to use workarounds like this for the time being.

modules/default/calendar/calendarfetcher.js
KristjanESPERANTO: You are right that `fetchInterval` handles the normal update loop perfectly fine.

However, the problem happens specifically during **page reloads** (e.g. when using Fully Kiosk Browser as a screensaver).

1.  **The Trigger:** When the page reloads, the client sends `ADD_CALENDAR`. The server treats this as a new initialization and triggers an **immediate fetch** to show data right away. This bypasses the `fetchInterval` completely.

2.  **The Burst:** In the issue report (#3971), the user has **8 calendars**. If their screensaver toggles just 3 times in a few minutes (e.g. someone walking by), that means:
    `3 reloads * 8 calendars = 24 requests` to Google in a very short time.

3.  **The Evidence:** The user's logs in the issue confirm the **429 Too Many Requests** error from Google:
    ```
    [2025-11-29 11:07:35.025] [ERROR] Calendar Error. Could not fetch calendar: ... Error: Too Many Requests
    [2025-11-29 11:07:35.283] [ERROR] Calendar Error. Could not fetch calendar: ... Error: Too Many Requests
    [2025-11-29 11:07:36.290] [ERROR] Calendar Error. Could not fetch calendar: ... Error: Too Many Requests
    ...
    ```

**The solution of this PR:**
With this PR, the server checks `lastFetch` on reload. If we just fetched data recently (< 3 minutes), we serve the **cached data** immediately instead of hitting Google again. This protects the API limits while still ensuring the user sees data instantly.

modules/default/calendar/calendarfetcherutils.js
sdetweil: Remember full day doesn’t have time, I ran into that here and in the front end

modules/default/calendar/calendarfetcher.js
sdetweil: But you could have fixed this with addCalendar 
Already exists, poof, do the already exists path. None of this caching logic, timeouts, etc 
It’s the exact same calendar def coming in cause it’s to the same server 

and this extra browser is such a small percentage, no need for more code to handle it

modules/default/calendar/calendarfetcherutils.js
KristjanESPERANTO: Yes, it's still necessary. Here's a test script that demonstrates the bug:

```javascript
const { RRule } = require('rrule');
const moment = require('moment-timezone');

// Full-day event: daily at 00:00 UTC, starting March 10
const dtstartUTC = new Date(Date.UTC(2024, 2, 10, 0, 0, 0));

// UNTIL=20240315 parsed in Europe/Berlin (UTC+1)
// "March 15 00:00 Berlin" = "March 14 23:00 UTC"
const untilRaw = moment.tz('2024-03-15', 'Europe/Berlin').toDate();
const untilFixed = moment.tz('2024-03-15', 'Europe/Berlin').endOf('day').toDate();

const rule1 = new RRule({ freq: RRule.DAILY, dtstart: dtstartUTC, until: untilRaw });
const rule2 = new RRule({ freq: RRule.DAILY, dtstart: dtstartUTC, until: untilFixed });

console.log('WITHOUT endOf("day"):', rule1.all().length, 'days, last:', rule1.all().pop()?.toISOString().slice(0,10));
console.log('WITH endOf("day"):   ', rule2.all().length, 'days, last:', rule2.all().pop()?.toISOString().slice(0,10));
```

**Output:**
```
WITHOUT endOf("day"): 5 days, last: 2024-03-14
WITH endOf("day"):    6 days, last: 2024-03-15
```

The issue: When rrule.js checks if March 15 should be included, it compares:
- Event time: `00:00:00 UTC` on March 15
- UNTIL: `23:00:00 UTC` on March 14 (because of timezone conversion)

Since `23:00 < 00:00`, March 15th is excluded. The `endOf("day")` fix ensures UNTIL is always *after* the event time on that day.

modules/default/calendar/calendarfetcher.js
KristjanESPERANTO: For purely solving the rate limit issue, we could simply do:

```javascript
} else {
    fetcher.broadcastEvents();  // Just broadcast, never fetch
}
```

But this creates a **UX problem**: If the mirror was in screensaver mode for 8 hours and wakes up, the user would see 8-hour-old calendar data. They would have to wait for the next `fetchInterval` cycle to see current events.

The `shouldRefetch()` logic is the **compromise**:
- **Reload after 30 seconds:** Use cache → Protects API
- **Reload after 8 hours of sleep:** Fetch → Fresh data for user

So the caching with `lastFetch` gives us the best of both worlds: rate limit protection during rapid reloads, but still fresh data when the mirror wakes up after a long idle period.

KristjanESPERANTO: By the way: Thank you for your critical questions! I really appreciate them :slightly_smiling_face:

KristjanESPERANTO: > But you could have fixed this with addCalendar

I think you mean handling this on the client side in `calendar.js` - checking if we already added the calendar before sending `ADD_CALENDAR`?

The problem is: **the client doesn't know if the server still has the fetcher.**

On a page reload:
1. The browser context is completely reset
2. The module (`calendar.js`) starts fresh - it has no memory of what happened before
3. It **must** send `ADD_CALENDAR` to get data
4. Only the **server** knows: "I still have this fetcher from 30 seconds ago"

So the logic has to be server-side because only the Node helper persists across browser reloads. The client can't avoid sending `ADD_CALENDAR` - it genuinely needs calendar data.

sdetweil: BUT, the scenario is really a bug on the UI side.. resume should force a new fetch(refetch), which should, because it KNOWS it was asleep.. cause a refetch...  the fetcher kept going, and sent useless junk (cause there is no shutdown to the fetcher on suspend of the UI
this separation of tasks work has always been a problem in design..
maybe your cache is another workaround to fixing it correctly. 
we'll see what happens..

sdetweil: We already covered the browser sending a new add calendar. It doesn’t, it’s first.
It does, is the fetch active, wait
Fetch not active, send what’s there, wake-up waiting(promise) 

if the browser is resuming  it could send a refresh instead of an add calendar 
Which basically goes thru the same logic 
First refresh fetch etc

but refresh could happen a lot w pages/carousel
So maybe lastfetch  fits there, less than fetchInterval don’t send or fetch
It was all asynchronous anyhow

KristjanESPERANTO: > So maybe lastfetch fits there, less than fetchInterval don’t send or fetch

Oh, that's a great idea! I will check this out.

KristjanESPERANTO: With the latest commit I simplified `shouldRefetch()` to use `fetchInterval` directly instead of a hardcoded 3-minute threshold. This way users control the rate limit protection through their existing config - no magic numbers needed. Great idea, Sam! 👍

sdetweil:
Seems ok.

modules/default/calendar/calendarfetcher.js
sdetweil: Looks ok

KristjanESPERANTO:
> Seems ok.

Great. It would be awesome if you'd merge it :slightly_smiling_face:

PR #3986, opened 2025-12-10
update dependencies and min. node version (corresponding to electron version v39)

khassel:


khassel:
Did I ever mention that I hate github actions?

The above failed job "Enforce changelog" was removed on `develop` so why does it still starts? Is there some caching?

Should be merged anyway.

khassel:
Maybe the problem is that the workflows are run from `master`, in this case we will get this error in every PR until next release is done.

Or we add the new `enforce-pullrequest-rules.yaml` with a cherry pick to `master`?

KristjanESPERANTO:
> Or we add the new enforce-pullrequest-rules.yaml with a cherry pick to master?

That would probably be the cleanest solution. But waiting for the next release would also be okay, I think :slightly_smiling_face:

sdetweil:
As we are only a couple weeks away

PR #3987, opened 2025-12-10
testing: update "Enforce Pull-Request Rules" workflow

khassel:
With this update the workflow file from inside the feature branch is used, not the old stuff coming from `master` as before. This does not help for the currently failing job which still comes from `master` (we have to live with this until next release), but this will help in the future to prevent such errors.

Tested this on my fork:
- base against `develop`: workflow is skipped
- base against `master`: workflow fails
- base against `master` with label `mastermerge`: workflow is skipped

I took this new workflow from the same repo where the previous workflows was taken (see diff for the link) so this is the further development.


KristjanESPERANTO (approved):

KristjanESPERANTO:
Looks good :slightly_smiling_face:  :+1:

PR #3988, opened 2025-12-12
add checksum to test whether calendar event list changed

### LINKED ISSUE #3971: [Bug] CALENDAR_EVENTS are being resend all the time with no updates

sdetweil:
for issue #3971 add checksum to test if event list changed to reduce/eliminate no change screen update
fixes #3971

crc32 checksum created in node helper, easy require, vs trying to do in browser. 
added to socket notification payload, used in browser

### DIFF
--- a/modules/default/calendar/calendar.js
+++ b/modules/default/calendar/calendar.js
@@ -184,9 +184,9 @@ Module.register("calendar", {
 
 		if (notification === "CALENDAR_EVENTS") {
 			if (this.hasCalendarURL(payload.url)) {
-				// have we received data for this url
+				// have we received events for this url
 				if (!this.calendarData[payload.url]) {
-					// no,m setup tehe structure
+					// no, setup the structure to hold the info
 					this.calendarData[payload.url] = { events: null, checksum: null };
 				}
 				// save the event list


sdetweil:
i sent a PR without CHANGELOG.. figuring it might succeed... guess not

khassel:
> i sent a PR without CHANGELOG.. figuring it might succeed... guess not

without updating `master` we have to live with these test fails until next release ...

modules/default/calendar/calendar.js
```diff
@@ -184,13 +184,27 @@ Module.register("calendar", {
 
 		if (notification === "CALENDAR_EVENTS") {
 			if (this.hasCalendarURL(payload.url)) {
-				this.calendarData[payload.url] = payload.events;
+				// have we received data for this url
+				if (!this.calendarData[payload.url]) {
+					// no,m setup tehe structure
```
khassel: can you please fix this comment (don't understand)

KristjanESPERANTO:
Nice!

I have a question about the order of operations: `broadcastEvents()` is called *before* the checksum comparison happens. Wouldn't that mean the broadcast still gets executed on every fetch, even when the data hasn't changed?

```javascript
if (this.config.broadcastEvents) {
    this.broadcastEvents();  // ← called here
}
// ...
if (this.calendarData[payload.url].checksum === payload.checksum) {
    return;  // ← but checksum check is here
}
```

Shouldn't the checksum check be moved earlier, before any processing happens, to get the full benefit of this optimization?

sdetweil:
Not our decision on how the other modules use the data. 
this would be a material change for them. We just want to stop OUR ui from flashing

KristjanESPERANTO:
I see your point about not wanting to change behavior for other modules. But wouldn't they actually benefit from not receiving duplicate data? 

If a module truly needs a "heartbeat" style notification regardless of changes, that seems like a design issue on their end rather than something the calendar should accommodate.

Just a thought - the current approach is fine too, I just wonder if we're being overly cautious here.

sdetweil:
@rejas  can you look at this?

modules/default/calendar/calendar.js
```diff
@@ -602,7 +616,7 @@ Module.register("calendar", {
 		let events = [];
 
 		for (const calendarUrl in this.calendarData) {
-			const calendar = this.calendarData[calendarUrl];
+			const calendar = this.calendarData[calendarUrl].events;
```
rejas, line 619: why this addition?

sdetweil: because we need to save the events and the CRC.. before the calendarData was just the events.. 
now we need to save the CRC that came with them to be able to compare next time. 

so, instead of just setting the calendarData, we need to save JUST the events, until we test the Crc
```
all new
			if (this.hasCalendarURL(payload.url)) {
				// have we received events for this url
				if (!this.calendarData[payload.url]) {
					// no, setup the structure to hold the info
					this.calendarData[payload.url] = { events: null, checksum: null };
				}
				// save the event list
this line replaced after above		this.calendarData[payload.url].events = payload.events;
```

sdetweil:
@rejas  did you see my answer?  hope to get this in next release..

rejas:
> I see your point about not wanting to change behavior for other modules. But wouldn't they actually benefit from not receiving duplicate data?
> 
> If a module truly needs a "heartbeat" style notification regardless of changes, that seems like a design issue on their end rather than something the calendar should accommodate.
> 
> Just a thought - the current approach is fine too, I just wonder if we're being overly cautious here.

Yeah, optimizing for the modules would be nice of course, but we can safely assume that we also would break some old module that lots of people are using :-)

So yeah, I am fine with this approach too. Lets merge it

rejas (approved):

PR #3993, opened 2025-12-21
[demo] feat: add demo

KristjanESPERANTO:
This PR started as an **experiment** to see if it's possible to run MagicMirror² for showcasing directly in the browser without a server. The demo uses the real MagicMirror core files and default modules, only mocking server components (API calls, Socket.IO, node helpers).

**Live preview:** https://kristjanesperanto.github.io/MagicMirror/

**After merge and next release**, the demo would be available at: https://magicmirrororg.github.io/MagicMirror/

Even if the code doesn't add too many lines to the repository, it increases our maintenance effort. So the question is whether we want that. I'm not so sure myself. For me, my experiment was successful and actually complete. I think a live demo is kind of cool, and if the effort is too great for us, we could easily remove the demo at some point, since it's not interwoven with the rest.

But it's no problem if we drop it :slightly_smiling_face:

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,0 +1,69 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches: ["master"]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Create redirect index.html
+        run: |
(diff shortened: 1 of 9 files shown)


khassel:
I'm not sure but are these the right fonts on the demo page?

khassel:
Another demo approach could be to run a normal server instance on our server and publish it e.g. under `https://demo.magicmirror.builders/`

sdetweil:
I think this is the better way. Keep it up to date as we release

khassel:
because we have already everything I [added a few lines in the server setup](https://github.com/MagicMirrorOrg/MagicMirror-Server/compare/32fee8556a5f6593758d27110cc8867fa79b6569...8b157dc4b76fc39bad834804555654c488e02761), result see https://demo.magicmirror.builders/

KristjanESPERANTO:
> I'm not sure but are these the right fonts on the demo page?

Oh, I have to fix this, but let's see if we decide for my approach or not :slightly_smiling_face:

KristjanESPERANTO:
Replaced by another approach (#3995) :slightly_smiling_face:

PR #4001, opened 2026-01-01
update Collaboration.md and dependencies

khassel:


sdetweil:
What is the xxxxx NPM_TOKEN=npm_xxxxx
I generate a token but it doesn’t give me the value, only allows me to set the name

khassel:
it is the new token, e.g.:

grafik

rejas:
I dont like bypassing 2FA, more when I am back at my desk during the weekend 

Am 1. Januar 2026 23:03:10 MEZ schrieb Karsten Hassel ***@***.***>:

khassel:
> I dont like bypassing 2FA

I described only one way to get `npm publish` running. I'm not involved in this topic and the only use case where I have to login and publish something is ~ once a year here.

Please consider that I have no node/npm stuff on my working system (windows) so I have to do this per command line in a linux container which has no browser. Let me know how to do this with 2FA. Or better change the content of Collaboration.md in this PR.

sdetweil:
I also have no npm content other than MM,

rejas:
> Please consider that I have no node/npm stuff on my working system (windows) so I have to do this per command line in a linux container which has no browser. Let me know how to do this with 2FA. Or better change the content of Collaboration.md in this PR.

Without a browser it indeed could be tricky. When I publish on npm (via console), it does open a brwoser for me and I can enter my 2fa code there.

Maybe add your setup (with the NPM_TOKEN instrcutions) as a secondary way for publishing (one with brwoser, one without brwoser)?

khassel:
as I already mentioned: I know only the method I wrote down in Collaboration.md.

If you prefer a method with 2fa write it down and commit it into this PR. Leaving my method as secondary would be helpful if the 2fa-method is not usable in headless setups.

KristjanESPERANTO:
I've just added a commit (b73aa213e20774c752705f787cfca6f83ca430de) with both methods for npm publishing.

This should address both @rejas' security concerns and @khassel's headless setup requirements. The 2FA method is clearly marked as recommended, while the token method remains available as a documented fallback for environments without browser access.

What do you think? :slightly_smiling_face:

khassel:
Thanks @KristjanESPERANTO , can't approve (because I created this PR) but I'm fine with this.

sdetweil:
npm login worked for me just now

KristjanESPERANTO:
> npm login worked for me just now

Because you probably chose the "bypass 2FA option" and saved the token.

---

The problem with the bypass is that if one of those systems with bypass are getting compromised, malicious code could be easily published via npm. There have already been npm dependency chain attacks, so it's not entirely unrealistic.

khassel:
> The problem with the bypass is that if one of those systems with bypass are getting compromised, malicious code could be easily published via npm. There have already been npm dependency chain attacks, so it's not entirely unrealistic.

I know about the npm problems.

When I do the `npm publish` using a container this container only lives for the ~ 5 min. doing this. I revoke the used token after the container is destroyed.

sdetweil:
It asked me to login with my passkey

PR #4002, opened 2026-01-02
run systeminformation in subprocess so the info is always displayed

khassel:
If an error occurs during startup, we request system information from the user. The problem is that this information is displayed too late, for example, if the configuration check fails.

My initial idea was to use `await Utils.logSystemInformation(global.version);`, but this increased the startup time.

Therefore, the function is now called in a subprocess. This approach provides the information in all cases and does not increase the startup time.

Before this change:

```bash
node@9914e2eea3c7:~/magicmirror$ node --run server
[2026-01-02 22:48:38.508] [LOG]   [app] Starting MagicMirror: v2.35.0-develop
[2026-01-02 22:48:38.514] [LOG]   [app] Loading config ...
[2026-01-02 22:48:38.522] [LOG]   [app] config template file not exists, no envsubst
[2026-01-02 22:48:38.661] [INFO]  [check_config] Checking config file /home/node/magicmirror/config/config.js ...
[2026-01-02 22:48:38.693] [ERROR] [check_config] Your configuration file contains syntax errors :(
Line 43 column 5: Unexpected comma in middle of array.
```

After this change:

```bash
node@9914e2eea3c7:~/magicmirror$ node --run server
[2026-01-02 22:47:30.209] [LOG]   [app] Starting MagicMirror: v2.35.0-develop
[2026-01-02 22:47:30.213] [LOG]   [app] Loading config ...
[2026-01-02 22:47:30.222] [LOG]   [app] config template file not exists, no envsubst
[2026-01-02 22:47:30.358] [INFO]  [check_config] Checking config file /home/node/magicmirror/config/config.js ...
[2026-01-02 22:47:30.391] [ERROR] [check_config] Your configuration file contains syntax errors :(
Line 43 column 5: Unexpected comma in middle of array.
node@9914e2eea3c7:~/magicmirror$ [2026-01-02 22:47:30.566] [INFO]  [utils]
- SYSTEM:   manufacturer: Microsoft; model: Docker Container; virtual: true; MM: v2.35.0-develop
- OS:       platform: linux; distro: Debian GNU/Linux; release: 13; arch: x64; kernel: 6.6.87.2-microsoft-standard-WSL2
- VERSIONS: electron: undefined; used node: 25.2.1; installed node: 25.2.1; npm: 11.7.0; pm2:
- ENV:      XDG_SESSION_TYPE: undefined; MM_CONFIG_FILE: undefined
            WAYLAND_DISPLAY:  wayland-0; DISPLAY: unix:0.0; ELECTRON_ENABLE_GPU: undefined
- RAM:      total: 14872.19 MB; free: 13091.86 MB; used: 1780.34 MB
- OTHERS:   uptime: 839 minutes; timeZone: UTC
```

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -1,7 +1,5 @@
-const os = require("node:os");
 const fs = require("node:fs");
-const si = require("systeminformation");
-const Log = require("./logger"); // needed with relative path because logSystemInformation is called in an own process in app.js
+const Log = require("logger");
 
 const modulePositions = []; // will get list from index.html
 const regionRegEx = /"region ([^"]*)/i;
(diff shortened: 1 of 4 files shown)


js/utils.js
KristjanESPERANTO: Would it make sense to move the function `logSystemInformation` from `utils.js` to the new `systeminformation.js`?

khassel: I had this approach too but decided against, now `js/systeminformation.js` exists already in this PR so we would need another naming for one of both.

khassel: Okay, now I understand, I'll take another look.

KristjanESPERANTO: Could there be a conflict between the package name `systeminformation` and the file name `systeminformation.js`? Maybe `sysinfo.js` would be an alternative name? **Just thinking loud**

Issue #4003, opened 2026-01-03
[Bug] repeating full day events show on the wrong day


sdetweil:
####  System Information  ####
- SYSTEM:   manufacturer: System manufacturer; model: System Product Name; virtual: false; MM: 2.34.0
- OS:       platform: linux; distro: Ubuntu; release: 22.04.5 LTS; arch: x64; kernel: 5.15.0-164-generic
- VERSIONS: electron: undefined; used node: 22.21.1; installed node: 22.21.1; npm: 10.9.4; pm2: 6.0.14
- ENV:      XDG_SESSION_TYPE: x11; MM_CONFIG_FILE: undefined
            WAYLAND_DISPLAY:  undefined; DISPLAY: :0.0; ELECTRON_ENABLE_GPU: undefined
- RAM:      total: 48077.13 MB; free: 35123.34 MB; used: 12953.39 MB
- OTHERS:   uptime: 23038 minutes; timeZone: America/Chicago 

### Which start option are you using?

node --run start

### Are you using PM2?

No

### Module

calendar

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

<details>
<summary>Configuration</summary>

```
<!-- Paste your configuration here. Don't forget to remove any sensitive information! -->
use the sample config.js
replace calendar url with whatever file would hold that 
```
</details>

```js
<!-- Paste relevant code here -->
```

Steps to reproduce the issue:

use this sample ics 
```text
BEGIN:VALENDAR
BEGIN:VEVENT
DTSTART;VALUE=DATE:20250414
DTEND;VALUE=DATE:20250415
RRULE:FREQ=WEEKLY;WKST=SU;INTERVAL=2
DTSTAMP:20260103T123138Z
UID:@google.com
CREATED:20250409T000033Z
DESCRIPTION:Paying half the mortgage
LAST-MODIFIED:20250409T000033Z
SEQUENCE:0
STATUS:CONFIRMED
SUMMARY:Mortgage Payment
TRANSP:TRANSPARENT
END:VEVENT
```

### What did you expect to happen?

the event is schedule for Monday 

### What actually happened?

, but shows on Sunday, repeating

full day events are local timezone only.  so altho the event LOOKS like DST (at the time it was created),
it is DST/STD time insensitive.

this was the startofday/endofday discussion had earlier

### Additional comments

reported on discord
[...]

PR #4004, opened 2026-01-03
fix(calendar): correct day-of-week for full-day recurring events across all timezones

KristjanESPERANTO:
Fixes **full-day recurring events showing on wrong day** in timezones west of UTC (reported in #4003).

**Root cause**: `moment.tz(date, eventTimezone).startOf("day")` interprets UTC midnight as local time:
- `2025-11-03T00:00:00.000Z` in America/Chicago (UTC-6)
- → Converts to `2025-11-02 18:00:00` (6 hours back)
- → `.startOf("day")` → `2025-11-02 00:00:00` ❌ **Wrong day!**

**Impact**: The bug affects:
- All timezones west of UTC (UTC-1 through UTC-12): Americas, Pacific
- Timezones east of UTC (UTC+1 through UTC+12): Europe, Asia, Africa - work correctly
- UTC itself - works correctly

The issue was introduced with commit c2ec6fc2 (#3976), which fixed the time but broke the date. This PR fixes both.

| | Result | Day | Time | Notes |
|----------|--------|-----|------|-------|
| **Before c2ec6fc2** | `2025-11-03 05:00:00 Monday` | ✅ | ❌ | Wrong time, but correct day |
| **Current (c2ec6fc2)** | `2025-11-02 00:00:00 Sunday` | ❌ (west of UTC)<br>✅ (east of UTC) | ✅ | Wrong day - visible bug! |
| **This fix** | `2025-11-03 00:00:00 Monday` | ✅ | ✅ | Correct in all timezones |

Note: While the old logic had incorrect timing, it produced the correct calendar day due to how it handled UTC offsets. The current logic fixed the timing issue but introduced the more visible calendar day bug.

### Solution

Extract UTC date components and interpret as local calendar dates:

```javascript
const utcYear = date.getUTCFullYear();
const utcMonth = date.getUTCMonth();
const utcDate = date.getUTCDate();
return moment.tz([utcYear, utcMonth, utcDate], eventTimezone);
```

### Testing

To prevent this from happening again in future refactorings, I wrote a test for it.

```bash
npm test -- tests/unit/modules/default/calendar/calendar_fetcher_utils_spec.js
```

### DIFF
--- a/modules/default/calendar/calendarfetcherutils.js
+++ b/modules/default/calendar/calendarfetcherutils.js
@@ -82,8 +82,12 @@ const CalendarFetcherUtils = {
 		// rrule.js returns UTC dates with tzid cleared, so we interpret them in the event's original timezone
 		return dates.map((date) => {
 			if (isFullDayEvent) {
-				// For all-day events, anchor to calendar day in event's timezone
-				return moment.tz(date, eventTimezone).startOf("day");
+				// For all-day events, extract UTC date components and interpret as local calendar date
+				// This prevents timezone offsets from shifting the date to the previous/next day
+				const utcYear = date.getUTCFullYear();
+				const utcMonth = date.getUTCMonth();
+				const utcDate = date.getUTCDate();
+				return moment.tz([utcYear, utcMonth, utcDate], eventTimezone);
 			}
 			// For timed events, preserve the time in the event's original timezone
 			return moment.tz(date, "UTC").tz(eventTimezone, true);
--- a/tests/unit/modules/default/calendar/calendar_fetcher_utils_spec.js
+++ b/tests/unit/modules/default/calendar/calendar_fetcher_utils_spec.js
@@ -111,5 +111,47 @@ END:VEVENT`);
 			expect(januaryFirst[0].toISOString(true)).toContain("09:00:00.000+01:00");
 			expect(julyFirst[0].toISOString(true)).toContain("09:00:00.000+02:00");
 		});
+
+		it("should return correct day-of-week for full-day recurring events across DST transitions", () => {
+			// Test case for GitHub issue #3976: recurring full-day events showing on wrong day
+			// This happens when DST transitions change the UTC offset between occurrences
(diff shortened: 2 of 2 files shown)


sdetweil:
Is the a matching edge case with one hour close to
DST which should move forward , but doesn’t

KristjanESPERANTO:
I initially thought this was DST-related too, but after testing I just updated the PR text - it's **not** about DST.

The bug affects **all** full-day events in western timezones (UTC-1 to UTC-12), regardless of DST:

```javascript
// Western timezones: Always wrong
moment.tz(new Date("2025-11-03T00:00:00Z"), "America/Chicago")
// → 2025-11-02 18:00 → .startOf("day") → 2025-11-02 ❌

// Eastern timezones: Always correct
moment.tz(new Date("2025-11-03T00:00:00Z"), "Asia/Tokyo")  
// → 2025-11-03 09:00 → .startOf("day") → 2025-11-03 ✅
```

No matching edge case exists - the bug is asymmetric. UTC midnight converts to the **previous day** in western timezones, but **same day** in eastern timezones. That's why eastern timezones work correctly by design.

KristjanESPERANTO:
I just fixed a comment that still referred to DST. No changes in the logic.

KristjanESPERANTO:
I'm too tired to continue researching right now, but after further investigation, it seems to me that the old logic before c2ec6fc is not as wrong as I described it above 🤯. But that doesn't change the fact that we need a fix (this PR) now.

I hope that we switch to Temporal at some point, then all the time zone stuff will be a little less confusing.

sdetweil:
No rush.  Not shipping til April 1

sdetweil:
User reports PR fixed issue

KristjanESPERANTO:
Great :slightly_smiling_face: Then we can merge it, right?

KristjanESPERANTO:
@sdetweil Is the problem serious enough that we should do a fix release? :thinking:

sdetweil:
no, we can teach them how to get the PR

KristjanESPERANTO:
Okay, I think that's a good call :+1:

PR #4010, opened 2026-01-10
Update node-ical and support it's rrule-temporal changes

KristjanESPERANTO:
Updating `node-ical` and adapt logic to new behaviour.

## Problem

node-ical 0.23.0 switched from `rrule.js` to `rrule-temporal`, changing how recurring event dates are returned. Our code assumed the old behavior where dates needed manual timezone conversion.

## Solution

Updated `getMomentsFromRecurringEvent()` in `calendarfetcherutils.js`:
- Removed `tzid = null` clearing (no longer needed)
- Simplified timed events: `moment.tz(date, eventTimezone)` instead of `moment.tz(date, "UTC").tz(eventTimezone, true)`
- Kept UTC component extraction for full-day events to prevent date shifts

PR #4013, opened 2026-01-11
fix(calendar): update to node-ical 0.23.1 and fix full-day recurrence lookup

KristjanESPERANTO:
Adapts calendar module to node-ical changes and fixes a bug with moved full-day recurring events in eastern timezones.

### 1. Update node-ical to 0.23.1
- Includes upstream fixes for UNTIL UTC validation errors from CalDAV servers (reported by @rejas in PR #4010)
- Changes to `getDateKey()` behavior for VALUE=DATE events (now uses local date components)
- Fixes issue with malformed DURATION values (reported by MagicMirror user here: https://github.com/jens-maus/node-ical/issues/381)

### 2. Remove dead code
- Removed ineffective UNTIL modification code (rule.options is read-only in rrule-temporal)
- The code attempted to extend UNTIL for all-day events but had no effect

### 3. Fix recurrence lookup for full-day events
node-ical changed the behavior of `getDateKey()` - it now uses local date components for VALUE=DATE events instead of UTC. This broke recurrence override lookups for full-day events in eastern timezones.

**Why it broke:**
- **before node-ical update:** Both node-ical and MagicMirror used UTC → keys matched ✅
- **after node-ical update:** node-ical uses local date (RFC 5545 conform), MagicMirror still used UTC → **mismatch** ❌

**Example:**
- Full-day recurring event on October 12 in Europe/Berlin (UTC+2)
- node-ical 0.23.1 stores override with key: `"2024-10-12"` (local date)
- MagicMirror looked for key: `"2024-10-11"` (from UTC: Oct 11 22:00)
- **Result:** Moved event not found, appears on wrong date

**Solution:** Adapt to node-ical's new behavior by using local date components for full-day events, UTC for timed events.

**Note:** This is different from previous timezone fixes - those addressed event generation, this fixes the lookup of recurrence overrides.

## Background

node-ical 0.23.0 switched from `rrule` to `rrule-temporal`, introducing breaking changes. Version 0.23.1 fixed the UNTIL validation issue and formalized the `getDateKey()` behavior for DATE vs DATE-TIME values, following RFC 5545 specification that DATE values represent local calendar dates without timezone context.

### DIFF
--- a/modules/default/calendar/calendarfetcherutils.js
+++ b/modules/default/calendar/calendarfetcherutils.js
@@ -308,7 +308,12 @@ const CalendarFetcherUtils = {
 			let recurringEventStartMoment = startMoment.clone().tz(CalendarFetcherUtils.getLocalTimezone());
 			let recurringEventEndMoment = recurringEventStartMoment.clone().add(durationMs, "ms");
 
-			const dateKey = recurringEventStartMoment.tz("UTC").format("YYYY-MM-DD");
+			// For full-day events, use local date components to match node-ical's getDateKey behavior
+			// For timed events, use UTC to match ISO string slice
+			const isFullDay = CalendarFetcherUtils.isFullDayEvent(event);
+			const dateKey = isFullDay
+				? recurringEventStartMoment.format("YYYY-MM-DD")
+				: recurringEventStartMoment.tz("UTC").format("YYYY-MM-DD");
 
 			// Check for overrides
 			if (curEvent.recurrences !== undefined) {


rejas:
Thanks @KristjanESPERANTO this fixed my calendar again

KristjanESPERANTO:
Great :slightly_smiling_face:

The switch from rrule to rrule-temporal at node-ical was certainly the right strategic move, but it does come with some migration pains.

PR #4015, opened 2026-01-16
Switch to undici Agent for HTTPS requests

in-voker:
Replace https.Agent with undici's Agent for handling self-signed certificates.

Hello and thank you for wanting to contribute to the MagicMirror² project!

**Please make sure that you have followed these 3 rules before submitting your Pull Request:**

>
no

allow the selfSignedCert: true flag in calenders array to work

>
> 3. Please run `node --run lint:prettier` before submitting so that
>    style issues are fixed.

### DIFF
--- a/modules/default/calendar/calendarfetcher.js
+++ b/modules/default/calendar/calendarfetcher.js
@@ -1,8 +1,9 @@
 const ical = require("node-ical");
 const Log = require("logger");
+const { Agent } = require("undici");
 const CalendarFetcherUtils = require("./calendarfetcherutils");
 const { getUserAgent } = require("#server_functions");
-const { Agent } = require('undici');
+
 
 const FIFTEEN_MINUTES = 15 * 60 * 1000;
 const THIRTY_MINUTES = 30 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


KristjanESPERANTO:
Thanks for addressing this! :slightly_smiling_face: I think your approach using undici's `Agent` with the `dispatcher` option is the right way to go :+1:

Please handle the linter issues. And optinal but nice to have: Could you add a test case verifying that self-signed certificates work correctly with the new implementation?

khassel:
@KristjanESPERANTO When I search for `node:https` in the project, I get two more hits. Shouldn't that be changed as well?

KristjanESPERANTO:
> When I search for `node:https` in the project, I get two more hits. Shouldn't that be changed as well?

I had a look. The other two uses should remain unchanged:

1. **`js/server.js`**: Creates an HTTPS **server** (receives connections) - this is unrelated to undici. `https.Server` is the correct API for creating servers.

2. **`clientonly/index.js`**: Uses the old callback-based `https.get()` API - not using `fetch()`. We should refactor this on occasion, but that is unrelated to this PR.

KristjanESPERANTO:
@in-voker Thanks again for this improvement! It will be in the next release :slightly_smiling_face:

sdetweil:
There are other modules that need the same fix

PR #4019, opened 2026-01-23
move default modules from /modules/default to /defaultmodules

khassel:
Since the project's inception, I've missed a clear separation between default and third-party modules.

This increases complexity within the project (exclude `modules`, but not `modules/default`), but the mixed use is particularly problematic in Docker setups.

Therefore, with this pull request, I'm moving the default modules to a different directory.

~~I've chosen `default/modules`, but I'm not bothered about it; `defaultmodules` or something similar would work just as well.~~

Changed to `defaultmodules`.

Let me know if there's a majority in favor of this change.

### DIFF
--- a/defaultmodules/updatenotification/node_helper.js
+++ b/defaultmodules/updatenotification/node_helper.js
@@ -2,7 +2,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 const NodeHelper = require("node_helper");
 
-const defaultModules = require(`${global.root_path}/defaultmodules/defaultmodules`);
+const defaultModules = require(`${global.root_path}/${global.defaultModulesDir}/defaultmodules`);
 const GitHelper = require("./git_helper");
 const UpdateHelper = require("./update_helper");
 
--- a/js/app.js
+++ b/js/app.js
@@ -13,7 +13,6 @@ global.root_path = path.resolve(`${__dirname}/../`);
 const Server = require(`${__dirname}/server`);
 const Utils = require(`${__dirname}/utils`);
 
-const defaultModules = require(`${global.root_path}/defaultmodules/defaultmodules`);
 // used to control fetch timeout for node_helpers
 const { setGlobalDispatcher, Agent } = require("undici");
 const { getEnvVarsAsObj, getConfigFilePath } = require("#server_functions");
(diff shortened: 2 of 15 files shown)


khassel:
I understand that a few third-party modules would also need to be modified, such as MMM-Config and MMM-RemoteControl.

Perhaps it's also possible to store the new relative path in `js/defaults.js` so that it can be accessed via the config file.

sdetweil:
I prefer defaultmodules as well

there is nothing else we would put in the default folder, right?

khassel:
> there is nothing else we would put in the default folder, right?

yes, everything which is now under `modules/default` would move to `defaultmodules`

grafik

sdetweil:
No. I meant is it conceivable that there are other entities besides modules we would put in a folder called default

I understand the modules 
Just trying to think of use cases

css, separating ours from users more
Anything in our 3 runtimes we would separate?
Client, server, combined

khassel:
yes, the `custom.css`, at the moment in the `css` folder which is a folder owned by this repo. I would see `custom.css` inside `config` folder. But don't want to do this in this PR ...

Another approach could be to do everything into `config` folder which is user related:
- `config.js`
- `custom.css`
- 3rd-party-modules

The last approach would allow the standard modules to remain in their current location, but I think that would be a massive change for every user, so I immediately dismissed that idea.

defaultmodules/updatenotification/node_helper.js
```diff
@@ -2,7 +2,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 const NodeHelper = require("node_helper");
 
-const defaultModules = require(`${global.root_path}/defaultmodules/defaultmodules`);
+const defaultModules = require(`${global.root_path}/${global.defaultModulesDir}/defaultmodules`);
```
KristjanESPERANTO, line 5: Should it be ``const defaultModules = require(`${global.root_path}/${global.defaultModulesDir}`);``?

khassel: the const should load the content of the file `defaultmodules.js` (which is inside `${global.root_path}/${global.defaultModulesDir}`) so I think it is correct

KristjanESPERANTO: Ah, yes, I was seemingly irritated by the fact that the directory name and file name are the same. But that's fine :-)

sdetweil:
I have updated MMM-Config to handle either default modules location

khassel:
from my side ready to merge

PR #4020, opened 2026-01-27
move custom.css from css to config

khassel:
This is another change to cleanup structure, already mentioned in https://github.com/MagicMirrorOrg/MagicMirror/pull/4019#issuecomment-3792953018

After separating default and 3rd-party modules this PR moves the `custom.css` from the mm-owned directory `css` into user owned directory `config`.

It has a built-in function which moves the `css/custom.css` to the new location `config/custom.css` (if the target not exists).

Let me know if there's a majority in favor of this change.


KristjanESPERANTO:
Perhaps instead of automatically moving it, outputting a deprecation warning if custom.css is still in its old location? This would have the advantage that users would have to do this consciously by themself.

khassel:
I don't care at all; I've implemented the most convenient solution for now.

Let the majority decide.

A warning is already displayed; you can change the text, abort with an error message, or...

sdetweil:
I’m ok with moving it. And moving if detected in old place

rejas (approved):

giddyhup:
Late to the party but I ran into this. In my fresh install it looks like the css directory is still there. It would have helped me if a readme.txt in it would have pointed out the new structure.

PR #4022, opened 2026-01-28
refactor(clientonly): modernize code structure and add comprehensive tests

KristjanESPERANTO:
This PR improves `clientonly` start option with better code structure, validation, and comprehensive test coverage.

### Changes

**Refactoring:**
- Improved parameter handling with explicit function parameter passing instead of closure
- Added port validation (1-65535) with proper NaN handling
- Removed unnecessary IIFE wrapper (Node.js modules are already scoped)
- Extracted `getCommandLineParameter` as a reusable top-level function
- Enhanced error reporting with better error messages
- Added connection logging for debugging

**Testing:**
- Added comprehensive e2e tests for parameter validation
- Test coverage for missing/incomplete parameters
- Test coverage for local address rejection (localhost, 127.0.0.1, ::1, ::ffff:127.0.0.1)
- Test coverage for port validation (invalid ranges, non-numeric values)
- Test coverage for TLS flag parsing
- Integration test with running server

### Testing

All tests pass:
```bash
npm test -- tests/e2e/clientonly_spec.js
# ✓ 18 tests passed

### DIFF
--- a/clientonly/index.js
+++ b/clientonly/index.js
@@ -3,165 +3,165 @@
 const http = require("node:http");
 const https = require("node:https");
 
-// Use separate scope to prevent global scope pollution
-(function () {
+/**
+ * Get command line parameters
+ * Assumes that a cmdline parameter is defined with `--key [value]`
+ *
+ * example: `node clientonly --address localhost --port 8080 --use-tls`
+ * @param {string} key key to look for at the command line
+ * @param {string} defaultValue value if no key is given at the command line
+ * @returns {string} the value of the parameter
+ */
+function getCommandLineParameter (key, defaultValue = undefined) {
+	const index = process.argv.indexOf(`--${key}`);
+	const value = index > -1 ? process.argv[index + 1] : undefined;
+	return value !== undefined ? String(value) : defaultValue;
+}
 
-	/**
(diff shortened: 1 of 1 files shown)


khassel (approved):

khassel:
Not directly related to this pull request, but I've been wondering whether we need these subdirectories for `clientonly` and `serveronly`. We could put everything under `js` and simply change the call in `package.json` for `serveronly`. And we could do the same for `clientonly`.

But maybe I've overlooked something that argues against it...

KristjanESPERANTO:
I'm a bit divided on this.

For client-only users, this would be a breaking change. Until now, they have been using
`node clientonly --address 192.168.1.5 --port 8080`
to start. After a migration like suggested, it would be
`node js/client.js --address 192.168.1.5 --port 8080` or
`node --run client -- --address 192.168.1.5 --port 8080`.

For server users it would stay the same command.

khassel:
yes, I know that it would be a breaking change for client users, so we better don't change it.

KristjanESPERANTO:
Yes, if we had strong arguments for a breaking change, we could certainly make the change, but I don't see them here yet.

PR #4024, opened 2026-01-29
Update compliments.js

SaeedHurzuk:
- Boosted the mirror's ego with better compliments.
- Upgrade compliment variety and tone

### DIFF
--- a/defaultmodules/compliments/compliments.js
+++ b/defaultmodules/compliments/compliments.js
@@ -4,11 +4,34 @@ Module.register("compliments", {
 	// Module config defaults.
 	defaults: {
 		compliments: {
-			anytime: ["Hey there sexy!"],
-			morning: ["Good morning, handsome!", "Enjoy your day!", "How was your sleep?"],
-			afternoon: ["Hello, beauty!", "You look sexy!", "Looking good today!"],
-			evening: ["Wow, you look hot!", "You look nice!", "Hi, sexy!"],
-			"....-01-01": ["Happy new year!"]
+			anytime: [
+				"Unstoppable energy today!",
+				"Confidence looks incredible on you.",
+				"The room lights up when you walk in.",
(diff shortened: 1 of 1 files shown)


sdetweil:
Why? You can put yours in config.js to show anything you want

sdetweil:
Why? You can put yours in config.js to show anything you want

and you can put them in a file instead of config.js and update them live. No need to update the default sayings

sdetweil:
MagicMirror design is everything in config.js overrides what is in the modules.

SaeedHurzuk:
> Why? You can put yours in config.js to show anything you want
> 
> and you can put them in a file instead of config.js and update them live. No need to update the default sayings

Give's it fresh new compliments, i noticed it has not been updated in ages.

rejas:
> Give's it fresh new compliments, i noticed it has not been updated in ages.

I agree with Sam, this doesnt need it updating. Thanks for your interest in MM though!

KristjanESPERANTO:
I'm wondering: could there be value in refreshing the defaults? Not everyone customizes immediately, and some new users will see these defaults for weeks or months. Since they haven't changed in a long time, a refresh could improve that first impression - though I admit I can't judge the specific wording as a non-native speaker.

I don't want to drag this out, but I wanted to share that perspective. I'm also perfectly fine with not pursuing this further. Sam's argument is totally valid.

sdetweil:
I think the current values would motivate users to learn how to change them.  Which is one of our goals

KristjanESPERANTO:
> I think the current values would motivate users to learn how to change them. Which is one of our goals

Oh, that's a very good point :+1: 

Okay, then I hereby close this PR. @SaeedHurzuk, thank you for your efforts! Don't let this discourage you from suggest other changes :slightly_smiling_face:

PR #4028, opened 2026-02-05
calendar.js: remove useless hasCalendarURL function

khassel:
Found this while debugging.

The `hasCalendarURL` function does a check if the url is in the config. But the calendar sees only his own config part, so this check is always true (tested with more than one calendar module in `config.js`).


KristjanESPERANTO:
Good catch! :+1: :slightly_smiling_face:

PR #4029, opened 2026-02-05
change loading config.js, allow variables in config.js and try to protect sensitive data

khassel:
### Previously

Loaded on server-side in `app.js` and in the browser by including `config.js` in `index.html`. The web server has an endpoint `/config` providing the content of server loaded `config.js`.

### Now

Loaded only on server-side in `app.js`. The browser loads the content using the web server endpoint `/config`. So the server has control what to provide to the clients.

Loading the `config.js` was moved to `Utils.js` so that `check_config.js` can use the same functions.

### Previously

Environment variables were not allowed in `config.js`. The workaround was to create a `config.js.template` with curly braced bash variables allowed. While starting the app the `config.js.template` was converted via `envsub` into a `config.js`.

### Now

Curly braced bash variables are allowed in `config.js`. Because only the server loads `config.js` he can substitute the variables while loading.

## Secrets in MagicMirror²

To be honest, this is a mess.

### Previously

All content defined in the `config` directory was reachable from the browser. Everyone with access to the site could see all stuff defined in the configuration e.g. using the url http://ip:8080/config. This included api keys and other secrets.

So sharing a MagicMirror² url to others or running MagicMirror² without authentication as public website was not possible.

### Now

With this PR we add (beta) functionality to protect sensitive data. This is only possible for modules running with a `node_helper`. For modules running in the browser only (e.g. default `weather` module), there is no way to hide data (per construction). This does not mean, that every module with `node_helper` is safe, e.g. the default `calendar` module is not safe because it uses the calendar url's as sort of id and sends them to the client.

For adding more security you have to set `hideConfigSecrets: true` in `config.js`. With this:
- `config/config.env` is not deliverd to the browser
- the contents of environment variables beginning with `SECRET_` are not published to the clients

This is a first step to protect sensitive data and you can at least protect some secrets.


KristjanESPERANTO:
Wow, that's great! :clap:

I'm in the process of rewriting the weather modules to be server-side. That would also make them a bit more secure :slightly_smiling_face:

sdetweil:
Nice!

PR #4033, opened 2026-02-08
allow environment variables in cors urls

khassel:
and centralize and optimize replace regex.

Another follow up to #4029 

With this PR you can use secrets in urls in browser modules if you use the cors proxy.

I will provide a new section in the docs concerning secret handling before next release.


KristjanESPERANTO:
This looks great! Would it make sense to add a test for the new replaceSecretPlaceholder function?

khassel:
> This looks great! Would it make sense to add a test for the new replaceSecretPlaceholder function?

done ...

PR #4039, opened 2026-02-27
fix(newsfeed): fix full article view and add framing check

KristjanESPERANTO:
I was playing around with the newsfeed notification system (`ARTICLE_MORE_DETAILS`, `ARTICLE_TOGGLE_FULL`, …) and discovered some issues with the full article view:

The iframe was loading the CORS proxy URL instead of the actual article URL, which could cause blank screens depending on the feed. Also, many news sites block iframes entirely (`X-Frame-Options: DENY`) and the user got no feedback at all — just an empty page. On top of that, scrolling used `window.scrollTo()` which moved the entire MagicMirror page instead of just the article.

This PR cleans that up:

- Use the raw article URL for the iframe (CORS proxy is only needed for server-side feed fetching)
- Check `X-Frame-Options` / `Content-Security-Policy` headers server-side before showing the iframe — if the site blocks it, show a brief "Article cannot be displayed here." message and return to normal view
- Show the iframe as a fixed full-screen overlay so other modules aren't affected, scroll via `container.scrollTop`
- Keep the progressive disclosure behavior for `ARTICLE_MORE_DETAILS` (title → description → iframe → scroll)
- Delete `fullarticle.njk`, replace with `getDom()` override
- Fix `ARTICLE_INFO_RESPONSE` returning proxy URL instead of real URL
- A few smaller fixes (negative scroll, null guard)
- Add `NEWSFEED_ARTICLE_UNAVAILABLE` translation to all 47 language files
- Add e2e tests for the notification handlers (`ARTICLE_NEXT`, `ARTICLE_PREVIOUS`, `ARTICLE_INFO_REQUEST`, `ARTICLE_LESS_DETAILS`)

## What this means for users

- The full article view now works reliably across different feeds
- If a news site blocks iframes, the user sees a brief message instead of a blank screen
- Additional e2e tests make the module more robust and less likely to break silently in future MagicMirror versions

### DIFF
--- a/tests/configs/modules/newsfeed/notifications.js
+++ b/tests/configs/modules/newsfeed/notifications.js
@@ -1,0 +1,26 @@
+let config = {
+	address: "0.0.0.0",
+	ipWhitelist: [],
+	timeFormat: 12,
+
+	modules: [
+		{
+			module: "newsfeed",
+			position: "bottom_bar",
+			config: {
+				feeds: [
+					{
+						title: "Rodrigo Ramirez Blog",
+						url: "http://localhost:8080/tests/mocks/newsfeed_test.xml"
+					}
+				],
+				updateInterval: 3600 * 1000 // 1 hour - prevent auto-rotation during tests
+			}
+		}
+	]
+};
+
+/*************** DO NOT EDIT THE LINE BELOW ***************/
+if (typeof module !== "undefined") {
+	module.exports = config;
+}
--- a/tests/e2e/modules/newsfeed_spec.js
+++ b/tests/e2e/modules/newsfeed_spec.js
@@ -70,6 +70,148 @@ const runTests = () => {
 	});
 };
 
+describe("Newsfeed module > Notifications", () => {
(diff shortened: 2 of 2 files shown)


sdetweil:
Nice work. 
one investigative question 

why not use the jdk file? 

Supposedly it’s as functional as direct dom.  
just a question. I have used similar functionality via Angular

KristjanESPERANTO:
> why not use the jdk file?

The full article view now has an async step (server checks headers before showing the iframe), so the DOM needs to react to a callback. With a template you'd render it and then immediately manipulate the result via querySelector anyway — at that point the template doesn't add much. The old one was only 3 lines, so moving it into getDom() felt simpler.

rejas:
No complains from my side other then the question if the translations were done with AI?

KristjanESPERANTO:
> if the translations were done with AI?

Yes, I used [Gemini](https://gemini.google.com/) for the translation. I will try to include this information in future PR texts.

PR #4040, opened 2026-03-01
chore: update dependencies + add exports, files, and sideEffects fields to package.json

KristjanESPERANTO:
This updates all dependencies to their latest versions - except two packages:

- eslint: Some plugins we use here aren't compatible yet to ESLint v10.
- node-ical: The new version has revealed an issue in our calendar logic. I would prefer to address this in a separate PR.

After updating the dependencies, eslint-plugin-package-json rules complained about missing fields in the package.json. I added them in the second commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,18 @@
 		}
 	},
 	"main": "js/electron.js",
+	"exports": "./js/electron.js",
+	"files": [
+		"clientonly",
+		"css",
+		"defaultmodules",
+		"js",
+		"serveronly",
+		"translations",
+		"favicon.svg",
+		"index.html"
+	],
+	"sideEffects": true,
 	"scripts": {
 		"config:check": "node js/check_config.js",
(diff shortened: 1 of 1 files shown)


KristjanESPERANTO:
I'll check the failing tests tomorrow 🤔

KristjanESPERANTO:
Checked and resolved by rolling back the update for node-ical.

KristjanESPERANTO:
These fields are only relevant for users who install MagicMirror via npm (`npm i magicmirror`) and use it programmatically — an edge case, but with these benefits:

- **`exports`** – Node.js resolves the entry point cleanly and prevents unintended access to internal files outside the defined entry point.
- **`files`** – Reduces the installed package size: only the necessary runtime files are included, excluding dev configs, tests, CI files, etc.
- **`sideEffects: true`** – Prevents bundlers (webpack/rollup) from incorrectly tree-shaking code that has side effects on import (e.g. global registrations in Electron).

PR #4032, opened 2026-02-08
[weather] refactor: migrate to server-side providers with centralized HTTPFetcher

KristjanESPERANTO:
This migrates the Weather module from client-side fetching to use the server-side centralized HTTPFetcher (introduced in #4016), following the same pattern as the Calendar and Newsfeed modules.

## Motivation

This brings consistent error handling and better maintainability and completes the refactoring effort to centralize HTTP error handling across all default modules.

Migrating to server-side providers with HTTPFetcher brings:
- **Centralized error handling**: Inherits smart retry strategies (401/403, 429, 5xx backoff) and timeout handling (30s)
- **Consistency**: Same architecture as Calendar and Newsfeed modules
- **Security**: Possibility to hide API keys/secrets from client-side
- **Performance**: Reduced API calls in multi-client setups - one server fetch instead of one per client
- **Enabling possible future features**: e.g. server-side caching, rate limit monitoring, and data sharing with third-party modules

## Changes

- All 10 weather providers now use HTTPFetcher for server-side fetching
- Consistent error handling like Calendar and Newsfeed modules

## Breaking Changes

None. Existing configurations continue to work.

## Testing

To ensure proper functionality, I obtained API keys and credentials for all providers that require them. I configured all 10 providers in a carousel setup and tested each one individually. Screenshots for each provider are attached below demonstrating their working state.

I even requested developer access from the Tempest/WeatherFlow team to properly test this provider.

**Comprehensive test coverage**: A major advantage of the server-side architecture is the ability to thoroughly test providers with unit tests using real API response snapshots. Don't be alarmed by the many lines added in this PR - they are primarily test files and real-data mocks that ensure provider reliability.

## Review Notes

I know this is an enormous change - I've been working on this for quite some time. Unfortunately, breaking it into smaller incremental PRs wasn't feasible due to the interdependencies between providers and the shared architecture.

Given the scope, it's nearly impossible to manually review every change. To ensure quality, I've used both CodeRabbit and GitHub Copilot to review the code multiple times in my fork, and both provided extensive and valuable feedback. Most importantly, my test setup with all 10 providers working successfully is very encouraging.

## Related

Part of the HTTPFetcher migration #4016.

## Screenshots

Ekrankopio de 2026-02-08 13-06-54
Ekrankopio de 2026-02-08 13-07-02
Ekrankopio de 2026-02-08 13-07-07
Ekrankopio de 2026-02-08 13-07-12
Ekrankopio de 2026-02-08 13-07-17
[...]

PR #4046, opened 2026-03-01
Use getDateString in openmeteo

### LINKED ISSUE #4045: [Bug] Openmeteo weather forecast stuck in the past
### Environment

Help, my mm thinks its yesterday :-)

node --run start

### Are you using PM2?

No

### Module

None

- [x] Yes
- [ ] No

- [x] Yes

### What did you do?

Configuration:

```

		{
			module: "weather",
			//disabled: true,
			position: "top_left",
			config: {
				weatherProvider: "openmeteo",
				fade: false,
				ignoreToday: false,
				type: "forecast",
				location: "Berlin",
				lat: 52.5012,
				lon: 13.4789
			}
		},

```

it's sunday

### What actually happened?

it's saturday

### Additional comments

_No response_

### Participation

rejas:
Fixes #4045 

Otherwise the calculation comes to the result that its yesterday...

### DIFF
--- a/defaultmodules/weather/providers/openmeteo.js
+++ b/defaultmodules/weather/providers/openmeteo.js
@@ -270,13 +270,13 @@ class OpenMeteoProvider {
 		const endDate = new Date(startDate);
 		endDate.setDate(endDate.getDate() + Math.max(0, Math.min(7, maxNumberOfDays)));
 
-		params.start_date = startDate.toISOString().split("T")[0];
+		params.start_date = getDateString(startDate);
 
 		switch (this.config.type) {
 			case "hourly":
 			case "daily":
 			case "forecast":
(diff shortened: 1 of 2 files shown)


KristjanESPERANTO:
Oh, this was introduced by my rewrite in #4032 – the migration from `moment.format("YYYY-MM-DD")` (local time) to `.toISOString().split("T")[0]` (UTC) was a careless mistake. I also checked `envcanada.js` and `smhi.js`, which also use `toISOString()`, but there it's intentional – UTC hour extraction for API URLs and timestamp storage – so they should be fine.

The new test is really nice: mocking `toISOString` to return a wrong date and asserting that the URL params are still correct is a good approach for this kind of regression.

Issue #2909, opened 2022-09-11
incorporate remote njk template file configuration for weather


Kayakbabe:
I want to do a pull request so I can modify the weather module.
By changing it slightly, it will not affect users if they update, they will not have to edit their configs.. existing configs and future updates to MM will continue to work with my modification.

Now that templating is a real thing for MagicMirror,  I think that having a config variable for a custom weather template would make updating less stressful and easier for those who have modified their weather. Custom templates can have different names than the default current.njk and forcast.njk and could even be placed in the config folder if a path is given for the template file.

To that end I have modified my local weather module instance as follows.

In the weather module weather.js defaults section I added

	defaults: {
		remoteTemplate: "current.njk",

and I altered the getTemplate function to the following:

	// Select the template depending on the display type.
	getTemplate: function () {
		
		if (this.config.remoteTemplate !="") {return this.config.remoteTemplate}
		 else {
		
			switch (this.config.type.toLowerCase()) {
				case "current":
					return "current.njk";
				case "hourly":
					return "hourly.njk";
				case "daily":
				case "forecast":
					return "forecast.njk";
				//Make the invalid values use the "Loading..." from forecast
				default:
					return "forecast.njk";
			}
	   }
	},

If the config value is not placed in config.js, the module will not fail, it will get the template as it does without the modification.

However, this modification allows using a custom template that will not be overwritten by updates.

Benefit of this mod is being able to change the css and html layout without having to create a new module. A user can put their own template in the same folder as current.njk and forecast.njk or as I did they can use a relative or absolute path in the config.

my usage of this mode.
I created mycurrent.njk and myforcast.njk and put them in the mm config folder. 

Then,
in the weather section of my config.js
I added the relative path to my custom template file.

	config: {
		remoteTemplate: "../../../../config/kellyforecast.njk",
	

**MagicMirror² Version**: 2.20.0

rejas:
Hi @Kayakbabe this sounds like an interesting and helpful thing. Do you have the PR already in progress or need some help?

rejas:
by the way: is that a duplicate of https://github.com/MichMich/MagicMirror/issues/2880 ? can one be closed

rejas:
Yes, in the long run this might be interesting for other modules as well. 

But I would propose to first do it in this here since @Kayakbabe seems to be invested in this module. That would not bloat the scope of a potential PR and allow to see how it might be forged into a more general approach later.

khassel:
[...]

PR #4051, opened 2026-03-06
weather: add possibility to override njk's and css

khassel:
This is an approach for #2909

It adds 2 values to the config:

```js
		themeDir: "./",
		themeCustomScripts: []
```

`themeDir` must be specified relative to the weather dir.

Example config:

```js
		{
			module: "weather",
			position: "top_center",
			config: {
				weatherProvider: "openmeteo",
				type: "current",
				lat: 40.776676,
				lon: -73.971321,
				themeDir: "../../../modules/MyWeatherTemplate/",
				themeCustomScripts: [ "skycons.js", "weathertheme.js" ],
			}
		},
```

The `themeDir` must contain the 4 files
- current.njk
- forecast.njk
- hourly.njk
- weather.css

You can add more files (if needed) and add them to the `themeCustomScripts` so they are loaded as script.

There are 2 methods inserted which are called if defined:
- initWeatherTheme: For doing special things when starting the module
- updateWeatherTheme: For doing special things before updating the dom

I see this as a simple approach for overriding the default njk templates and css. I did already convert my [MMM-WeatherBridge](https://gitlab.com/khassel/MMM-WeatherBridge) into a template.

### DIFF
--- a/defaultmodules/weather/weather.js
+++ b/defaultmodules/weather/weather.js
@@ -104,7 +104,7 @@ Module.register("weather", {
 		// All providers run server-side: generate unique instance ID and initialize via node_helper
 		this.instanceId = `${this.identifier}_${Date.now()}`;
 
-		if (window.initWeatherTheme) window.initWeatherTheme();
+		if (window.initWeatherTheme) window.initWeatherTheme(this);
 
 		Log.log(`[weather] Initializing server-side provider with instance ID: ${this.instanceId}`);
 
@@ -254,7 +254,7 @@ Module.register("weather", {
 			window.updateWeatherTheme(this);
 		} else {
 			this.updateDom(300);
-		};
+		}
 
 		const currentWeather = this.currentWeatherObject;
 


sdetweil:
Weather doesn’t use njk in next release

khassel:
I used last `develop` as base, there are still njk's.

sdetweil:
Weird. I thought he removed them, we had that conversation..

khassel:
weather was moved from browser-only to using server-side but the "gui" was not touched

KristjanESPERANTO:
Nice! I think it would be great to have this possibility 🙌

One question without digging into the changes: in the linked issue #2909, @sdetweil suggested a simpler approach – just adding config variables for the template file names and css, defaulting to the current values, something like:

```js
currentNjk: "defaultmodules/weather/current.njk",
hourlyNjk: "defaultmodules/weather/hourly.njk",
forecastNjk: "defaultmodules/weather/forecast.njk",
weatherCss: "defaultmodules/weather/weather.css",
```

Users could then override only what they need with a path relative to the MM root, e.g.:

```js
currentNjk: "config/mycurrent.njk",
```

That would already allow users to point to custom files without touching the module code.

Would that cover the main use case, or do you think the `themeDir` + `themeCustomScripts` + `window.*` hooks are genuinely needed for what you have in mind? Just want to understand your approach :slightly_smiling_face:

KristjanESPERANTO:
> You can add more files (if needed) and add them to the themeCustomScripts so they are loaded as script.

Ah, that wouldn't be possible with the simpler approach... Right?

khassel:
I think to define another dir is a simpler approach instead of 4 variables.

You find my (testing) example [here](https://gitlab.com/khassel/MMM-WeatherBridge/-/tree/develop?ref_type=heads), I need 2 additional scripts `skycons.js` and `weathertheme.js` from that repo.

Can explain more tomorrow if needed (now too late) ...

KristjanESPERANTO:
I think I'm starting to understand where you are going with this – looking at MMM-WeatherBridge it makes more sense and gives impressive flexibility 👍

One concern though: wouldn't `window.initWeatherTheme` and `window.updateWeatherTheme` cause problems when having multiple weather instances with different themes? The second instance would overwrite the globals defined by the first, wouldn't it?

khassel:
think so, had no better idea as `window` ...

defaultmodules/weather/weather.js
```diff
@@ -98,6 +104,8 @@ Module.register("weather", {
 		// All providers run server-side: generate unique instance ID and initialize via node_helper
 		this.instanceId = `${this.identifier}_${Date.now()}`;
 
+		if (window.initWeatherTheme) window.initWeatherTheme();
```
KristjanESPERANTO: Shouldn't this pass `this` as well, for consistency with `updateWeatherTheme` and so the theme script knows which instance it belongs to?

```suggestion
		if (window.initWeatherTheme) window.initWeatherTheme(this);
```

KristjanESPERANTO:
> think so, had no better idea as `window` ...

Me neither, but maybe that's (multiple weather instances with different themes) an edge case that we don't need to support and should just accept as a known limitation.

khassel:
> edge case that we don't need to support and should just accept as a known limitation.

yes, and the user can implement the multiple instances stuff inside the 2 procedures if needed.

KristjanESPERANTO:
Looks good to me! :slightly_smiling_face: 

One thought: now that the infrastructure is in place, would it make sense to already ship some themes in core (e.g. under `modules/default/weather/themes/`)? The MMM-WeatherBridge layout would be a great candidate and would give users something to work with right away.

On the other hand, bundled themes also mean ongoing maintenance responsibility for us. Maybe better to merge this as-is and see if the community picks it up first?

khassel:
> Maybe better to merge this as-is and see if the community picks it up first?

I think so. I didn't want to burden this repository with my weather setup as well.

The idea is to add a detailed example to the documentation after the merge.

KristjanESPERANTO:
Nice! I'll definitely play around with it, as it has the potential to make many third-party weather modules unnecessary.

PR #4052, opened 2026-03-07
chore: update ESLint and plugins, simplify config, apply new rules

KristjanESPERANTO:
This PR updates ESLint and the ESLint plugins to their latest versions and takes advantage of the new versions to simplify the config.

The main cleanup: removed all explicit `plugins: {}` registrations from `eslint.config.mjs`. When passing direct config objects like `js.configs.recommended`, the plugin registration is already included – we were just doing it twice.

Two lint warnings are also fixed:
- A wrong import style for `eslint-plugin-package-json` (named vs. default)
- `playwright/no-duplicate-hooks` is disabled for e2e tests – the rule doesn't handle plain `beforeAll()`/`afterAll()` (Vitest style) correctly and produces false positives. I've created an issue for that: https://github.com/mskelton/eslint-plugin-playwright/issues/443.

Built-in Node.js imports were manually updated to use the `node:` prefix (e.g. `require("fs")` → `require("node:fs")`). Minor formatting fixes were applied automatically by `eslint --fix`.

### DIFF
--- a/js/alias-resolver.js
+++ b/js/alias-resolver.js
@@ -3,7 +3,7 @@
 // For a future ESM migration, replace this with a public export/import surface.
 
 const path = require("node:path");
-const Module = require("module");
+const Module = require("node:module");
 
 const root = path.join(__dirname, "..");
 
--- a/serveronly/watcher.js
+++ b/serveronly/watcher.js
@@ -1,11 +1,11 @@
 // Load lightweight internal alias resolver to enable require("logger")
 require("../js/alias-resolver");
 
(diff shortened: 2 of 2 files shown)


rejas:
So, how much AI did you use on this and the bug report?

KristjanESPERANTO:
> So, how much AI did you use on this and the bug report?

None for the PR - just Deepl translate for some wording.

For the bug report I used AI, I described in German the problem I saw and had the text formulated. On the result I did some fine-tuning and code snippets.

PR #4056, opened 2026-03-11
refactor: simplify internal `require()` calls

KristjanESPERANTO:
Remove unnecessary `__dirname` template-literal prefix from relative `require()` paths. Node.js resolves relative require() paths correctly without it.

While this may look like nitpicking: `require("./server")` is transparent to static analysis tools - IDEs can resolve the path and support go-to-definition. Template literals with `__dirname` are opaque to them. It also removes another usage of `__dirname`, which has no native equivalent in ESM and would need to be replaced there anyway (when we switch to ESM anytime in the future).

The import reordering is a side effect: `import-x/order` treats template-literal `require()` calls differently from plain strings, so the previous order was no longer valid.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -9,11 +9,12 @@ const Log = require("logger");
 // global absolute root path
 global.root_path = path.resolve(`${__dirname}/../`);
 
-const Server = require(`${__dirname}/server`);
-const Utils = require(`${__dirname}/utils`);
-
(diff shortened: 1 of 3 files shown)


khassel:
test fails because of #4055 ?

Will you rebase so the tests are green before merging?

KristjanESPERANTO:
Yes. I'll rebase tomorrow 🙂

PR #4059, opened 2026-03-13
fix(calendar): make showEnd behavior more consistent across time formats

### LINKED ISSUE #4053: [Change Request] support 'showEnd' for fullDayEvents with multiple day
### What problem do you want to solve with this change?

Currently the `showEnd` option doesn't show the end date of multiple-days events. The only case that the end date is seen is when the `timeFormat` is set to `absolute` and an event is not a fullday event.

Image

> timeFormat: "relative"

Image

> timeFormat: "absolute"

### What do you think is the correct solution?

It would be nicer if we can see the end date for multiple-days events, even with `relative` timeFormat or when the event is a fullday event.

Image

### Participation

### Additional comments

_No response_

KristjanESPERANTO:
Closes #4053

This started as a small fix.
After feedback and more debugging, I found more issues and inconsistencies and went a bit down the rabbit hole.
The PR is bigger now, but I think the result is better: behavior is more predictable, and the output is more consistent.

## Changes

- Multi-day full-day events now show an end date in `relative` and `dateheaders` when `showEnd` is enabled.
- With `absolute` + `nextDaysRelative`, full-day events now keep the end date when the start is replaced by TODAY/TOMORROW/etc.
- Timed events in `absolute` now also respect `showEndsOnlyWithDuration`.
- Tests were expanded and refactored to cover more showEnd cases and keep the setup easier to maintain.
- I also refactored parts of the calendar time rendering to reduce duplicate logic.

## Before

before

## After

after


kkangshawn:
Hi @KristjanESPERANTO , thank you for your interest and work. I verified your patch and confirmed it works perfectly with fullday event. However, this doesn't show the end for the 'Multiday events with time(full day not checked)', e.g. from 10:00AM 30.03.2026 to 11:00AM 02.04.2026 .
I'd appreciate it if you also consider this kind of events.

KristjanESPERANTO:
Thanks for testing and for the feedback! :slightly_smiling_face: 

I just added showEnd handling for timed multi-day events in relative/dateheaders as well. The exact end output is controlled by `dateEndFormat` (same behavior as in absolute mode).

By default `dateEndFormat` is `LT`, so only the end time is shown. If you want end date + time, set for example:
`dateEndFormat: "MMM Do, HH:mm"`

bildo

What do you think of it?

defaultmodules/calendar/calendar.js
```diff
@@ -501,11 +526,21 @@ Module.register("calendar", {
...
+								if (!eventStartDateMoment.isSame(endMomentAdjusted, "d")) {
+									timeWrapper.innerHTML += `-${CalendarUtils.capFirst(endMomentAdjusted.format(this.config.fullDayEventDateFormat))}`;
+								}
+							}
 							Log.info("[calendar] event fullday");
 						} else if (eventStartDateMoment.diff(now, "h") < this.config.getRelative) {
 							Log.info("[calendar] not full day but within getRelative size");
 							// If event is within getRelative hours, display 'in xxx' time format or moment.fromNow()
 							timeWrapper.innerHTML = `${CalendarUtils.capFirst(eventStartDateMoment.fromNow())}`;
+						} else if (this.config.showEnd && (!this.config.showEndsOnlyWithDuration || event.startDate !== event.endDate)) {
```
kkangshawn: Thank you @KristjanESPERANTO for the quick response! And additional functionality for dateEndFormat as well.
This works perfectly with the corner case which is the multiple day with specific time values.

Luckily or not, I found this line affects the existing behavior. The single day event is showing the end date even though the start date and the end date are same.

Image

I am not sure if this is expected behavior but this line seems like it is supposed to show the end date when the startDate and endDate are different. From my debugging, the event.startDate !== event.endDate returns always true even if the dates are same. (Please see the reds above)

For draft, I checked below code fits this case. I'd appreciate it if you check this once again.
```
} else if (this.config.showEnd && (!this.config.showEndsOnlyWithDuration && !eventStartDateMoment.isSame(eventEndDateMoment, "d"))) {
```

KristjanESPERANTO: I see your point. I need to think this through more deeply.

KristjanESPERANTO: @kkangshawn Can you have another look :slightly_smiling_face:

kkangshawn: This works PERFECTLY. Thank you so much.
P.S. The new `dateEndFormat` distinguishing timed multiday from timed sameday is absolutely cool!

KristjanESPERANTO: Nice. Thanks for testing! :smiley: 

Now let's wait for a review of another core maintainer.

khassel (approved):

KristjanESPERANTO:
@khassel Thanks for merging!

@kkangshawn Now that the merge is complete, I’d like to encourage you to use the develop branch on your mirror (at least until the next release) to see if these (and other) changes have caused any regressions :slightly_smiling_face:

Issue #4063, opened 2026-03-22
[Bug] Openmeteo weather forecast stuck in the past


rejas:
### Environment

MagicMirror² version: develop

### Which start option are you using?

node --run start

### Are you using PM2?

No

### Module

weather

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

<details>
<summary>Configuration</summary>

```
		{
			module: "weather",
			//disabled: true,
			position: "top_left",
			config: {
				weatherProvider: "openmeteo",
				appendLocationNameToHeader: false,
				fade: false,
				ignoreToday: false,
				type: "forecast",
				location: "Berlin",
				lat: 52.5012,
				lon: 13.4789
			}
		}
```
</details>

Steps to reproduce the issue:

Start the weather forecast, wait a day, still on the same day as before.

### What did you expect to happen?

Weather forecast gets updated each day

### What actually happened?

On sunday it still thinks its thursday 

![Image](https://github.com/user-attachments/assets/4c086144-e73a-4a05-b1b6-06c4208cde44)

### Additional comments

Very likely the url for fetching the weather is only updated on startup of mm. So the start-data parameter is never updated, sicne the url stays the same for each call. 

```js

	#initializeFetcher () {
		const url = this.#getUrl();
...

	#getUrl () {
		return `${this.config.apiBase}/forecast?${this.#getQueryParameters()}`;
	}


```

Maybe other fetcher users have the same problem?

### Participation

- [ ] I am willing to submit a pull request for this change.

KristjanESPERANTO:
Oh, I introduced that one with #4032. Sorry! I've already almost finished the fix. Just a few final test left.

htoonar1737-byte:
5000

sdetweil:
@htoonar1737-byte 
If you pull the develop branch you can get this fix
See this for instructions 

https://forum.magicmirror.builders/topic/14327/testing-new-fixes-or-solving-current-problems-with-next-release-code

htoonar1737-byte:
20000

PR #4065, opened 2026-03-22
fix(weather): fix weathergov forecast day labels off by one

KristjanESPERANTO:
While looking at all weather providers for #4063, I noticed another unrelated bug: The weathergov forecast has been off by one day. This has been since its first commit in 2019 and was not caused by the recent weather refactor.

The provider built the days array correctly, but then threw away the first entry (`slice(1)`) because it was considered "incomplete". The problem: the template uses index position to decide what to label "Today" and "Tomorrow" — so dropping today made Monday show up as "Today", Tuesday as "Tomorrow", and one day disappeared entirely.

The fix is a one-liner: remove `slice(1)`. Today's entry now shows min/max from the remaining hours of the day, which is the right behaviour anyway. In my understanding it's not a problem at all that the first day is incomplete.

## Before

Notice in the screenshot that it is Sunday. So after today and tomorrow, Tuesday should come next, but it says **Wednesday** instead.

Ekrankopio de 2026-03-22 14-37-55

## After

Ekrankopio de 2026-03-22 14-38-34

### DIFF
--- a/tests/unit/modules/default/weather/providers/weathergov_spec.js
+++ b/tests/unit/modules/default/weather/providers/weathergov_spec.js
@@ -241,6 +241,36 @@ describe("WeatherGovProvider", () => {
 			expect(day).toHaveProperty("weatherType");
 			expect(day).toHaveProperty("precipitationProbability");
 		});
+
+		it("should not skip the first day of forecast data", async () => {
+			const provider = new WeatherGovProvider({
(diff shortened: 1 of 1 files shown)


rejas:
Just a small thing: Can a test be added?

KristjanESPERANTO:
I just added a regression test.

PR #4068, opened 2026-03-31
update dependencies

khassel:
last update before release


khassel:
what can we do here if the dependency review fails?

KristjanESPERANTO:
Good question. If I understand correctly, the new version of markdownlint-cli2 introduces a dependency with a vulnerability. I see two options:

1. Don't update markdownlint-cli2
2. Proceed with the update despite the missing test, since it's “only” a devDependency

I'm leaning toward the second option and will keep an eye on that in the next PRs.

KristjanESPERANTO:
Oh! What did you change? :-)

khassel:
I reverted the downgrade in package-lock.json

KristjanESPERANTO:
Interesting solution. The new check definitely makes things more complicated. But it’s probably better than introducing a known vulnerability.

PR #4086, opened 2026-04-03
ci(codeql): also scan develop branch on push and PR

KristjanESPERANTO:
Previously only master was scanned via the default CodeQL setup. Since development happens on develop, this PR replaces the default setup with a custom workflow that covers both branches. This gives an overview of the security status across the current release (master) and the development branch (develop).

As a result we should also see issues in the develop branch here: https://github.com/MagicMirrorOrg/MagicMirror/security/code-scanning

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,37 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [develop, master]
+  pull_request:
+    branches: [develop]
+  schedule:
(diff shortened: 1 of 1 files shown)


KristjanESPERANTO:
Oh, I'm afraid it only works with the default branch... :-( I'll look into it.

KristjanESPERANTO:
It works now after disabling the default CodeQL setup.

rejas (approved):

PR #4087, opened 2026-04-03
add option to disable or restrict cors endpoint

khassel:


js/server_functions.js
```diff
@@ -25,6 +25,7 @@ async function isPrivateTarget (url) {
 	const hostname = parsed.hostname.replace(/^\[|\]$/g, "");
 
 	if (hostname.toLowerCase() === "localhost") return true;
+	if (config.cors === "allowWhitelist" && !config.corsDomainWhitelist.includes(hostname.toLowerCase())) return true;
```
```suggestion
	if (global.config.cors === "allowWhitelist" && !global.config.corsDomainWhitelist.includes(hostname.toLowerCase())) return true;
```

Using global.config is more explicit and consistent with the rest of the code.

js/server_functions.js
```diff
@@ -68,6 +69,10 @@ function replaceSecretPlaceholder (input) {
  * @returns {Promise<void>} A promise that resolves when the response is sent
  */
 async function cors (req, res) {
+	if (config.cors === "disabled") {
```
```suggestion
	if (global.config.cors === "disabled") {
```

js/server_functions.js
```diff
@@ -68,6 +69,10 @@ function replaceSecretPlaceholder (input) {
...
+		Log.error("CORS is disabled, you need to enable it in `config.js` by setting `cors` to `allowAll` or `allowWhitelist`");
+		return false;
```
```suggestion
		return res.status(403).json({ error: "CORS proxy is disabled" });
```

I think `return false` would leave the HTTP request hanging - Express doesn't send any response to the client, which will time out.

With `res.status(403).json(...)` so the request should properly closed.

KristjanESPERANTO (changes requested):

KristjanESPERANTO (approved):

PR #4091, opened 2026-04-04
fix cors function for alpine linux

khassel:
I provide docker images with alpine linux and tested the new cors approach.

It didn't work because after calling

```js
  const dispatcher = new Agent({ connect: { lookup: (_h, _o, cb) => cb(null, address, family) } });
```

the dispatcher variable was undefined.

This PR solves this and I tested this under debian too.

The mix of internal fetch and newer undici did not work and alpine needs additionally the `process.nextTick`.

### DIFF
--- a/js/server_functions.js
+++ b/js/server_functions.js
@@ -2,7 +2,7 @@ const dns = require("node:dns");
 const fs = require("node:fs");
 const path = require("node:path");
 const ipaddr = require("ipaddr.js");
-const { Agent } = require("undici");
+const { fetch, Agent } = require("undici");
 const Log = require("logger");
 
 const startUp = new Date();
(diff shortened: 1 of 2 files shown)


khassel:
to late to fix this now ...

KristjanESPERANTO:
Yes, I had that problem with my previous PR and I think I have a solution. Can I add a commit?

khassel:
sure

KristjanESPERANTO:
The problem was in the destructuring of `undici` while requiring it. In the production code `const { fetch, Agent } = require("undici")` copies the `fetch` function into a local variable. But in the test code `vi.spyOn(undici, "fetch")` replaces the property on the module object - so the production code holds its own copy and never sees the spy => tests failed (while production worked). Sounds simple, but it took me a few hours to figure that out for the last PR.

Ah, and on top of that, the spy was originally on `global.fetch`, which is an entirely different object than `undici.fetch` after the import change.

KristjanESPERANTO (approved):

khassel:
This should be merged by @rejas or @sdetweil, as Kristjan and I worked on it, thank you.

rejas (approved):

PR #4102, opened 2026-04-10
restrict replaceSecretPlaceholder to cors with allowWhitelist

khassel:
and cleanup spelling


js/server_functions.js
```diff
@@ -22,9 +22,14 @@ function getStartup (req, res) {
  * @returns {string} the input with real variable content
  */
 function replaceSecretPlaceholder (input) {
-	return input.replaceAll(/\*\*(SECRET_[^*]+)\*\*/g, (match, group) => {
-		return process.env[group];
-	});
+	if (config?.cors === "allowWhitelist") {
```
```suggestion
	if (global.config.cors === "allowWhitelist") {
```

- `global` is more explicit
- optional chaining shouldn't be necessary here

KristjanESPERANTO (changes requested):

PR #4104, opened 2026-04-11
fix replaceSecretPlaceholder

khassel:
Disable secret substitution only for cors=allowAll, the last attempt in #4102 was to restrictive.

Secret substitution should also work if cors=disabled.

### DIFF
--- a/js/server_functions.js
+++ b/js/server_functions.js
@@ -22,12 +22,12 @@ function getStartup (req, res) {
  * @returns {string} the input with real variable content
  */
 function replaceSecretPlaceholder (input) {
-	if (global.config.cors === "allowWhitelist") {
+	if (global.config.cors !== "allowAll") {
 		return input.replaceAll(/\*\*(SECRET_[^*]+)\*\*/g, (match, group) => {
 			return process.env[group];
 		});
 	} else {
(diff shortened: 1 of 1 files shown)


js/server_functions.js
```diff
@@ -22,12 +22,12 @@ function getStartup (req, res) {
...
  */
 function replaceSecretPlaceholder (input) {
-	if (global.config.cors === "allowWhitelist") {
+	if (global.config.cors !== "allowAll") {
 		return input.replaceAll(/\*\*(SECRET_[^*]+)\*\*/g, (match, group) => {
 			return process.env[group];
 		});
 	} else {
-		Log.error("Replacing secrets works only with CORS and `allowWhitelist`, you need to set this in `config.js`, set `cors: allowWhitelist`");
+		Log.error("Replacing secrets doesn't work with CORS `allowAll`, you need to set `cors` to `disabled` or `allowWhitelist` in `config.js`");
```
```suggestion
		if (input.includes("**SECRET_")) {
			Log.error("Replacing secrets doesn't work with CORS `allowAll`, you need to set `cors` to `disabled` or `allowWhitelist` in `config.js`");
		}
```

What do you think of something like that? To avoid verbose error logging, only log a warning when a SECRET placeholder is actually detected in the input. Normal CORS requests without secrets won't trigger unnecessary error messages in the logs.

KristjanESPERANTO (approved):

Issue #4116, opened 2026-04-21
[Change Request] Weather forecast: Low before High Temperature


PDStig:
### What problem do you want to solve with this change?

Pedantic. Usually, the low temperature occurs before the high temperature on a given day (around sunrise). However, the UI shows the high temperature first. This may confuse people. Example:

Sunday: high 70, low 30

Someone might think that the "30" low for Sunday will occur Sunday night [/ Monday morning] (actually, it usually occurs Sunday morning).

### What do you think is the correct solution?

Flip the High and Low temperature on the UI. Present the low temperature first. Example:

Sunday: Low 30, High 70.

This presents a better understanding that the low temperature will probably occur before the high temperature does.

### Participation

- [x] I am willing to submit a pull request for this change.

PR #4118, opened 2026-04-21
feat(weather): enable CSS ordering for forecast and hourly columns

### LINKED ISSUE #4117: [Change Request] Weather: Fix inconsistency between hourly and forecast UIs
### What problem do you want to solve with this change?

The UI for the hourly forecast and the daily forecast presents information in different orders.

Hourly forecast: Time, condition, temperature, UV Index, humidity, precipitation amount, precipitation probability
Daily forecast: Day, condition, high/low temperature, precipitation amount, precipitation probability, UV Index

### What do you think is the correct solution?

Adjust the order of the information as follows:

Hourly: Keep the same
Daily: Day, condition, low/high temperature (#4116), UV Index, precipitation amount, precipitation probability

### Participation

### Additional comments

_No response_

MannXo:
The daily forecast previously displayed its columns in a fixed order that did not match the order requested in the issue.

This PR makes the forecast column order configurable through CSS custom properties. The default order is now:

condition, min temp (low), max temp (high), UV Index, precipitation amount, precipitation probability

The hourly forecast is updated to use the same approach, with its own configurable column order.

- Moves `min-temp` before `max-temp` (low before high, per #4116)
- Moves UV Index block before precipitation blocks
- Adds configurable CSS ordering for forecast and hourly columns
- Replaces an invalid `<tr>` in the current weather view with a `<div>`

Fixes #4117

## forecast before
Ekrankopio de 2026-09-12 23-48-38
image

## forecast after
Ekrankopio de 2026-09-12 23-47-47
image

## hourly before
Ekrankopio de 2026-09-12 23-49-12

## hourly after
Ekrankopio de 2026-09-12 23-49-28

----

*Note:* The PR was updated based on reviews, and the PR text was updated accordingly.

### DIFF
--- a/defaultmodules/weather/current.njk
+++ b/defaultmodules/weather/current.njk
@@ -36,10 +36,10 @@
         </span>
       {% endif %}
       {% if config.showUVIndex %}
-        <td class="align-right bright uv-index">
+        <div class="align-right bright uv-index">
           <div class="wi dimmed wi-hot"></div>
           {{ current.uvIndex }}
-        </td>
+        </div>
       {% endif %}
     </div>
   {% endif %}
--- a/defaultmodules/weather/forecast.njk
+++ b/defaultmodules/weather/forecast.njk
@@ -21,19 +21,19 @@
         <td class="bright weather-icon">
           <span class="wi weathericon wi-{{ f.weatherType }}"></span>
         </td>
-        <td class="align-right bright max-temp">{{ f.maxTemperature | roundValue | unit("temperature") | decimalSymbol }}</td>
         <td class="align-right min-temp">{{ f.minTemperature | roundValue | unit("temperature") | decimalSymbol }}</td>
-        {% if config.showPrecipitationAmount %}
-          <td class="align-right bright precipitation-amount">{{ f.precipitationAmount | unit("precip", f.precipitationUnits) }}</td>
-        {% endif %}
-        {% if config.showPrecipitationProbability %}
-          <td class="align-right bright precipitation-prob">{{ f.precipitationProbability | unit('precip', '%') }}</td>
-        {% endif %}
+        <td class="align-right bright max-temp">{{ f.maxTemperature | roundValue | unit("temperature") | decimalSymbol }}</td>
         {% if config.showUVIndex %}
           <td class="align-right dimmed uv-index">
             {{ f.uvIndex }}
             <span class="wi dimmed weathericon wi-hot"></span>
           </td>
+        {% endif %}
+        {% if config.showPrecipitationAmount %}
+          <td class="align-right bright precipitation-amount">{{ f.precipitationAmount | unit("precip", f.precipitationUnits) }}</td>
+        {% endif %}
+        {% if config.showPrecipitationProbability %}
+          <td class="align-right bright precipitation-prob">{{ f.precipitationProbability | unit('precip', '%') }}</td>
         {% endif %}
       </tr>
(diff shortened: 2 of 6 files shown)


KristjanESPERANTO:
Thanks for the PR! Can you attach a before and after screenshot?

rejas:
I am not sure if changing the order is a good idea. Our users are used to how this is ordered now, changing it might confuse some...

But I am willing to see the change in screenshots, before making up my mind finally...

KristjanESPERANTO:
@MannXo Did you noticed the request for screenshots? :slightly_smiling_face:

MannXo:
@KristjanESPERANTO thank you for the reminder.
the before after screenshots are attached.
after
before

KristjanESPERANTO:
@MannXo Thanks for the screenshots :slightly_smiling_face: Please mark which is which. I think the first is the "after" one, right?

MannXo:
@KristjanESPERANTO indeed.
First one is the "after" and second one is "before"

KristjanESPERANTO:
@rejas What do you think?

rejas:
Sorry for my late answer... Thanks fro your PR. I could see something like this merged, IF we could make the order configurable. So that people could arrange them to their likings. Maybe using a flex order css property would be the solution (so people could change that order via custom css)
@MannXo are you comfortable trying something like this out? Or how good are your css skills?

KristjanESPERANTO:
@MannXo Did you noticed the last comment? :slightly_smiling_face:

MannXo:
@KristjanESPERANTO I feel I can do that. shall I create an issue for it first? or shall I just revert my commits on this branch and implement the flex?

KristjanESPERANTO:
I think it would be best to do it in this PR here. When we merge, we'll squash all the commits into one, so you don't have to worry too much about keeping the Git history clean :slightly_smiling_face:

MannXo:
@rejas @KristjanESPERANTO 
I updated this PR to make the forecast column order configurable via CSS.

The forecast table now has a `weather-forecast` class, and each forecast cell has a default `order` value backed by CSS variables, so users can override the visual order from `custom.css` without changing the module template.
I also added an e2e assertion for the CSS order hooks.

example:

```
.weather .weather-forecast {
  --weather-forecast-min-temp-order: 4;
  --weather-forecast-max-temp-order: 3;
}
```

KristjanESPERANTO:
Thanks for the updated approach :clap: 

Two concerns though:

1. **`display: flex` on `<tr>`**
    `order` is only defined for flex/grid items, so applying flex to a table row to make it work feels a bit hacky. Since the table has no `<th>` headers anyway, a `<div>`-based flex layout might be the cleaner path forward.

2. **Order-dependent padding**
    `min-temp` loses `padding-left`, `max-temp` gains it. That works for the new default, but breaks as soon as someone actually swaps the order via the CSS variables. Padding should be independent of position.

What do you think?

KristjanESPERANTO:
@MannXo Did you noticed my comment? :slightly_smiling_face:

KristjanESPERANTO:
Okay, I decided to take this on myself since there hasn’t been any response from @MannXo so far.

Migrating the forecast layout to CSS Grid was a bit tricky, but I think the result is now clean and stable. For consistency, I adapted the hourly view as well.

I added screenshots into the PR text.

While checking the current view, I noticed that an incorrect `<td>` element was being used, so I replaced it with a proper `<div>`. That is a bit out of scope, but too small for a separate PR.

khassel (approved):

PR #4120, opened 2026-04-24
Allow HTTPFetcher to pass through 304 responses

### LINKED ISSUE #4119: HTTPFetcher drops HTTP 304 responses needed by yr provider cache handling
## Summary

In MagicMirror `v2.35.0`, the `yr` weather provider has explicit logic to handle `304 Not Modified` responses and reuse cached data, but `js/http_fetcher.js` still routes all non-OK responses into the error path.

That means a `304` response never reaches `defaultmodules/weather/providers/yr.js`, even though the provider is written to handle it.

## Affected Version

- Confirmed locally on `v2.35.0` (`d05ea751`)

### Provider expects `304`

`defaultmodules/weather/providers/yr.js` contains logic like:

```js
action on response:
if (response.status === 304) {
  use cached data
  return
}
```

### Core fetcher blocks `304`

Unpatched `js/http_fetcher.js` in `v2.35.0` currently behaves like:

```js
if (!response.ok) {
  this.emit("error", errorInfo)
} else {
  this.emit("response", response)
}
```

Since `response.ok` is false for HTTP `304`, the response is treated as an error instead of being emitted to the provider.

## Why This Matters

- `yr` sends cache-aware responses.
- The provider already contains `304` reuse logic.
- The core fetcher prevents that logic from running.
- Result: `304 Not Modified` behaves like an error instead of a cache hit.

## Minimal Fix

Allow `304` to flow through the normal `response` event so cache-aware providers can decide what to do:

```js
if (response.status === 304) {
  this.emit("response", response)
} else if (!response.ok) {
  const { delay, errorInfo } = this.#getDelayForResponse(response)
  nextDelay = delay
  this.emit("error", errorInfo)
} else {
  this.emit("response", response)
}
```

## Expected Behavior

When the upstream server returns `304 Not Modified`, providers that are built to reuse cached data should receive that response and handle it themselves.

## Actual Behavior

`HTTPFetcher` classifies `304` as an error before the provider sees it.

## Local Impact

This was observed while using the built-in weather `yr` provider with cache headers and existing cached weather data. The local workaround is a small patch to `js/http_fetcher.js` that lets `304` responses reach the provider.

sonnyb9:
## Summary

This updates `js/http_fetcher.js` so HTTP `304 Not Modified` responses are emitted through the normal `response` event instead of being treated as errors.

That aligns the core fetcher with the built-in `yr` weather provider in `v2.35.0`, which already includes explicit logic to handle `304` and reuse cached data.

Closes #4119.

## Root cause

`HTTPFetcher` currently treats all non-OK responses as errors. Since `304` is not an OK response, it never reaches providers listening on the `response` event.

At the same time, `defaultmodules/weather/providers/yr.js` expects to receive `304` and handle it by reusing cached data.

## What changed

- special-case HTTP `304` in `HTTPFetcher`
- emit `304` through the normal `response` event
- preserve the existing error path for other non-OK statuses

## Validation

- confirmed `defaultmodules/weather/providers/yr.js` already expects `response.status === 304`
- compared the unpatched `v2.35.0` `HTTPFetcher` behavior to the provider logic
- ran `node --check js/http_fetcher.js`

### DIFF
--- a/js/http_fetcher.js
+++ b/js/http_fetcher.js
@@ -273,7 +273,9 @@ class HTTPFetcher extends EventEmitter {
 				signal: controller.signal
 			});
 
-			if (!response.ok) {
+			const isSuccessfulResponse = response.ok || response.status === 304;
+
+			if (!isSuccessfulResponse) {
 				const { delay, errorInfo } = this.#getDelayForResponse(response);
 				nextDelay = delay;
 				this.emit("error", errorInfo);
--- a/tests/unit/functions/http_fetcher_spec.js
+++ b/tests/unit/functions/http_fetcher_spec.js
@@ -51,6 +51,31 @@ describe("HTTPFetcher", () => {
 			expect(text).toBe(responseData);
 		});
 
+		it("should treat 304 responses as successful and reset error counters", async () => {
+			server.use(
+				http.get(TEST_URL, () => {
+					return new HttpResponse(null, { status: 304 });
+				})
+			);
+
+			fetcher = new HTTPFetcher(TEST_URL, { reloadInterval: 60000 });
+			fetcher.serverErrorCount = 2;
+			fetcher.networkErrorCount = 3;
+
+			const responsePromise = new Promise((resolve) => {
+				fetcher.on("response", (response) => {
+					resolve(response);
+				});
+			});
+
+			fetcher.startPeriodicFetch();
+			const response = await responsePromise;
+
+			expect(response.status).toBe(304);
+			expect(fetcher.serverErrorCount).toBe(0);
+			expect(fetcher.networkErrorCount).toBe(0);
+		});
+
 		it("should emit error event on network failure", async () => {
 			server.use(
 				http.get(TEST_URL, () => {


rejas:
Please base your PR against the develop branch

js/http_fetcher.js
```diff
@@ -268,7 +268,10 @@ class HTTPFetcher extends EventEmitter {
 				signal: controller.signal
 			});
 
-			if (!response.ok) {
+			if (response.status === 304) {
+				// Allow providers with cache-aware handling to reuse prior data.
+				this.emit("response", response);
+			} else if (!response.ok) {
```
KristjanESPERANTO: This would skip the error count reset for 304s - but a 304 means the server is healthy, so serverErrorCount/networkErrorCount should still be cleared.

This would fix that:

```js
			const isSuccessfulResponse = response.ok || response.status === 304;

			if (!isSuccessfulResponse) {
```

 The reset would happen in the else branch.

Additionally, we could even reverse the if condition to handle the happy path first, which would make the code even easier to understand. That wouldn't strictly fall within the scope of this PR, but it would be a small refactoring that I would consider as a minor acceptable tweak.

sonnyb9:
Thanks for the reply.

Retargeted this PR to `develop` and force-pushed a rebased branch so the diff stays limited to the `HTTPFetcher` change.

I also updated the implementation to treat `304` as part of the successful response path, which preserves the existing `serverErrorCount` / `networkErrorCount` reset behavior, and added a focused unit test for that case.

Validation on this checkout: `node --check js/http_fetcher.js`. I also attempted `npm test -- --run tests/unit/functions/http_fetcher_spec.js`, but `vitest` is not installed locally in this environment.

KristjanESPERANTO:
> but vitest is not installed locally in this environment.

Once you run `npm run install`, you can run the tests.

Is there a reason why the PR is still marked as a draft?

sonnyb9:
Tests are passing locally now after installing the repo dependencies.

Ran:
`./node_modules/.bin/vitest run tests/unit/functions/http_fetcher_spec.js`

Result:
`Test Files  1 passed (1)`
`Tests  20 passed (20)`

Marked the PR ready for review.

KristjanESPERANTO (approved):

KristjanESPERANTO:
Looks good, thank! :slightly_smiling_face: Even though there are plenty of clear indications of this, let me ask specifically: This is all Vibe Coded, right?

It’s not a problem for me; I’d just like to have it explicitly stated. Maybe we should include this in our PR requirements, like other projects - specifying how much of the process was handled by AI.

sonnyb9:
Yes, it was coded with codex/gpt-5.4.

khassel (approved):

PR #4121, opened 2026-04-25
fix(weather): avoid loading state after reconnect

KristjanESPERANTO:
When a client reconnects while the backend is still in its rate-limit protection phase, the weather module has no data to show and stays on `Loading...` until the next scheduled API call. This mainly affects server mode setups, where the server keeps running while a remote client temporarily loses its connection and reloads. It was [raised in the forum](https://forum.magicmirror.builders/topic/20218/request-loop-loading...-in-standard-weather-module-open-meteo-after-update/11?_=1777106416020) and is worthy of a fix to improve the user experience.

With this PR the node helper caches the last successful `WEATHER_DATA` payload per instance and replays it immediately on reconnect. The client gets its last known state right away instead of waiting for the next fetch. The cache is cleaned up when the provider stops.

Tests are included to cover reconnect with and without cached data, and the cleanup path.

### DIFF
--- a/defaultmodules/weather/node_helper.js
+++ b/defaultmodules/weather/node_helper.js
@@ -4,6 +4,7 @@ const Log = require("logger");
 
 module.exports = NodeHelper.create({
 	providers: {},
+	lastData: {},
 
 	start () {
 		Log.log(`Starting node helper for: ${this.name}`);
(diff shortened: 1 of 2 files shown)


sdetweil:
We are always providing server mode functionality.

KristjanESPERANTO:
Yes, but what’s your point in the context of this PR?

sdetweil:
You said applies in server mode. We always have server. Even the electron launches as client

KristjanESPERANTO:
Sure, and I used the terms "mainly" and "remote client" in the same sentence, and I wouldn't call Electron on the same machine a remote client. English isn't my mother tongue, so I'm sorry, I don't get your point. Is there anything wrong with the change?

sdetweil:
Nothing.

khassel (approved):

PR #4134, opened 2026-05-02
refactor(updatenotification): replace pm2 usage with node logic

KristjanESPERANTO:
We can rely on PM2's native restart-on-exit behavior instead of the programmatic pm2 API.

Fixes https://github.com/MagicMirrorOrg/MagicMirror/security/dependabot/82 by removing pm2.

Note: Originally this PR was intended to update pm2, but after discussion, we decided to replace it instead. See the discussion below.

### DIFF
--- a/defaultmodules/updatenotification/node_helper.js
+++ b/defaultmodules/updatenotification/node_helper.js
@@ -49,7 +49,6 @@ module.exports = NodeHelper.create({
 			case "CONFIG":
 				this.config = payload;
 				this.updateHelper = new UpdateHelper(this.config);
-				await this.updateHelper.check_PM2_Process();
 				break;
 			case "MODULES":
 				// if this is the 1st time thru the update check process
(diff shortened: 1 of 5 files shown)


khassel:
I have a general question to the pm2 stuff. We use it here to restart mm after updating.

AFAIS the pm2 stuff does restart mm whenever it exits. So why do we need the whole pm2 stuff inside mm?

Would it be sufficient to exit mm and pm2 does the restart from outside?

Background:
- pm2 is a big dependency
- in my docker setup I don't need it at all, I uninstall it (now I have to remove the test too, otherwise it will fail)

sdetweil:
Good question. I think it was not forcing the error path. Oops, app just disappeared, better restart. 

Vs crash and hope, in the auto update module path

KristjanESPERANTO:
Yes, good question! I took another look and don't see any significant downside for PM2 users if we replace the PM2 logic with node logic. I'm open to provide another PR or adapt this one if we decide to remove PM2.

@sdetweil since Karsten's take is clear, what is yours about the removal?

sdetweil:
I see no reason to keep the pm2 restart code. If we can kill the process we are in as there may be other instances running too

khassel (approved):

PR #4139, opened 2026-05-05
fix: skip postinstall git clean when not in a git repository

angeldeejay:
## What does this PR accomplish?

The `postinstall` script runs `git clean -df fonts vendor modules/default`
unconditionally. When `magicmirror` is installed as an npm dependency
(e.g. `npm install magicmirror` or as a transitive dep in another project),
npm runs `postinstall` in a directory that is **not** a git repository.
This causes:

```
fatal: not a git repository (or any of the parent directories): .git
```

npm treats the non-zero exit as a failure → the entire installation aborts
with `code 128`. This makes it impossible to use `magicmirror` as an npm
dependency in any third-party project.

The issue has been present since at least **v2.34.0**. In **v2.35.0**
`modules/default` was added to the clean targets, but the root cause
predates that change.

## Fix

Added `scripts/postinstall.js` — a cross-platform Node.js script that
guards the `git clean` call with a `git rev-parse --git-dir` check.

- When running inside a real git repository: behaviour is identical to before.
- When running outside one (e.g. installed as an npm package): step is silently skipped.

`package.json` `postinstall` updated from the bare `git clean` call to:

```json
"postinstall": "node scripts/postinstall.js"
```

This approach is cross-platform (Linux, macOS, Windows) and keeps the
logic readable rather than inlined as a one-liner.

Does this solve a related issue?

No existing issue tracked. Discovered while using magicmirror as a
devDependency in a companion tooling project — installation failed
unconditionally on all platforms.

Checklist

- Targets the develop branch
- No visual changes (script + package.json only)
- node --run lint:prettier run — no changes needed for .js script

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
 	"sideEffects": true,
 	"scripts": {
 		"config:check": "node js/check_config.js",
-		"postinstall": "git clean -df fonts vendor modules/default",
 		"install-mm": "npm install --no-audit --no-fund --no-update-notifier --only=prod --omit=dev",
 		"install-mm:dev": "npm install --no-audit --no-fund --no-update-notifier && npx playwright install chromium",
 		"lint:css": "stylelint 'css/**/*.css' 'defaultmodules/**/*.css' --fix",


angeldeejay:
@KristjanESPERANTO can you take a look on this please? I've found this  issue while developing MagicMirror sandbox

KristjanESPERANTO:
Looks good to me. But I'm not the only maintainer :slightly_smiling_face: 

@khassel, what do you think? Maybe we should take this opportunity to add a note to the script explaining that this is necessary for migrating from an older systems (<2.35.0) to a newer one, and that we can remove it in a few years. We wouldn't have been able to include a comment like that in package.json.

khassel:
As far as I'm concerned, it can be merged. Another option would be to remove the `postinstall` command entirely.

Its sole purpose is to remove folders whose contents have been moved and which would otherwise appear in `git status`.

For installations that haven't been migrated yet, that's just how it is; the user will have to clean it up themselves.

sdetweil:
I will update the update script to remove the old default module folder

khassel:
> I will update the update script to remove the old default module folder

with this I think the best choice is to remove the `postinstall` command entirely.

@angeldeejay could you please update this PR accordingly? Thanks and sorry for the inconvenience...

angeldeejay:
@khassel done!

khassel (approved):

angeldeejay:
@khassel @KristjanESPERANTO  Do you want me to drop lock files changes and leave only package.json ones?

KristjanESPERANTO (approved):

angeldeejay:
Assumed yes. I'm guessing your pipelines update lock files by themselves

Issue #3796, opened 2025-05-31
[Bug] Nodejs v24 lacks binaries for arm32


khassel:
### Environment

MagicMirror² version: irrelevant
Node version: v24.x
npm version: irrelevant
Platform: arm32

### Which start option are you using?

npm run start

### Are you using PM2?

No

### Module

None

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

Trying to install nodejs v24 on an arm32 device

### What did you expect to happen?

Finding binaries under https://nodejs.org/dist/latest-v24.x/ and docker images for arm32.

### What actually happened?

The support for arm32 was declared as experimental in nodejs v24. Concerning [this comment](https://github.com/nodejs/docker-node/issues/2232#issuecomment-2867510141) compiling of nodejs v24 fails under arm32.

Nodejs v24 is the upcoming lts release. The current lts release v22 ends 30 Apr 2027.

As consequence we should not deprecate nodejs v24 in this projects as long as possible.

### Participation

- [ ] I am willing to submit a pull request for this change.

KristjanESPERANTO:
That would mean if we would drop support for Node v22 MagicMirror would only run on Raspberry Pi 3 and newer, right?

khassel:
yes, AFAIK the 32-bit versions of the Raspberry Pi OS could no longer be used.

sdetweil:
yes, and only in 64 bit mode, which has a bigger working set.

we have already dropped pi0w

sdetweil:
Still following for next release

khassel:
Nodejs v24 is meanwhile the current lts version.

sdetweil:
and still without armv7l binaries

khassel:
> and still without armv7l binaries

I think they will not come back ... so we are running into this problem faster than expected because upcoming electron v40 already uses node v24.

KristjanESPERANTO:
To put that into concrete terms:

- End of life node 22: 30 Apr 2027 <--- what we have expected until now
- End of life electron 39: 5 May 2026 <--- new date

khassel:
yes, to avoid running an unsupported electron version we would have to migrate to electron v40 (and with that to nodejs v24) with the April release

sdetweil:
yeh, dang

khassel:
[...]

PR #4140, opened 2026-05-05
update dependencies and workflows to node v26

khassel:


sdetweil:
What is  the status of armv7l/32 bit systems?

KristjanESPERANTO:
> What is the status of armv7l/32 bit systems?

Unchanged. Also for Node 26 there are no arm32 binaries.

Note: We are tracking this issue in #3796.

KristjanESPERANTO (approved):

PR #4147, opened 2026-05-07
refactor: rewrite NodeHelper as an ES6 class

KristjanESPERANTO:
This PR rewrites `node_helper.js` to use a native ES6 `class` instead of `Class.extend()` - a manual inheritance helper from 2008, written back when `class` syntax didn't exist yet. Node.js has supported native classes since v6, so there's no reason to keep the workaround around.

The public API is unchanged - module authors still write the same `NodeHelper.create({...})` they always have.

Outcome: same behavior, normal modern JavaScript, better stack traces, and `node_helper.js` no longer pulls in `class.js` on the server side. Removing `class.js` altogether is a follow-up (it's still used on browser side).

### DIFF
--- a/js/node_helper.js
+++ b/js/node_helper.js
@@ -1,20 +1,19 @@
 const express = require("express");
 const Log = require("logger");
-const Class = require("./class");
 const { replaceSecretPlaceholder } = require("#server_functions");
 
-const NodeHelper = Class.extend({
+class NodeHelper {
 	init () {
 		Log.log("Initializing new module helper ...");
-	},
+	}
 
 	loaded () {
 		Log.log(`Module helper loaded: ${this.name}`);
-	},
+	}
 
 	start () {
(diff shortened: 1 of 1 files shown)


KristjanESPERANTO:
The failed Node 26 test seems unrelated to the changes to me. It hangs at the MM installation step; at that point, the modified code is irrelevant.

khassel (approved):

khassel:
> The failed Node 26 test seems unrelated to the changes to me. It hangs at the MM installation step; at that point, the modified code is irrelevant.

yes, tested it on my fork without this PR and same result, maybe introduced with node v26.1.0, will try to test this locally later ...

khassel:
I found that the line `npx playwright install chromium` is the problem, there is already an issue in the playwright repo, see https://github.com/microsoft/playwright/issues/40724

KristjanESPERANTO:
Great find!

PR #4151, opened 2026-05-11
refactor: rewrite Module as an ES6 class

KristjanESPERANTO:
This PR rewrites `module.js` to use a native ES6 `class` instead of `Class.extend()` - the same old inheritance helper that was removed from `node_helper.js` in #4147.

The normal module API stays the same: modules still use `Module.register({...})`. Internally, `Module.create()` now creates a named subclass for each module, copies over a cloned definition, and only calls `init()` when it is actually a function.

Outcome: one less file in the browser bundle, no more magic `_super()` wiring, better stack traces, and tests for the module creation path that did not exist before. `module.js` is also now a plain class with no external dependencies - an intentional step towards `export default Module` when browser scripts eventually move to native ES modules.

Since these changes touch the core of how modules are created, I'd appreciate a close review and any feedback on edge cases I may have missed.


khassel (approved):

khassel:
@KristjanESPERANTO let me know if you want to wait for other reviews, otherwise I will merge it ...

KristjanESPERANTO:
@khassel I think one review is enough. Thanks! :slightly_smiling_face:

PR #4155, opened 2026-05-16
fix systeminformation thread not ending

khassel:
This was introduced when running systeminformation in an own thread.

If the main thread ends before (e.g. when config has errors) the systeminformation thread never ends.

As this can have side effects under pm2 or docker (because the app never exits) this PR restores the old behavior.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -328,6 +328,20 @@ function App () {
 		await this.stop();
 		process.exit(0);
 	});
+
+	/**
+	 *
+	 * @param {number} ms milliseconds to wait
+	 */
+	function blockingSleep (ms) {
+		const int32 = new Int32Array(new SharedArrayBuffer(4));
+		Atomics.wait(int32, 0, 0, ms);
+	}
+
+	process.on("exit", () => {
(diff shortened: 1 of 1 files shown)


KristjanESPERANTO (approved):
Nice solution 👍 Why is this still a draft?

khassel:
it still does not work ...

KristjanESPERANTO:
If you need help with further investigation: How are you reproducing the issue? I'm having trouble seeing it on my end - Worker threads die with the parent process, so no orphans under pm2.

khassel:
as always within a container. This PR solves the problem where threads are not ending but systeminfo is not displayed ... will try again later or tomorrow.

Which setup did you test with pm2 (with this PR or without or both)?

KristjanESPERANTO:
Both, but now I see the issue with the missing output when the config has an error, but no approach to solve it.
Is the systeminfo really relevant in that situation?

khassel:
I have now an ugly approach to solve it ...

KristjanESPERANTO (approved):
I don't think it's that ugly. Since I don't see a better approach I'm okay with this one :)

PR #4158, opened 2026-05-17
refactor: use ES module imports in browser core

KristjanESPERANTO:
With these changes a few browser-side core files now use native ES modules. `Loader`, `MMSocket`, `Module` and `MM` can be imported directly instead of being read off `window`. `main.js` and `loader.js` are no longer wrapped in IIFEs - they're just normal modules now.

`Module`, `MM` and `MMSocket` are still exposed as globals, so third-party modules that use the old API keep working.

The changes are mostly structural, behavior should stay the same. A few internal helpers in `main.js` got an underscore prefix because their names clashed with public `MM` methods.

## Why

The old setup relied a lot on script order: a file could use `Loader` or `MMSocket` only because another script happened to put it on `window` first. Imports make that explicit.

The bigger goal is to move away from the legacy script-loading patterns - making it easier to understand and easier to test - in other words: easier to maintain.

More of the core could be "cleaned up" the same way, but that would blow up this PR.

For reviewing, I recommend to hide the whitespace changes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,11 +51,9 @@
     <script type="text/javascript" src="translations/translations.js"></script>
     <script type="text/javascript" src="js/translator.js"></script>
     <script type="text/javascript" src="config/basepath.js"></script>
-    <script type="text/javascript" src="js/module.js"></script>
-    <script type="text/javascript" src="js/loader.js"></script>
-    <script type="text/javascript" src="js/socketclient.js"></script>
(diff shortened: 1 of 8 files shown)


khassel:
Reviewing this is a chore. It’s also surprisingly how poor the `git diff` is in this case. We’ll see if it works...

khassel (approved):

KristjanESPERANTO:
Thanks for the trust and merging! :)

PR #4160, opened 2026-05-18
systeminformation thread not ending: move error handling from utils to app

khassel:
This reverts #4155 because that didn't work when running under electron (`process.on("exit",...)` is not called).

This approach puts a try/catch around the `start` function in `app.js` and in `utils.js` the `process.exit(1)` calls are replaced with `throw new Error("process.exit:1");`. The error message `process.exit:1` is not evaluated in `app.js` but needed in the tests.

The short delay is now implemented in the catch block so the systeminformation is displayed.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -183,80 +183,87 @@ function App () {
 	 * @returns {Promise<object>} the config used
 	 */
 	this.start = async function () {
-		const configObj = Utils.loadConfig();
-		global.config = configObj.fullConf;
-		const config = global.config;
-		Utils.checkConfigFile(configObj);
+		try {
+			const configObj = Utils.loadConfig();
+			global.config = configObj.fullConf;
+			const config = global.config;
+			Utils.checkConfigFile(configObj);
 
-		global.defaultModulesDir = config.defaultModulesDir;
-		defaultModules = require(`${global.root_path}/${global.defaultModulesDir}/defaultmodules`);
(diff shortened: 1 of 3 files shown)


js/app.js
```diff
@@ -183,80 +183,87 @@ function App () {
...
+				}
+			});
 
-		Log.log("Sockets connected & modules started ...");
+			Log.log("Sockets connected & modules started ...");
 
-		return global.config;
+			return global.config;
+		} catch {
+			const int32 = new Int32Array(new SharedArrayBuffer(4));
```
KristjanESPERANTO, line 268: What about adding a log line here, so unexpected errors don't get swallowed silently? Something like:

```suggestion
			// planned config exits already logged their message before throwing
			if (error?.message !== "process.exit:1") {
				Log.error("Unexpected error during startup:", error);
			}
			const int32 = new Int32Array(new SharedArrayBuffer(4));
```


Otherwise anything else that throws in `start()` (a broken `require`, `httpServer.open()` failing, …) just turns into `exit(1)` without a stack.

KristjanESPERANTO (commented):
Looks good! One additional thought: would it make sense to introduce a small dedicated error class (e.g. `ConfigError`) instead of matching on the `"process.exit:1"` message string? That might make the intent a bit clearer. No strong opinion though - could also just as well be a follow-up PR.

PR #4161, opened 2026-05-19
feat(electron): support object-based electronSwitches

### LINKED ISSUE #4159: [Bug] 2.37 dev branch,. fails on electronSwitches
- SYSTEM:   manufacturer: System manufacturer; model: System Product Name; virtual: false; MM: v2.37.0-develop
- OS:       platform: linux; distro: Ubuntu; release: 22.04.5 LTS; arch: x64; kernel: 5.15.0-177-generic
- VERSIONS: electron: 42.1.0; used node: 24.15.0; installed node: 22.22.1; npm: 10.9.4; pm2: 5.4.3
- ENV:      XDG_SESSION_TYPE: x11; MM_CONFIG_FILE: undefined
            WAYLAND_DISPLAY:  undefined; DISPLAY: :0.0; ELECTRON_ENABLE_GPU: undefined
- RAM:      total: 48076.70 MB; free: 40129.30 MB; used: 7947.40 MB
- OTHERS:   uptime: 11076 minutes; timeZone: America/Chicago
[2026-05-18 14:18:58.875] [LOG]   [electron] Launching applic

node --run start:x11

### Are you using PM2?

No

### Module

None

- [x] Yes
- [ ] No

- [x] Yes

### What did you do?

<details>
<summary>Configuration</summary>

```
    electronSwitches: {
      "password-store": "basic",
      "js-flags": "--max-old-space-size=8192"
    },
```
</details>

```js
<!-- Paste relevant code here -->
```

Steps to reproduce the issue:

has been working for 'years' (or maybe not working)

### What actually happened?

 I added the console.log  elsw...
```
[2026-05-18 14:18:58.876] [LOG]   [electron] elsw { 'password-store': 'basic', 'js-flags': '--max-old-space-size=8192' }
[2026-05-18 14:18:58.877] [ERROR] [unknown] (node:725432) UnhandledPromiseRejectionWarning: TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))
    at createWindow (/home/sam/MagicMirror.2.36.dev/js/electron.js:49:39)

```
config.js
```
    electronSwitches: {
      "password-store": "basic",
      "js-flags": "--max-old-space-size=8192"
    },
```

from the electron doc , SOME switches are keyword/value

```
app.commandLine.appendSwitch('remote-debugging-port', '8315')
app.commandLine.appendSwitch('host-rules', 'MAP * 127.0.0.1')
```

so the config value cannot be flat array of single values.. ,, I think appendswitch of {'remote-debugging-port': '8315'} , would fail too

i'm guessing the old code didn't work either
```
        let electronSwitchesDefaults = ["autoplay-policy", "no-user-gesture-required"];
        app.commandLine.appendSwitch(...new Set(electronSwitchesDefaults, config.electronSwitches));
```
at least, the defaults were connected , "autoplay-policy":"no-user-gesture-required" 

changing my switches to 
```
    electronSwitches: [
      {"password-store": "basic"},
      {"js-flags": "--max-old-space-size=8192"}
    ],
```

doesn't produce an error on startup, but I don't know if the switches are correctly passed.. 

### Additional comments

_No response_

### Participation

KristjanESPERANTO:
Previously, `electronSwitches` only accepted strings. This PR adds support for object entries, allowing switches with values:

```js
electronSwitches: [
  "no-sandbox",
  { "js-flags": "--max-old-space-size=8192" }
]
```

I decided to put the logic into a helper file so it can be unit-tested independently, since electron.js itself requires a live Electron environment and cannot be tested in isolation.

Fixes #4159

### DIFF
--- a/js/electron.js
+++ b/js/electron.js
@@ -3,6 +3,7 @@
 const electron = require("electron");
 const core = require("./app");
 const Log = require("./logger");
+const { applyElectronSwitches } = require("./electron_helper");
 
 // Config
 let config = process.env.config ? JSON.parse(process.env.config) : {};
@@ -43,10 +44,7 @@ function createWindow () {
 		Log.warn("Could not get display size, using defaults ...");
 	}
 
-	app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-	for (const electronSwitch of (config.electronSwitches || [])) {
-		app.commandLine.appendSwitch(electronSwitch);
-	}
+	applyElectronSwitches(app.commandLine, config.electronSwitches);
 	let electronOptionsDefaults = {
 		width: electronSize.width,
 		height: electronSize.height,
--- a/js/electron_helper.js
+++ b/js/electron_helper.js
@@ -1,0 +1,33 @@
+const Log = require("./logger");
+
+/**
+ * Applies Electron command-line switches (defaults + from config).
+ * @param {object} commandLine Electron commandLine API
+ * @param {Array<string|object>} [electronSwitches] User-configured switches
+ */
+function applyElectronSwitches (commandLine, electronSwitches) {
+	commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
+	Log.debug("Activated default switch: autoplay-policy=no-user-gesture-required");
+
+	if (electronSwitches === undefined) return;
+	if (!Array.isArray(electronSwitches)) {
+		Log.error(`electronSwitches must be an array of strings or objects, got: ${JSON.stringify(electronSwitches)}`);
+		return;
+	}
+
+	for (const sw of electronSwitches) {
+		if (typeof sw === "string") {
+			commandLine.appendSwitch(sw);
+			Log.debug(`Activated switch: ${sw}`);
+		} else if (sw && typeof sw === "object" && !Array.isArray(sw)) {
+			for (const [name, value] of Object.entries(sw)) {
+				commandLine.appendSwitch(name, String(value));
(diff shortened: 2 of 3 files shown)


sdetweil (changes requested):
autoplayPolicy is an object. And it looks like it’s moved to web preferences, altho there appears to be a cli parm with — prefix

Image

KristjanESPERANTO:
Good catch! I hadn't questioned it since it was already there before my changes. So you're right - since `webPreferences.autoplayPolicy` defaults to `"no-user-gesture-required"` anyway, the CLI switch is useless. I've removed it.

sdetweil (changes requested):
In the for loop name/value is there a way to catch value as  string or number not object, so String(value) doesn’t fail

KristjanESPERANTO:
Yes, we could add that check. I’m just not sure it’s a realistic case worth handling: valid values are normally strings or numbers, and passing an object/array would be a fairly unusual config mistake. It also wouldn’t crash - `String(value)` would produce `"[object Object]"` or similar, and Electron would just not apply the switch as intended. Happy to add it if you prefer, but I’d lean toward keeping this simple.

sdetweil:
Ok. I’m alright with no check if you think it is non fatal 

this is a breaking change for my very old configs, I wonder who else it will impact..

KristjanESPERANTO:
Since this never really worked correctly from what I can tell, I don’t see a big risk of breaking many existing setups. I might be wrong, of course, but my suggestion is to merge this and see if anyone reports a real-world config.

Issue #4150, opened 2026-05-11
Problems with new nodejs >=24.16.0 and >=v26.1.0


khassel:
I know that Node.js v26 is still far from being production-ready; however, for early testers, here are a few pointers:

- [playwright install hangs in node v26](https://github.com/microsoft/playwright/issues/40724), this is meanwhile solved with playwright `v1.60.0`
- electron `v42` fails to install. AFAIS the used `extract-zip` package fails to unzip the downloaded file. [Issue](https://github.com/max-mapper/extract-zip/issues/154)
- ...

At the moment we are using nodejs `v26.0.0` in our automated tests which works. Upgrading to newer versions is not possible at the moment.

KristjanESPERANTO:
The issue now also occurs with new point releases of node 24: 24.16.0 doesn't work anymore.

Spotted in PR #4167: https://github.com/MagicMirrorOrg/MagicMirror/actions/runs/26314372877

khassel:
yes, [my issue](https://github.com/electron/electron/issues/51619) was already updated and pinned.

There is now a simpler fix with overriding `yauzl` in the last comments. Don't know if we should implement this or just wait for the official fix from electron (pressure should be high enough now) ...

KristjanESPERANTO:
Let's wait some days :)

KristjanESPERANTO:
Status update: A workaround was implemented via #4175, so our next release will work with the new node versions.

khassel:
My electron issue was closed yesterday so I assume new releases with this fix are coming soon ... (then we can remove the override fix)

khassel:
The electron install problem is now fixed with electron `v42.4.0`. There is still no fix for electron `v41` which is used by our latest release.

khassel:
> There is still no fix for electron v41 which is used by our latest release.

fixed with electron `v41.7.2`

KristjanESPERANTO:
Nice. So we could close this issue, right?

PR #4162, opened 2026-05-20
chore: update dependencies

KristjanESPERANTO:
This updates the dependencies and implements a workaround for the failing ci tests on node 26.1.0 (#4150).

I'm not sure myself whether we should go with this workaround or wait for an upstream fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
 				"@fortawesome/fontawesome-free": "^7.2.0",
 				"animate.css": "^4.1.1",
 				"croner": "^10.0.1",
-				"eslint": "^10.3.0",
+				"eslint": "^10.4.0",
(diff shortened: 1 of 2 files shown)


khassel:
I think we should wait for the upstream fix, there is no need to have the newest node v26 version in the workflow.

Beside this I opened an [additional issue](https://github.com/electron/electron/issues/51619) which got a new status `blocks stable` today so this will hopefully fixed soon.

KristjanESPERANTO:
Totally fine with dropping the workaround. Just reduced the PR to the dependency updates. This should solve some of [the dependabot alerts](https://github.com/MagicMirrorOrg/MagicMirror/security/dependabot).

khassel (approved):

Issue #4103, opened 2026-04-11
[Bug] Calendar parsed prior to filtering - unable to load large ics file


Evarda:
### Environment

Hi, long time user. Don't think you need my system info for this one, so I'm going to skip this. I updated recently and pulled the code below from the main version of the Git Repo.
I have an issue with a calendar I've used for many years, has many events. All my other calendars work fine, all of my new calendars work fine. I run this on a raspberry pi with decent specs but also run my homebridge on it. This calendar used to be the slowest to load - one day it just stopped loading and has not loaded since.
I noticed the calendar fetcher parses the entire ical and then filters the events.
Can we fix this to make the filter happen when the calendar is loaded rather than loading the information first then filtering it?
I am fairly certain my calendar not loading is related to this. I have also seen others report calendar not loading issues, and I wonder if this is the root cause (I'm not sure what limits the size of parsed).

This is in calendarfetcher.js, see how it loads the ics and then parses after:

```
try {

			const responseData = await response.text();
			const parsed = ical.parseICS(responseData);

			Log.debug(`Parsed iCal data from ${this.url} with ${Object.keys(parsed).length} entries.`);

			this.events = CalendarFetcherUtils.filterEvents(parsed, {
				excludedEvents: this.excludedEvents,
				includePastEvents: this.includePastEvents,
				maximumEntries: this.maximumEntries,
				maximumNumberOfDays: this.maximumNumberOfDays
			});
```

I feel like this is what is causing the problem.
Hopefully this is possible to do and wouldn't be problematic to implement!

### Which start option are you using?

node --run server

### Are you using PM2?

No

### Module

None

### Have you tried disabling other modules?

- [x] Yes
- [ ] No

### Have you searched if someone else has already reported the issue on the forum or in the issues?

- [x] Yes

### What did you do?

Reproduce:
Have big old calendar file; try to load; have it never load.
Have many new calendar files; no problems loading

To be fair, I haven't checked the size of the big old calendar file, but I _know_ it has a lot of events on it...

### What did you expect to happen?

I expect calendar to load.

### What actually happened?

Calendar events do not load. No error.

### Additional comments

Sorry I'm not doing a great job filling out the fields with all my system information for this. If this is not the reason, I'm back to the drawing board on this issue... I'm hoping that pointing out the issue in data handling I believe is causing this issue will fix the bug I'm running into.

Also thank you!! :)

### Participation

- [x] I am willing to submit a pull request for this change.

[...]

PR #4168, opened 2026-05-25
perf(calendar): pre-filter ICS data before parsing

KristjanESPERANTO:
This adds a pre-filter that drops out-of-window non-recurring events from the raw ICS data before `node-ical` parses it. Recurring events and anything we cannot classify cheaply are kept - the existing logic handles those as before.

Some calendars accumulate years of old events. Pre-filtering them reduces the amount of work `node-ical` has to do and keeps event loop lag lower. My benchmarks on a fast machine showed a small but measurable speedup (roughly 10-20% with generated test data); on a Raspberry Pi it should be significantly more noticeable.

For the implementation I ended up using [`ics-filter`](https://github.com/runely/ics-filter), suggested by @rejas. I had a custom version first, but ICS date string parsing has enough complexity that it makes sense to delegate it to a package built specifically for this. I also contributed a couple of fixes to `ics-filter` along the way, and the maintainer was responsive and open to improvements, so it seems like a good fit.

Solves #4103.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -257,6 +257,7 @@
 		"pubdate",
 		"radokristof",
 		"rajniszp",
+		"RDATE",
 		"rebuilded",
 		"Reis",
 		"rejas",
--- a/defaultmodules/calendar/calendarfetcher.js
+++ b/defaultmodules/calendar/calendarfetcher.js
@@ -59,7 +59,13 @@ class CalendarFetcher {
 			}
 
 			const responseData = await response.text();
-			const parsed = await ical.async.parseICS(responseData);
+
+			const filteredData = await CalendarFetcherUtils.preFilterICS(responseData, {
+				includePastEvents: this.includePastEvents,
+				maximumNumberOfDays: this.maximumNumberOfDays
+			});
+
+			const parsed = await ical.async.parseICS(filteredData);
 
 			Log.debug(`Parsed iCal data from ${this.url} with ${Object.keys(parsed).length} entries.`);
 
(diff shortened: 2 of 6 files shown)


rejas:
Just a thought: would a library like https://github.com/runely/ics-filter offload some complexity from this PR?

sdetweil:
@rejas  that filters out ALL past events (we want SOME, sometimes)

> the filter will only filter out past events

sdetweil:
i also replied in the issue, not in the PR.

KristjanESPERANTO:
Thanks @rejas for the pointer to `ics-filter`! I had a closer look and did some tests - I think the general approach of that package fits well, though there are a few edges. I just opened a first PR there to address one of them (case-insensitive property names): https://github.com/runely/ics-filter/pull/1

I'm setting this to draft for now and waiting for the author to respond. I'd rather improve an existing package than maintain our own implementation here :)

KristjanESPERANTO:
> that filters out ALL past events (we want SOME, sometimes)

We can pass `now - maximumNumberOfDays` as the min date to `icsFilter()`, which is the same window `filterEvents()` (in `./defaultmodules/calendar/calendarfetcherutils.js`) already uses for past events.

KristjanESPERANTO:
Status update: The maintainer of ics-filter was open to my pull requests and has merged them :) From my view, ics-filter is now suitable for our use case.

KristjanESPERANTO:
I've switched the implementation to `ics-filter` and updated the PR text. The PR is ready for review :slightly_smiling_face:

khassel (approved):

khassel:
@KristjanESPERANTO let me know if you want to wait for others to approve, otherwise I will merge ...

KristjanESPERANTO:
I think one reviewer is enough here, even though I might be biased since this is my own PR :sweat_smile:

PR #4183, opened 2026-06-09
update dependencies incl. electron and revert yauzl-electron-install-fix

khassel:
Fixes electron problem from #4150 for electron v42.


KristjanESPERANTO (approved):

KristjanESPERANTO:
Nice, that we don't need this workaround in the next release :slightly_smiling_face:

PR #4185, opened 2026-06-12
refactor(main): simplify _updateDom with async/await

KristjanESPERANTO:
`_updateDom` was wrapping its entire body in a `new Promise(resolve => {...})` constructor just to chain `getDom()` and `updateDomWithContent()` together. Since `updateDomWithContent` was already converted to async in #4182, we can now just `await` it directly - the manual wrapper, the explicit `Promise.resolve()` normalization, and the nested `.then().catch()` chain all become unnecessary, making the control flow easier to follow.
Also added `.catch(Log.error)` at both call sites, since async functions reject on error instead of swallowing it silently.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -66,9 +66,11 @@ function createDomObjects () {
 
 	updateWrapperStates();
 
-	Promise.all(domCreationPromises).then(function () {
-		_sendNotification("DOM_OBJECTS_CREATED");
-	});
+	Promise.all(domCreationPromises)
+		.then(function () {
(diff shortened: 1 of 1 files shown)


sdetweil:
Cool, that was a head scratcher. Trouble caused long before await

KristjanESPERANTO:
I'm glad you like it. And kudos to everyone who set this up and managed to make sense of such pre-async/await patterns 🤯

khassel (approved):

PR #4187, opened 2026-06-19
fix: electron blank with "::"

MikeBishop:
- electron.js: "::" was not remapped to "localhost" for the Electron loadURL call, causing it to try loading http://:::8080, resulting in a blank screen. 0.0.0.0 was special-cased, but not the IPv6 equivalent.


MikeBishop:
Closed in favor of #4188 .

PR #4196, opened 2026-07-06
refactor: use const instead of let for variable declarations

KristjanESPERANTO:
This replaces `let` with `const` for variable declarations where it was possible. Using `const` [is considered best practice](https://www.xjavascript.com/blog/in-javascript-why-should-i-usually-prefer-const-to-let/#5-best-practices-const-first-let-second).

I have enabled the corresponding ESLint rule and let the ESLint auto-fix the "wrong" usages of `let`.

### DIFF
--- a/config/config.js.sample
+++ b/config/config.js.sample
@@ -8,7 +8,7 @@
  * which will be converted to `config.js` while starting. For more information
  * see https://docs.magicmirror.builders/configuration/introduction.html#enviromnent-variables
  */
-let config = {
+const config = {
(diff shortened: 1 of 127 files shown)


khassel (approved):

khassel:
@KristjanESPERANTO can you correct the description "This replaces const with let for " before merge?

KristjanESPERANTO:
@khassel Thanks 😅 Done :slightly_smiling_face:

PR #4197, opened 2026-07-06
add test if install works for minimal node version

khassel:
and update minimal node version to `v22.22.2`

Background:

The latest release requires node `v22.22.2` because this version is required by a dev dependency, see https://forum.magicmirror.builders/post/131522 

I added a new test for this (I didn't want to add another version to the test matrix ...) so that we see needed minimum node version upgrades earlier on develop (and not after release).


.github/workflows/automated-tests.yaml
```diff
@@ -34,6 +34,25 @@ jobs:
...
+      - name: "Checkout code"
+        uses: actions/checkout@v7
+      - name: "Get minimum node version"
+        id: get-version
+        run: |
+          NODE_VERSION="$(cat package.json | sed -rn 's|.*"node": ">=([0-9]+\.[0-9]+\.[0-9]+).*|\1|p')"
+          echo "Minimal node version is $NODE_VERSION"
+          echo "version=$NODE_VERSION" >> $GITHUB_OUTPUT
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
```
KristjanESPERANTO:
```suggestion
        uses: actions/setup-node@v6
```

KristjanESPERANTO (changes requested):

KristjanESPERANTO:
Nice approach :smiley:

PR #4207, opened 2026-07-17
[calendar] Fix sliced multi-day sub-events starting at 23:59 instead of next day 00:00

### LINKED ISSUE #4206: [Bug] sliceMultiDayEvents: sliced sub-events after the first start at 23:59 instead of next-day 00:00
**Platform**
- MagicMirror² v2.37.0
- default `calendar` module, `sliceMultiDayEvents: true`

**Bug**

With `sliceMultiDayEvents: true`, a multi-day event is split into per-day slices `(1/n)`, `(2/n)`, … Every slice **after the first** is displayed as starting at **23:59** instead of the following day's 00:00.

**Cause**

`modules/default/calendar/calendar.js`, in the slicing loop (around L513–535):

```js
let midnight = eventStartDateMoment.clone().startOf("day").add(1, "day").endOf("day");
...
event.startDate = midnight.format("x");   // start of the NEXT slice
...
midnight = midnight.clone().add(1, "day").endOf("day");
```

`midnight` is `.endOf("day")` (= 23:59:59.999), but it is assigned as the following slice's `startDate`. So each subsequent slice starts at 23:59:59.999 of the previous day instead of 00:00 of its own day, which is what gets rendered as "… 23:59".

**Fix**

Use the start of the day for the next slice's start:

```diff
- event.startDate = midnight.format("x");
+ event.startDate = midnight.clone().startOf("day").format("x");
```

**Reproduce**

Any all-day or timed event spanning ≥2 midnights with `sliceMultiDayEvents: true`; slices 2..n show "… 23:59".

ago1776:
Fixes #4206

### Problem
With `sliceMultiDayEvents: true`, a multi-day event is split into per-day slices `(1/n)`, `(2/n)`, … Every slice **after the first** was displayed as starting at **23:59** instead of the next day's 00:00.

### Cause
In `defaultmodules/calendar/calendar.js`, the slicing loop advances the day boundary with:

```js
let midnight = eventStartDateMoment.clone().startOf("day").add(1, "day").endOf("day");
...
event.startDate = midnight.format("x");   // start of the NEXT slice
```

`midnight` is `.endOf("day")` (23:59:59.999) but is assigned as the following slice's `startDate`, so each subsequent slice starts at 23:59:59.999 of the previous day rather than 00:00 of its own day.

### Fix
Use the start of the day for the next slice's start:

```diff
- event.startDate = midnight.format("x");
+ event.startDate = midnight.clone().startOf("day").format("x");
```

### Test
Adds an e2e regression test (`sliceMultiDayEvents sub-event start time`) that reuses the existing slice mock with a 24h `timeFormat` and asserts that no sliced sub-event time cell contains `23:59`. It fails on the previous behaviour and passes with the fix.

Note: I could not run the full Electron e2e suite locally (resource-constrained hardware running production services); the fix logic was verified with a standalone reproduction and the test follows the existing `sliceMultiDayEvents` patterns. Happy to adjust based on CI.

### DIFF
--- a/defaultmodules/calendar/calendar.js
+++ b/defaultmodules/calendar/calendar.js
@@ -531,7 +531,7 @@ Module.register("calendar", {
 						thisEvent.title += ` (${count}/${maxCount})`;
 						splitEvents.push(thisEvent);
 
-						event.startDate = midnight.format("x");
+						event.startDate = midnight.clone().startOf("day").format("x");
 						count += 1;
 						midnight = midnight.clone().add(1, "day").endOf("day"); // next day
 					}
--- a/tests/configs/modules/calendar/sliceMultiDayEventsStart.js
+++ b/tests/configs/modules/calendar/sliceMultiDayEventsStart.js
@@ -1,0 +1,28 @@
+let config = {
+	address: "0.0.0.0",
+	ipWhitelist: [],
+	timeFormat: 24,
+
+	modules: [
+		{
+			module: "calendar",
+			position: "bottom_bar",
+			config: {
+				hideDuplicates: false,
+				maximumEntries: 100,
+				sliceMultiDayEvents: true,
+				calendars: [
+					{
+						maximumEntries: 100,
+						url: "http://localhost:8080/tests/mocks/sliceMultiDayEvents.ics"
+					}
+				]
+			}
+		}
+	]
+};
+
+/*************** DO NOT EDIT THE LINE BELOW ***************/
+if (typeof module !== "undefined") {
+	module.exports = config;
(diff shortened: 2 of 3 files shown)


ago1776:
Closing as a duplicate of #4208, which carries the same one-line fix plus a regression test I validated locally (electron e2e: passes with the fix, fails on `23:59` without it). Consolidating to keep a single PR for #4206 — sorry for the noise.

KristjanESPERANTO:
Just a friendly hint: You could have adjusted this PR here. That would have caused less noise.

Issue #2547, opened 2021-04-21
Calendar module : reccuring annual events appear every month


lbalogh:
Hi,

Annualy reccuring events in Google Calendar appears every month in the calendar module.

See the ICS file bellow, taken directly from Google : it has 1 event that occurs 25 March of each year. But the calendar module display the event on 25 of every month.

**Platform** : Raspberry Pi 3
**MagicMirror Version** : 2.15.0

`BEGIN:VCALENDAR
PRODID:-//Google Inc//Google Calendar 70.9054//EN
VERSION:2.0
CALSCALE:GREGORIAN
METHOD:PUBLISH
X-WR-CALNAME:xxx@gmail.com
X-WR-TIMEZONE:Europe/Zurich
BEGIN:VTIMEZONE
TZID:Etc/UTC
X-LIC-LOCATION:Etc/UTC
BEGIN:STANDARD
TZOFFSETFROM:+0000
TZOFFSETTO:+0000
TZNAME:GMT
DTSTART:19700101T000000
END:STANDARD
END:VTIMEZONE
BEGIN:VEVENT
DTSTART;VALUE=DATE:20210325
DTEND;VALUE=DATE:20210326
RRULE:FREQ=YEARLY;WKST=MO;INTERVAL=1;BYMONTHDAY=25
DTSTAMP:20210421T154106Z
UID:zzz@google.com
CREATED:20200831T200244Z
DESCRIPTION:
LAST-MODIFIED:20200831T200244Z
LOCATION:
SEQUENCE:0
STATUS:CONFIRMED
SUMMARY:Birthday
TRANSP:OPAQUE
BEGIN:VALARM
ACTION:DISPLAY
DESCRIPTION:This is an event reminder
TRIGGER:-P0DT7H0M0S
END:VALARM
END:VEVENT
`

rejas:
I dont have much knowledge about the ics format, but I tried exporting a yearly date from my google calendar and the RRULE line was like this only:
`RRULE:FREQ=YEARLY`
and everythin looked fine on my MM.

In your ICS file the line is 
`RRULE:FREQ=YEARLY;WKST=MO;INTERVAL=1;`

so probably something in there tells ical/rrule to create more events than you want. MagicMirror itself relies on that and shows these bogus dates therefor.

Could you look into your Google Calendar to see what else is in that date you export?

lbalogh:
Well, I don't have much knowledge of the ics format either, but I did some tests :

[...]

Issue #3047, opened 2023-02-21
Default calendar module: Yearly event is displayed monthly


Tom-Hirschberger:
Hi,

as i already wrote in my pull request the default calendar module displays my yearly birthday events every month.
I attached a ics file (needed to rename it to .txt to be able to upload) with one of the events that causes the problem.

It happened with the current development version (without my changes of the pull request).

[jahrestage.txt](https://github.com/MichMich/MagicMirror/files/10795025/jahrestage.txt)

My configuration looks like the following:
```
		{
			module: "calendar",
			position: "top_center",
			config: {
				timeFormat: "absolute",
				fade: false,
				nextDaysRelative: true,
				displaySymbol: false,
				dateFormat: "dddd MMM Do",
				fullDayEventDateFormat: "dddd MMM Do",
				maxTitleLength: 50,
				//maximumNumberOfDays: 14,
				calendars: [
					{
						url: "http://localhost:8080/modules/jahrestage.ics",
						name: "Jahrestage", // <-- specify calendar name
						broadcastPastEvents: true, // <-- need to broadcast past events
						//maximumEntries: 7, // <-- assign enough number to prevent truncating new events by old events.
						symbol: [], // <-- when you want to display simbol. If you don't want, just set as `symbol:[],`
					}
				]
			},
		},
```

Tom-Hirschberger:
I removed some personal data of the ics file (did not want to share google id and birthday with name of family and friends to the public).
I tried running the ics file with `debug.js` and the effect is the same with the modified and unmodfied version.

khassel:
I compared the provided ics with one of my birthday events and found that

`RRULE:FREQ=YEARLY;WKST=MO;BYMONTHDAY=22`

seems to be problematic. My event contains only

`RRULE:FREQ=YEARLY`

The param `BYMONTHDAY` seems to be the problem, I found some documentation for RRULE [here](https://dateutil.readthedocs.io/en/stable/rrule.html).

Looking in the examples section I found

```bash
Every Friday the 13th, 4 occurrences.

 >>> list(rrule(YEARLY, count=4, byweekday=FR, bymonthday=13,
 ...            dtstart=parse("19970902T090000")))
 [datetime.datetime(1998, 2, 13, 9, 0),
  datetime.datetime(1998, 3, 13, 9, 0),
  datetime.datetime(1998, 11, 13, 9, 0),
  datetime.datetime(1999, 8, 13, 9, 0)]
```
[...]

PR #4222, opened 2026-08-01
fix(calendar): keep yearly events in DTSTART's month when BYMONTH is missing

pascalpfammatter:
A birthday in my Google Calendar showed up on my mirror as "tomorrow" when it is actually in October. It turns out it had been appearing on the 2nd of *every* month, and I had simply never noticed until it landed on a day I paid attention to.

The event looks like this in the ICS feed:

```ics
DTSTART;VALUE=DATE:20231002
DTEND;VALUE=DATE:20231003
RRULE:FREQ=YEARLY;WKST=MO;INTERVAL=1;BYMONTHDAY=2
SUMMARY:Ted Birthday
```

`FREQ=YEARLY` together with `BYMONTHDAY=2`, but no `BYMONTH`. `BYMONTHDAY` is an expanding rule part for `FREQ=YEARLY`, so the expander returns the 2nd of every month — twelve occurrences a year instead of one. Google Calendar's own UI, and the client that wrote the rule, show it once a year on 2 October.

Out of 3354 events in my calendar exactly two had this shape, both created by an older mobile calendar app. Newer clients write a plain `RRULE:FREQ=YEARLY`, which is why this is easy to miss — but the old events are never rewritten, so they keep misbehaving forever.

### Why not fix it in the RRULE library

This has come up before, and each time it was closed as bad input data rather than fixed: #2547 (2021), #3047 (2023), and recently jens-maus/node-ical#531. In that last one a fix was actually written (ggaabe/rrule-temporal#127) and then withdrawn, because a reviewer pointed out that restricting yearly `BYMONTHDAY` rules to the `DTSTART` month breaks legitimate rules such as `FREQ=YEARLY;BYMONTHDAY=1,3`, which really must expand across all twelve months.

That objection is correct, and it is why I don't think the RRULE engine is the right place for this. A general-purpose expander has to follow RFC 5545. A calendar display, on the other hand, can afford to be liberal about a well-known broken export as long as the detection is narrow enough that no correct rule is touched.

### What this does

`expandRecurringEvent` confines a yearly rule to `DTSTART`'s month, but only when the rule is provably a redundant restatement of `DTSTART` rather than a real expansion. All of these must hold:

- `FREQ=YEARLY`
- `BYMONTHDAY` holds exactly one value
- that value equals `DTSTART`'s day-of-month
- no `BYMONTH`, `BYDAY`, `BYYEARDAY`, `BYWEEKNO` or `BYSETPOS`

Under those conditions the expansion yields `DTSTART`'s own date plus eleven dates the author never wrote, so dropping the extras cannot lose a real occurrence.

| Rule | Affected? | Result |
| --- | --- | --- |
| `FREQ=YEARLY;...;BYMONTHDAY=2`, DTSTART Oct 2 | yes | once a year, 2 October |
| `FREQ=YEARLY;BYMONTHDAY=1,3` | no, two values | unchanged, expands over 12 months |
| `FREQ=YEARLY;BYMONTHDAY=13;BYDAY=FR` | no, `BYDAY` present | unchanged, every Friday the 13th |
| `FREQ=YEARLY;BYMONTHDAY=7`, DTSTART on the 2nd | no, day mismatch | unchanged, expands |
| `FREQ=YEARLY;BYMONTHDAY=2;BYMONTH=10` | no, `BYMONTH` present | unchanged, once a year |
| `FREQ=YEARLY` | no, no `BYMONTHDAY` | unchanged, once a year |

The rule is read through node-ical's public `options`, and both the current string `freq` and the older numeric one are accepted.

### Tests

Six new cases in `calendar_fetcher_utils_spec.js`. The five "unaffected" rows above are tests that already pass without the change — they are there to pin the behaviour that must not regress. Only the first one fails before the fix.

Calendar unit tests pass on `Europe/Zurich`, `America/New_York`, `Pacific/Auckland`, `Asia/Kolkata` and `UTC`. `lint:js` and `lint:prettier` are clean.

Against my own feed over the next 365 days the number of occurrences goes from 132 to 110, and the difference is exactly the 22 spurious instances from those two events — nothing else changes.

### One thing worth deciding

I've made this unconditional, since the detection is narrow enough that no correct rule is affected. If you would rather have it behind a calendar config option, say so and I'll move it.

### DIFF
--- a/defaultmodules/calendar/calendarfetcherutils.js
+++ b/defaultmodules/calendar/calendarfetcherutils.js
@@ -234,42 +234,28 @@ const CalendarFetcherUtils = {
 	 * instead of once. The clients that emit this, and the web UIs that render it, all show
 	 * it once a year on DTSTART's date, so we follow the author's evident intent.
 	 *
-	 * The test is deliberately narrow, so rules that genuinely expand are left alone:
-	 * BYMONTHDAY must carry exactly one value, that value must equal DTSTART's day-of-month
-	 * (i.e. it merely restates DTSTART), and no other BYxxx part may shape the recurrence.
+	 * Detection reads rrule.origOptions, which carries only the parts the author actually
+	 * wrote (not the defaults node-ical fills in), so a missing BYMONTH is unambiguous. It
+	 * stays deliberately narrow, leaving rules that genuinely expand alone: BYMONTHDAY must
+	 * hold exactly one value equal to DTSTART's day-of-month (merely restating DTSTART), and
+	 * no other BYxxx part may shape the recurrence.
 	 * @param {object} event The recurring event object
 	 * @returns {boolean} True if the rule should be confined to DTSTART's month
 	 */
 	isYearlyRuleMissingByMonth (event) {
-		const rule = event.rrule;
-		if (!rule) {
+		const options = event.rrule?.origOptions;
+		if (!options || options.freq !== "YEARLY") {
 			return false;
 		}
 
-		const options = typeof rule.options === "function" ? rule.options() : rule.options;
-		// node-ical >= 0.26 reports freq as a string, earlier versions used rrule.js' numeric RRule.YEARLY.
-		if (!options || (options.freq !== "YEARLY" && options.freq !== 0)) {
+		const isSingleMonthDay = Array.isArray(options.byMonthDay) && options.byMonthDay.length === 1;
+		// Any other BYxxx part means the rule shapes the recurrence on purpose.
+		const hasShapingPart = Boolean(options.byMonth || options.byDay || options.byYearDay || options.byWeekNo || options.bySetPos);
+		if (!isSingleMonthDay || hasShapingPart) {
 			return false;
 		}
 
-		const byMonthDay = options.byMonthDay ?? options.bymonthday;
-		if (!Array.isArray(byMonthDay) || byMonthDay.length !== 1) {
-			return false;
-		}
-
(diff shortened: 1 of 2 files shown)


KristjanESPERANTO:
Thanks for the deep dive here, and for the references to the earlier discussions. I think the direction here is fine.

What do you think of this: instead of reading `rrule.options` and guessing between camelCase/lowercase keys and string/numeric `freq`, read `rrule.origOptions`. It only contains what was actually written in the RRULE (not what the library fills in as a default), so detecting "no `BYMONTH`" becomes unambiguous.

```js
	isYearlyRuleMissingByMonth (event) {
		const options = event.rrule?.origOptions;
		if (!options || options.freq !== "YEARLY") {
			return false;
		}

		const isSingleMonthDay = Array.isArray(options.byMonthDay) && options.byMonthDay.length === 1;
		// Any other BYxxx part means the rule shapes the recurrence on purpose.
		const hasShapingPart = Boolean(options.byMonth || options.byDay || options.byYearDay || options.byWeekNo || options.bySetPos);
		if (!isSingleMonthDay || hasShapingPart) {
			return false;
		}

		return options.byMonthDay[0] === event.start.getDate();
	},
```

KristjanESPERANTO:
@pascalpfammatter Did you noticed my comment? :slightly_smiling_face:

KristjanESPERANTO:
Since we didn't receive any feedback, I went ahead and made the change I suggested myself.

khassel (approved):

PR #4244, opened 2026-08-31
Fixes issue 4243: Ensure while loop in calendar module does not get stuck when no entries

### LINKED ISSUE #4243: [Bug] All items not loading when calendar loading fails
### Environment

MagicMirror² version: 2.37.0
Node version: 26.3.0
npm version: 11.16.0
Platform: Raspberry Pi OS and MacOS

node --run start

### Are you using PM2?

Yes

### Module

calendar

- [x] Yes
- [ ] No

- [x] Yes

### What did you do?

When i started up the program on my mirror all of a sudden all modules wouldn't load anymore. I tried disabling all other modules and found that the issue was in the calendar module.

I expected everything to startup as usually where all modules load.

### What actually happened?

When the calendar has no entry for a single day, the while loop in the calendar module does not increment the counter and it gets into an infinite loop. That causes the whole system to become stuck and stop loading the calendar and other modules.

```js
while (daysCollected < this.config.limitDays) {
	const dateStr = currentDate.format("YYYY-MM-DD");
	// Check if there are events on the currentDate
	if (eventsByDate[dateStr] && eventsByDate[dateStr].length > 0) {
		// If there are any events today then get all those events and select the currently active events and the events that are starting later in the day.
		newEvents.push(...eventsByDate[dateStr].filter((ev) => this.timestampToMoment(ev.endDate).isAfter(moment())));
		// Since we found a day with events, increase the daysCollected by 1
		daysCollected++; // <---- THIS LINE SHOULD BE OUTSIDE THIS IF STATEMENT
	}
	// Search for the next day
	currentDate.add(1, "day");
}
```

### Additional comments

I have made a fix for this so i will submit this in the coming days when i find the time for it. Should be quite soon since it is only a one line fix.

### Participation

deBasMan21:
Fixes: https://github.com/MagicMirrorOrg/MagicMirror/issues/4243

### DIFF
--- a/defaultmodules/calendar/calendar.js
+++ b/defaultmodules/calendar/calendar.js
@@ -595,9 +595,9 @@ Module.register("calendar", {
 				if (eventsByDate[dateStr] && eventsByDate[dateStr].length > 0) {
 					// If there are any events today then get all those events and select the currently active events and the events that are starting later in the day.
 					newEvents.push(...eventsByDate[dateStr].filter((ev) => this.timestampToMoment(ev.endDate).isAfter(moment())));
-					// Since we found a day with events, increase the daysCollected by 1
-					daysCollected++;
 				}
+				// Increment the daysCollected by one to ensure the while loop does not get stuck.
+				daysCollected++;
 				// Search for the next day
 				currentDate.add(1, "day");
 			}


KristjanESPERANTO (approved):
Thanks for the explanation, that makes sense to me now. Would you also like to add a regression test for this? No problem if not :)

khassel (approved):

PR #4257, opened 2026-09-04
calendar: allow custom events to override symbol class

Antra:
Adds an optional `symbolClassName` property to calendar custom events.

When specified, the custom event's symbol class takes precedence over the calendar-level `symbolClassName`. If it isn't specified, the existing calendar configuration/default behavior is preserved.

For example:
```
{ keyword: "Tesla", symbolClassName: "fab fa-fw fa-", symbol: "tesla", color: "Blue" }
```

Added an E2E test covering a custom event with a custom symbol class.

Also verified the Tesla icon works in a real MagicMirror installation.


sdetweil (changes requested):
Nice.. 

on the testcase, it tests that the override works, but doesn't test that it is not applied when not matched.
could you add that to the test. 

not in this PR, but would you create a PR for the documentation too?

sdetweil (approved):
Thanks

PR #4267, opened 2026-09-15
fix(newsfeed): handle object summaries

KristjanESPERANTO:
This PR fixes #4212.

A fix in [FeedMe](https://github.com/fent/feedme.js) itself would be the ideal solution, but the project appears to be unmaintained. It has not had a release in about five years, and there has been no response to my PR: https://github.com/fent/feedme.js/pull/58

I’ve also added a few additional tests to help prevent regressions in future changes.

----

Note: I have also started looking into alternatives for FeedMe. [rowanmanning/feed-parser](https://github.com/rowanmanning/feed-parser) looks promising so far, but switching to it would require a more extensive code change, so I would prefer to handle that later (after the comming release).

PR #4269, opened 2026-09-16
refactor(newsfeed): replace feedme lib with feedparser

rejas:
feedme.js seems to be dead, lets use something more maintained.

Could supersede #4267


KristjanESPERANTO:
Nice, this looks good! :)

I brought over the tests from my experiments with different parsers and adapted them to your implementation. The existing coverage was a bit thin in this area, and the new tests give us more confidence that the parser switch works as expected. Hope this is okay.

KristjanESPERANTO (approved):

khassel (approved):

PR #4270, opened 2026-09-16
fix wayland display autodetect and add safe update.sh

### LINKED ISSUE #1: Your "secret" ical hash is saved in this repository

flightlesstux:
## Summary
- `start:wayland` defaulted `WAYLAND_DISPLAY` to `wayland-1` when unset, but that's not always the real compositor socket, so a saved pm2 env with no `WAYLAND_DISPLAY` made electron SIGSEGV on every launch, and pm2 auto-restarted the crash loop every ~5s, filling the disk with core dumps.
- `start:wayland` now probes `XDG_RUNTIME_DIR` for the actual `wayland-*` socket before falling back to `wayland-0`.
- Adds `update.sh`: checks disk space, cleans stray core dumps, refuses to run on a dirty working tree before pulling, restarts the pm2 process after `npm install`, and rolls back to the pre-update commit if the app crash-loops or dumps core.

Fixes #1

## Test plan
- [ ] Run `npm run start:wayland` on a system with `WAYLAND_DISPLAY` unset and confirm the correct socket is detected
- [ ] Run `update.sh` on a clean and a dirty working tree and confirm expected behavior


khassel:
Thanks! 

You addressed the wrong wayland default already in #4268 , did you read the comments there?

I don't think we want additional bash scripts as `update.sh` in this repo, the whole bash install part was moved to an own repo https://github.com/sdetweil/MagicMirror_scripts some years ago, I think `update.sh` belongs there.

KristjanESPERANTO:
I agree with @khassel. Let's close this one here and continue in #4268 :)

PR #4278, opened 2026-09-21
Add FMI weather provider

i-xul:
## Description

This pull request adds the Finnish Meteorological Institute (FMI) as a weather provider for the default weather module.

The provider uses FMI Open Data services and does not require an API key.

It supports:

- Current weather observations from FMI weather stations
- Automatic selection of the nearest station with the required observations
- Hourly forecasts using the FMI HARMONIE model
- Daily forecasts aggregated from HARMONIE forecast data
- Finnish local calendar days for daily forecast aggregation
- FMI weather symbol mapping to MagicMirror weather types
- Temperature, humidity, wind, precipitation, and related weather data

The implementation uses latitude and longitude coordinates, consistent with the existing weather provider configuration.

## Related issue

No related issue.

## Additional changes

- Adds FMI-specific unit tests covering configuration, URL construction, observation parsing, station selection, current weather generation, forecast parsing, hourly forecasts, daily forecasts, and forecast fetching.
- Extends the HTTP fetcher to support the FMI XML responses required by the provider.
- Adds tests for the HTTP fetcher changes.
- Adds FMI-related terms to the spelling dictionary.
- Adds FMI to the weather provider release-note category.

## Testing

Tested locally on Windows 11.

FMI provider tests:

`17 passed (17)`

Full unit test suite:

`489 passed, 1 failed (490 total)`

The single failing test is `tests/unit/classes/systeminformation_spec.js`, which expects `platform: linux`. The local test environment reports `platform: Windows`. The FMI provider tests pass successfully.

`npm run config:check` also passes successfully.

The provider was additionally tested with live FMI data in MagicMirror using current, forecast, and hourly weather module instances.


khassel (approved):

js/http_fetcher.js
```diff
@@ -295,7 +304,9 @@ class HTTPFetcher extends EventEmitter {
 			// because Node's global fetch and npm undici@8 Agents are incompatible.
 			// For regular requests, use globalThis.fetch so MSW and other interceptors work.
 			const fetchFn = requestOptions.dispatcher ? undiciFetch : globalThis.fetch;
-			const response = await fetchFn(this.url, {
+			const url = this.#getUrl();
```
KristjanESPERANTO: Could the resolved URL also be passed to the error handling and logging code?

The resolved URL is created here, but the following code still uses `this.url`. When a function is used as the URL, this contains the function itself instead of the actual URL string.

KristjanESPERANTO (changes requested):
Thanks for the PR! This looks good overall, except the url usage.

I see potential to consolidate things, since we have some similarities with the provider evcanada. I'd probably address that in a follow-up PR.

js/http_fetcher.js
i-xul: @KristjanESPERANTO Thanks! Fixed in e1c80e83. The resolved URL is now passed through the error handling and logging code as well, and I added a test to verify that the resolved dynamic URL is included in the error info.

KristjanESPERANTO: Great! Your contribution will be in the next release :)

KristjanESPERANTO:
@i-xul Can you please also add FMI to the docs? :)

https://github.com/MagicMirrorOrg/MagicMirror-Documentation/blob/master/modules/weather.md#general-options

PR #4279, opened 2026-09-21
newsfeed: don't use cors url in njk template

khassel:
A few years ago, the `showTitleAsUrl` property was added to the newsfeed module. It allows users to click on headlines and open the article in their browser (in a new tab or window).

However, it is a mystery to me today why I implemented the use of the cors url there. It makes no sense; the browser should open the original url directly.

This PR fixes that.


KristjanESPERANTO (approved):

KristjanESPERANTO:
Hm... good catch! :)

